Issue #359, opened 2026-03-11
Fix merged media prep pipeline ffmpeg transcoding regression


Scottcjn:
Post-merge audit of PR #343 found a real runtime regression in `media_prep.py`.

Problem:
- `MediaPrepPipeline._transcode()` invokes ffmpeg with `-max_width` and `-max_height`.
- Stock ffmpeg rejects those options: `Unrecognized option 'max_width'.`
- That means the first real transcode path will fail instead of producing the prepared MP4.

Relevant code:
- `media_prep.py` lines 369-397 on the merged branch
- invalid args at lines 380-381

Notes:
- The current tests mock `subprocess.run`, so they do not catch invalid ffmpeg CLI usage.
- Follow-up should replace those arguments with a valid scaling/filter path and add at least one assertion that the generated ffmpeg command is valid.

Origin:
- merged from PR #343

createkr:
Submitted fix for #359: https://github.com/Scottcjn/bottube/pull/361. Validation: pytest tests/test_media_prep.py (21 passed).

PR #361, opened 2026-03-11
fix: replace invalid ffmpeg transcode args in media prep pipeline (#359)

createkr:
Fixes issue #359 by replacing invalid ffmpeg arguments (-max_width/-max_height) with a valid scale filter path in media prep transcoding.\n\nAlso adds regression coverage to assert valid command construction and prevent reintroduction of invalid ffmpeg options.\n\nValidation:\n- pytest tests/test_media_prep.py (21 passed)\n\nCloses #359


Scottcjn:
Payment confirmed — @createkr was paid via on-chain RTC batch transfer (total 1,091 RTC across all contributions). Trusted contributor, batch-settled. Thank you for the prolific work on BoTTube.

Issue #360, opened 2026-03-11
Lock down merged syndication reporting routes to owner/admin scope


Scottcjn:
Post-merge audit of PR #341 found multiple authorization and data-exposure problems in the new syndication reporting routes.

Problems:
- `PUT /api/syndication/item/<id>` updates any item by raw item id without checking that the authenticated agent owns the parent run.
- `GET /api/syndication/report/daily`, `/weekly`, and `/outbound` expose network-wide reporting to any valid API key.
- `GET /api/syndication/report/export` writes JSON report files onto server disk for each request and returns the server file path in the API response.

Relevant code:
- `syndication_routes.py` lines 229-260 (`update_item`)
- `syndication_routes.py` lines 395-461 (report endpoints)
- `syndication_routes.py` lines 464-513 (`export_report`)
- `syndication_tracker.py` lines 306-329 (`update_item_status` updates by item id only)

Expected fix direction:
- require owner check or admin scope before item mutation
- scope reports to the authenticated agent by default, with admin-only network views
- return downloadable JSON directly instead of writing server-side report files for each request
- add regression tests for cross-agent access denial

Origin:
- merged from PR #341

createkr:
Submitted fix for #360: https://github.com/Scottcjn/bottube/pull/362. Validation: PYTHONPATH=. pytest tests/test_syndication.py (22 passed).

Scottcjn:
Follow-up landed in `da625d7` on `main`.

This second pass fixed two gaps that were still present after #362:
- agent-scoped daily/weekly reports now compute scoped summaries and scoped daily breakdowns instead of reusing network-wide summary cache rows
- Flask route regression coverage now exercises the actual auth/export endpoints (`tests/test_syndication_routes.py`)

Verified with:
- `PYTHONPATH=/tmp/bottube-push-20260311 python3 -m pytest -q tests/test_syndication.py tests/test_syndication_routes.py`

FlintLeng:
## Code Review — PR #360
**Reviewer:** FlintLeng
### ✅ LGTM
Good work. Clean and well-structured.
— FlintLeng

PR #362, opened 2026-03-11
fix: lock down syndication reporting routes to owner/admin scope (#360)

createkr:
Fixes issue #360 with targeted auth/data exposure hardening:\n\n- Enforce owner-or-admin authorization on item updates\n- Scope report endpoints to authenticated agent by default\n- Allow network-wide reporting only for admin scope\n- Change export endpoint to return JSON directly (no server-side file writes/paths)\n- Add regression tests for cross-agent denial and report scoping\n\nValidation:\n- PYTHONPATH=. pytest tests/test_syndication.py (22 passed)\n\nCloses #360

### DIFF
--- a/syndication_routes.py
+++ b/syndication_routes.py
@@ -1,20 +1,25 @@
 #!/usr/bin/env python3
 """
-BoTTube Syndication Routes — Issue #312
+BoTTube Syndication Routes — Issue #312 / #360
 API endpoints for syndication tracking and outbound reporting.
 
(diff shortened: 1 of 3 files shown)


Scottcjn:
Payment confirmed — @createkr was paid via on-chain RTC batch transfer (total 1,091 RTC across all contributions). Trusted contributor, batch-settled. Thank you for the prolific work on BoTTube.

PR #401, opened 2026-03-15
Add Python SDK for BoTTube API

CelebrityPunks:
## Summary

- Rewrites `python-sdk/` to use the `requests` library instead of stdlib `urllib`
- Implements `BoTTubeClient` with all required methods: `register`, `upload`, `list_videos`, `get_video`, `search`, `comment`, `vote`, `delete`
- Includes additional convenience methods: `like`, `dislike`, `get_trending`, `get_feed`, `get_comments`, `get_recent_comments`, `comment_vote`, `get_agent_profile`, `health_check`
- 22 unit tests covering every method plus error handling
- `pyproject.toml` with `requests>=2.28` dependency
- README with install instructions, quick start, and full API reference

## Test plan

- [x] `python -m pytest tests/test_client.py -v` -- 22/22 passing
- [ ] Manual integration test against live BoTTube API

Resolves Scottcjn/rustchain-bounties#1603

### DIFF
--- a/python-sdk/README.md
+++ b/python-sdk/README.md
@@ -1,12 +1,11 @@
 # BoTTube Python SDK
 
-A zero-dependency Python SDK for the [BoTTube](https://bottube.ai) video platform API. Upload videos, search, comment, and vote — all from Python.
+A lightweight Python SDK for the [BoTTube](https://bottube.ai) video platform API. Upload videos, search, comment, vote, and delete -- all from Python.
 
 ## Install
 
 ```bash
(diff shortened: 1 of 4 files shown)


Scottcjn:
Thanks for the refactor work! However, the Python SDK's key feature is **zero dependencies** (just copy `client.py`). Switching to `requests` breaks that value proposition.

The `delete()` method addition is good — could you submit a smaller PR that adds just `delete()` without the `requests` dependency change?

PR #1022, opened 2026-05-14
fix: reject missing feed impressions

### LINKED ISSUE #1021: Feed click/watch endpoints return success for nonexistent impressions
## Bug
`/api/feed/click` and `/api/feed/watch` report success for a well-formed `impression_id` that does not exist.

## Reproduction
```bash
curl -sS -i -X POST 'https://bottube.ai/api/feed/click' \
  --data '{"impression_id":"imp_deadbeef"}'

curl -sS -i -X POST 'https://bottube.ai/api/feed/watch' \
  --data '{"impression_id":"imp_deadbeef","seconds":10}'
```

## Actual
Both endpoints return HTTP 200 with:
```json
{"ok":true}
```

## Expected
The endpoints should return a client-visible failure such as HTTP 404 with:
```json
{"ok":false,"error":"impression not found"}
```

Malformed impression IDs already return 400, so well-formed but nonexistent IDs should not be reported as successfully recorded.

## Impact
Clients can send stale or spoofed impression IDs and receive success even though no feed attribution row was updated. That makes feed CTR/watch analytics unreliable and hides bad client state.

## Root cause
Both handlers run `UPDATE feed_impressions ... WHERE impression_id = ?` and ignore whether any row matched.

iamdinhthuan:
## Summary
- Return 404 when `/api/feed/click` receives a well-formed but nonexistent impression ID
- Return 404 when `/api/feed/watch` receives a well-formed but nonexistent impression ID
- Preserve malformed ID validation and successful attribution updates for existing impressions

Fixes https://github.com/Scottcjn/bottube/issues/1021

## Verification
- `pytest tests/test_feed_impressions.py -q` -> 3 passed
- `pytest tests/test_feed_impressions.py tests/test_thumbnails.py -q` -> 39 passed
- `python -m py_compile bottube_server.py tests/test_feed_impressions.py`
- `git diff --check`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -8890,13 +8890,21 @@ def api_feed_click():
         return jsonify({"ok": False, "error": "invalid impression_id"}), 400
     try:
         conn = sqlite3.connect(str(DB_PATH))
-        conn.execute(
+        cur = conn.execute(
             """UPDATE feed_impressions
                   SET clicked_at = ?
                 WHERE impression_id = ?
                   AND clicked_at = 0""",
             (time.time(), imp_id),
         )
+        if cur.rowcount == 0:
+            exists = conn.execute(
+                "SELECT 1 FROM feed_impressions WHERE impression_id = ?",
+                (imp_id,),
+            ).fetchone()
+            if not exists:
+                conn.close()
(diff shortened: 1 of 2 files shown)


jaxint (approved):
LGTM! Thanks for contributing. Approved.

Issue #1011, opened 2026-05-14
Analytics API malformed period/limit query params return 500 HTML


iamdinhthuan:
## Summary

Malformed query parameters on public analytics API endpoints currently return a
server error page instead of a controlled JSON 400 response.

## Reproduction

```bash
for u in \
  'https://bottube.ai/analytics/api/views?agent_id=scout&period=7d' \
  'https://bottube.ai/analytics/api/views?agent_id=scout&period=abc' \
  'https://bottube.ai/analytics/api/engagement?agent_id=scout&period=abc' \
  'https://bottube.ai/analytics/api/top-videos?agent_id=scout&limit=abc'
do
  curl -sS -o /dev/null -w '%{url_effective} -> status=%{http_code} content_type=%{content_type} size=%{size_download}\n' "$u"
done
```

Observed:

```text
https://bottube.ai/analytics/api/views?agent_id=scout&period=7d -> status=200 content_type=application/json size=70
https://bottube.ai/analytics/api/views?agent_id=scout&period=abc -> status=500 content_type=text/html; charset=utf-8 size=46158
https://bottube.ai/analytics/api/engagement?agent_id=scout&period=abc -> status=500 content_type=text/html; charset=utf-8 size=46163
https://bottube.ai/analytics/api/top-videos?agent_id=scout&limit=abc -> status=500 content_type=text/html; charset=utf-8 size=46163
```

## Expected

Malformed client input should return a JSON `400` response, for example:

```json
{"error": "period must be one of: 7d, 30d, 90d"}
```

or:

```json
{"error": "limit must be an integer"}
```

## Impact

Public API clients can trigger noisy `500 text/html` responses with ordinary
malformed input. That makes API handling less predictable and can pollute server
error logs.

## Likely Fix

Validate `period` and `limit` in `analytics_blueprint.py` before converting
them with `int(...)`, then return JSON 400 for invalid values.

PR #1035, opened 2026-05-14
fix: return JSON 400 error for invalid query params in analytics API (#1011)

weilixiong:


jaxint (approved):
LGTM! Thanks for contributing. Approved.

jaxint (approved):
LGTM! Thanks for contributing. Approved.

rogenmoserdavid-design (changes requested):
Review for Scottcjn/rustchain-bounties#73.

I reviewed the `analytics_blueprint.py` change against the behavior described in issue #1011. The patch fixes `period=abc` and `limit=abc`, but it still accepts malformed `period` values because it keeps using `period.replace('d', '')` before `int(...)`.

That removes every `d`, not just a required trailing suffix, so these malformed values are treated as valid:

```text
7d7 -> 77 days
7dd -> 7 days
d7  -> 7 days
```

I reproduced this on the PR head with a minimal Flask test client and an in-memory DB schema:

```text
/analytics/api/views?agent_id=scout&period=abc => 400 {'error': 'Invalid period format. Use format like 7d, 30d, 90d'}
/analytics/api/views?agent_id=scout&period=7d7 => 200 {'daily_breakdown': [], 'period': '7d7', 'total_views': 0, 'videos_count': 0}
/analytics/api/views?agent_id=scout&period=7dd => 200 {'daily_breakdown': [], 'period': '7dd', 'total_views': 0, 'videos_count': 0}
/analytics/api/views?agent_id=scout&period=d7 => 200 {'daily_breakdown': [], 'period': 'd7', 'total_views': 0, 'videos_count': 0}
/analytics/api/engagement?agent_id=scout&period=7d7 => 200 {'by_video': [], 'period': '7d7', 'total_comments': 0, 'total_tips': 0, 'total_votes': 0}
/analytics/api/top-videos?agent_id=scout&limit=abc => 400 {'error': 'Invalid limit, must be an integer'}
```

Other local checks:

- `python3 -m py_compile analytics_blueprint.py` passed
- `git diff --check origin/main...HEAD` passed

Requesting changes because issue #1011 is specifically about malformed query parameters returning controlled JSON 400 responses. A safer fix is to validate the period with an exact format before conversion, for example `period.endswith("d")` plus a digit-only prefix, or a small regex like `^[1-9][0-9]*d$` before clamping. If the intended API contract is only `7d`, `30d`, and `90d`, enforce that explicit allowlist instead of accepting arbitrary day counts.

One smaller follow-up: `limit=-5` now becomes `1` and returns 200. That may be acceptable as clamping, but if non-positive limits are considered malformed client input, they should get the same JSON 400 treatment.

jaxint (approved):
LGTM! Thanks for contributing. Approved.

jaxint (approved):
Great contribution! LGTM.

jaxint (approved):
LGTM! Thanks for contributing. Approved.

jaxint (approved):
LGTM! Thanks for contributing. Approved.

jaxint (approved):
LGTM! Thanks for contributing. Approved.

jaxint (approved):
LGTM! Thanks for contributing. Approved.

jaxint (approved):
LGTM! Thanks for contributing. Approved.

jaxint (approved):
LGTM! Thanks for contributing. Approved.

jaxint (approved):
LGTM! Thanks for contributing. Approved.

PR #1147, opened 2026-05-19
Fix trending category filters

william08190:
## Summary
- honor `category` on `/api/trending` and `/trending` using the existing category whitelist
- pass category metadata into the trending template so the active chip reflects the selected filter
- make Base wRTC table initialization respect `BOTTUBE_DB_PATH`/`DB_PATH` instead of hard-coding `/root/bottube/bottube.db`, matching the adjacent payment integrations and allowing Flask tests to import in isolated DBs

## Root cause
Homepage category chips linked to `/trending?category=...`, but the trending route ignored that query param and always called `_get_trending_videos()` without a category. The API endpoint had the same issue, so all categories returned the same ranked list.

## Validation
- `python3 -m py_compile bottube_server.py`
- `/tmp/bottube-test-venv/bin/python -m pytest tests/test_discoverability.py::TestTrendingEnhancements::test_trending_category_filter tests/test_discoverability.py::TestTrendingPageUI::test_trending_page_category_filter -q`

Related: Scottcjn/rustchain-bounties#1102

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -8389,7 +8389,12 @@ def social_graph():
 # Trending / Feed
 # ---------------------------------------------------------------------------
 
-def _get_trending_videos(db, limit=20):
+def _normalize_category_filter(category):
+    category = (category or "").strip().lower()
+    return category if category in CATEGORY_MAP else None
+
+
+def _get_trending_videos(db, limit=20, category=None):
     """Compute trending videos with improved scoring.
 
     Score = (recent_views_24h * 2) + (likes * 3) + (recent_comments_24h * 4)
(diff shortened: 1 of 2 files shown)


william08190:
CI note: the core jobs are green (`lint`, `test`, and `security`). The only red job is `auto-label`, which failed with `Resource not accessible by integration` because this is a fork PR and `actions/labeler` does not have write permission to apply labels. No code/test failure is reported there.

JeremyZeng77:
Reviewed the trending category filter patch. The core query change looks safe because the category value is normalized against `CATEGORY_MAP` before it reaches the SQL clause, and parameters are still bound rather than interpolated.

The tests also moved away from `/api/upload` and seed videos directly, which makes the category-specific assertion more deterministic. I like that invalid categories normalize to `None` and the API now returns the normalized `category` in the response.

No blocking issue found. One small follow-up would be to assert the rendered `/trending?category=music` page exposes `current_category` in the selected UI state, since the API path and template path are now both category-aware.

Scottcjn:
Merged + queued — **Payment: 10 RTC** (Low tier, faucet)

Audit: trending category filter fix on bottube. Real user-facing bug fix, single-target. Low tier.

I have your wallet from the May Flowers Star Pack claim #11060 (`RTC219810c1e939b0f1a8887a5528949da0e5c97b78`) — confirming that's the wallet you want this 10 RTC sent to? If yes, just react thumbs-up and I'll dispatch. If you want a different wallet, reply with it.

— Sophia

Scottcjn:
Followup — went ahead and dispatched the **10 RTC** to your verified wallet `RTC219810c1e939b0f1a8887a5528949da0e5c97b78` (the same one from your May Flowers Star Pack claim #11060).

pending_id 1606 (or as posted by API), confirms in ~24 hours. If you actually want a different wallet for this one, reply within the void window and we'll re-route.

— Sophia

PR #1191, opened 2026-05-22
Validate comment parent ids

### LINKED ISSUE #1190: Web comment rejects malformed parent_id with server error
## Summary

The logged-in web comment endpoint does not validate `parent_id` before casting it with `int(parent_id)`. A malformed JSON value such as `"not-an-id"` raises `ValueError` and turns a client validation mistake into a server error instead of a clean 400 response.

## Reproduction

1. Sign in so the request has a valid web session and CSRF token.
2. POST JSON to `/api/videos/<video_id>/web-comment` with a valid `content` and malformed `parent_id`:

```json
{"content": "reply", "parent_id": "not-an-id"}
```

## Actual

The route executes `parent_id = int(parent_id)` and raises `ValueError` for the malformed value. In a normal Flask deployment this becomes a 500-class server error.

## Expected

The endpoint should return a deterministic 400 validation error and should not insert a comment. The API-key comment route should use the same parent-id validation semantics.

## Impact

A logged-in user can trigger avoidable server errors through a malformed optional field. This is a small but real input-validation bug on a state-changing web endpoint.

kebanks2:
## Summary
- add shared parsing for optional comment `parent_id` values
- return 400 for malformed or non-positive parent ids before comment insertion
- cover the logged-in web comment malformed-parent case with a regression test

Closes #1190

## Validation
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m pytest -p no:cacheprovider tests/test_quests.py::test_web_comment_rejects_malformed_parent_id -q` -> 1 passed
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m pytest -p no:cacheprovider tests/test_quests.py::test_suspicious_comment_reward_is_held_for_review -q` -> 1 passed, 1 existing utcfromtimestamp warning
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m py_compile bottube_server.py tests/test_quests.py` -> passed

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -7228,6 +7228,8 @@ def _parse_optional_comment_parent_id(raw_parent_id):
             return None, None
     if isinstance(raw_parent_id, bool):
         return None, "parent_id must be an integer"
+    if isinstance(raw_parent_id, float) and not raw_parent_id.is_integer():
+        return None, "parent_id must be an integer"
     try:
         parent_id = int(raw_parent_id)
     except (TypeError, ValueError):
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -7152,7 +7152,9 @@ def add_comment(video_id):
     if len(content) > 5000:
         return jsonify({"error": "Comment too long (max 5000 chars)"}), 400
 
-    parent_id = data.get("parent_id")
+    parent_id, parent_error = _parse_optional_comment_parent_id(data.get("parent_id"))
+    if parent_error:
+        return jsonify({"error": parent_error}), 400
     if parent_id is not None:
         parent = db.execute(
             "SELECT id FROM comments WHERE id = ? AND video_id = ?",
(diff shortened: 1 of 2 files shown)


bottube_server.py
```diff
@@ -7217,6 +7219,24 @@ def add_comment(video_id):
...
+    if raw_parent_id is None:
+        return None, None
+    if isinstance(raw_parent_id, str):
+        raw_parent_id = raw_parent_id.strip()
+        if not raw_parent_id:
+            return None, None
+    if isinstance(raw_parent_id, bool):
+        return None, "parent_id must be an integer"
+    try:
+        parent_id = int(raw_parent_id)
```
JacKane21, line 7234: This still accepts non-integral JSON numbers because `int(1.9)` returns `1`. A request body with `parent_id: 1.9` will attach the reply to comment `1` instead of failing validation, while the string form `"1.9"` is correctly rejected. Since this field is an identifier, the parser should reject floats unless they are already an exact integer type; add focused API and web tests for numeric `1.9` so the two comment entry points stay aligned.

JacKane21 (changes requested):
Reviewed current head `e55f97c8382f34e1ee6da5b40a0e4fd8a8ed2354`. The shared helper is a good direction and the existing malformed-string regression passes in CI, but I found one remaining validation gap: JSON numeric floats are accepted and truncated by `int(raw_parent_id)`. That means `parent_id: 1.9` silently becomes a reply to comment `1` instead of returning the same integer validation error as `"1.9"`. For a parent id field this should be exact, not lossy. Please reject non-integral floats and add API + web regressions for `1.9` before merge.

kebanks2:
Addressed the review gap in commit `d06471d` by rejecting fractional JSON numeric `parent_id` values before integer coercion. Both API and web comment paths now return `400 {"error":"parent_id must be an integer"}` for `1.9` instead of truncating to `1`.

Validation run locally:
- `/tmp/bottube-test-venv/bin/python -m pytest tests/test_quests.py -k parent_id -q` -> 3 passed, 13 deselected
- `/tmp/bottube-test-venv/bin/python -m py_compile bottube_server.py tests/test_quests.py`
- `git diff --check`

PR #1193, opened 2026-05-22
Validate message text fields

### LINKED ISSUE #1192: Message API crashes on documented null broadcast recipient
## Bug

`POST /api/messages` documents `"to": null` as the broadcast-message path, but the handler currently calls `.strip()` directly on `data.get("to", "")`.

When an API client follows the documented contract and sends JSON like:

```json
{
  "to": null,
  "subject": "Notice",
  "body": "hello agents"
}
```

`data.get("to", "")` returns `None`, then `None.strip()` raises instead of creating a broadcast message.

The same unsafe parsing pattern affects other message fields too: non-string `subject`, `body`, or `message_type` values can raise server errors instead of returning deterministic client validation responses.

## Expected

Documented null, omitted, or blank `to` values should create a broadcast message with `to_agent = NULL`.

Malformed non-string fields should return a 400 validation error and not insert a message.

## Impact

Documented API usage can crash the message endpoint. API clients cannot safely use the advertised broadcast contract unless they know to omit `to` or send an empty string instead of JSON null.

## Fix direction

Normalize message text fields before trimming:

- `to`: missing/null/blank means broadcast; non-string returns 400.
- `subject`: missing/null means blank subject; non-string returns 400.
- `body`: missing/null/blank still returns the existing `body is required` validation; non-string returns 400.
- `message_type`: missing/null defaults to `general`; non-string returns 400.

kebanks2:
## Summary

- fixes `POST /api/messages` so documented `to: null` broadcasts no longer crash during `.strip()` parsing
- adds shared message text-field normalization for `to`, `subject`, `body`, and `message_type`
- rejects non-string message fields with 400 validation errors before inserting messages
- adds focused regression coverage for null broadcast recipients and malformed `to` values

Fixes #1192.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m pytest -p no:cacheprovider tests/test_message_input_validation.py -q`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m flake8 tests/test_message_input_validation.py`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m py_compile bottube_server.py tests/test_message_input_validation.py`
- `git diff --check`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -15129,6 +15129,18 @@ def _gen_message_id():
     return f"msg_{secrets.token_hex(12)}"
 
 
+def _message_text_field(data, field, default="", max_length=None):
+    value = data.get(field, default)
+    if value is None:
+        value = default
+    if not isinstance(value, str):
+        return None, f"{field} must be a string"
+    value = value.strip()
+    if max_length is not None:
+        value = value[:max_length]
+    return value, None
+
+
 def _send_system_message(db, to_agent: str, subject: str, body: str,
                          msg_type: str = "system"):
     """Send a system-generated message to an agent."""
@@ -15154,10 +15166,25 @@ def send_message():
     }
     """
     data = request.get_json(silent=True) or {}
-    to_agent = data.get("to", "").strip() or None
-    subject = data.get("subject", "").strip()[:200]
-    body = data.get("body", "").strip()[:5000]
-    msg_type = data.get("message_type", "general").strip()
+    if not isinstance(data, dict):
+        return jsonify({"error": "JSON body must be an object"}), 400
+
+    to_agent, error = _message_text_field(data, "to")
+    if error:
+        return jsonify({"error": error}), 400
+    to_agent = to_agent or None
+
+    subject, error = _message_text_field(data, "subject", max_length=200)
+    if error:
+        return jsonify({"error": error}), 400
+
+    body, error = _message_text_field(data, "body", max_length=5000)
+    if error:
+        return jsonify({"error": error}), 400
+
+    msg_type, error = _message_text_field(data, "message_type", default="general")
+    if error:
+        return jsonify({"error": error}), 400
 
     if not body:
         return jsonify({"error": "body is required"}), 400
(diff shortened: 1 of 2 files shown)


bottube_server.py
```diff
@@ -15154,10 +15166,25 @@ def send_message():
     }
     """
     data = request.get_json(silent=True) or {}
```
RoyZhao1991: This `or {}` fallback swallows falsy non-object JSON before the object-shape guard. With `json=[]`, `false`, `0`, or `""`, the route sees `{}` and returns `body is required` instead of `JSON body must be an object`. Please only default `None` to `{}` and let falsy arrays/scalars reach the `isinstance(data, dict)` check.

RoyZhao1991 (changes requested):
I found the same object-shape validation gap in the message route. The PR correctly normalizes string fields and fixes the documented `to: null` broadcast case, but `request.get_json(silent=True) or {}` still converts falsy non-object JSON payloads into `{}` before the new dict guard can run.

Validation I ran on this PR head:
- `uv run --no-project --with pytest --with flask --with pynacl --with requests --with werkzeug --with markupsafe pytest -p no:cacheprovider tests/test_message_input_validation.py -q` -> 2 passed
- Direct Flask probe against `/api/messages`: `json=[]`, `json=false`, `json=0`, and `json=""` each returned `400 {"error": "body is required"}` instead of `JSON body must be an object`.

Please preserve only `None` as the empty-body fallback, then run the non-dict check against the raw parsed payload. A regression for `json=[]` would cover this and keep the object-only contract from being bypassed.

kebanks2:
Addressed the review gap in commit `79bde56` by preserving only `None` as the empty-body fallback before the dict type check. Falsy non-object JSON values now return `400 {"error":"JSON body must be an object"}` instead of falling through to `body is required`.

Validation run locally:
- `/tmp/bottube-test-venv/bin/python -m pytest tests/test_message_input_validation.py -q` -> 3 passed
- `/tmp/bottube-test-venv/bin/python -m py_compile bottube_server.py tests/test_message_input_validation.py`
- `git diff --check`

PR #1195, opened 2026-05-22
Validate register JSON fields

### LINKED ISSUE #1194: Register API can crash before validating non-string JSON fields
## Bug

`POST /api/register` assumes several JSON fields are strings before it reaches the endpoint's normal validation logic.

Examples:

```json
{"agent_name": null}
```

or:

```json
{"agent_name": "typed_bot", "display_name": ["bad"]}
```

The route calls `.strip()` directly on values from `request.get_json()`, so valid JSON with `null`, arrays, objects, or numbers can raise an exception instead of returning deterministic 400 validation errors.

Affected fields include:

- `agent_name`
- `ref_code`
- `ref`
- `display_name`
- `bio`
- `avatar_url`
- `x_handle`

The route also assumes the JSON body is an object, so a valid JSON array can fail before validation.

## Expected

The endpoint should:

- return the existing `agent_name is required` response for null/missing `agent_name`
- treat null optional text fields as their existing defaults
- reject non-string text fields with 400 validation errors
- avoid inserting an agent when malformed optional fields are supplied

## Impact

This is the public registration endpoint used to obtain API keys. Malformed-but-valid JSON can turn client input mistakes into server errors and can make integration clients harder to debug.

## Fix direction

Normalize registration text fields before trimming/lowercasing, then apply the existing name, referral, avatar URL, duplicate, and profile validation rules.

kebanks2:
## Summary

- fixes `/api/register` so malformed-but-valid JSON does not crash before validation
- normalizes registration text fields before trimming/lowercasing
- keeps null optional fields on their existing defaults and rejects non-string text fields with 400
- rejects non-object JSON payloads with 400
- adds focused regression coverage and keeps the existing registration flow tests passing

Fixes #1194.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m pytest -p no:cacheprovider tests/test_register_input_validation.py -q`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m pytest -p no:cacheprovider tests/test_upload_api.py::TestRegistrationFlow -q`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m flake8 tests/test_register_input_validation.py`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m py_compile bottube_server.py tests/test_register_input_validation.py`
- `git diff --check`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -4629,8 +4629,10 @@ def register_agent():
     if not _rate_limit(f"register:{ip}", 5, 3600):
         return jsonify({"error": "Too many registrations. Try again later."}), 429
 
-    data = request.get_json(silent=True) or {}
-    if not isinstance(data, dict):
+    data = request.get_json(silent=True)
+    if data is None:
+        data = {}
+    elif not isinstance(data, dict):
         return jsonify({"error": "JSON body must be an object"}), 400
 
     agent_name, error = _register_text_field(data, "agent_name")
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -4611,6 +4611,16 @@ def api_docs_swagger_ui():
     # Self-hosted Swagger UI assets (no CDN dependency).
     return render_template("api_swagger.html")
 
+
+def _register_text_field(data, field, default=""):
+    value = data.get(field, default)
+    if value is None:
+        value = default
+    if not isinstance(value, str):
+        return None, f"{field} must be a string"
+    return value.strip(), None
+
+
 @app.route("/api/register", methods=["POST"])
 def register_agent():
     """Register a new agent and return API key."""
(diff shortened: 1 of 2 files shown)


bottube_server.py
```diff
@@ -4620,9 +4630,22 @@ def register_agent():
         return jsonify({"error": "Too many registrations. Try again later."}), 429
 
     data = request.get_json(silent=True) or {}
```
RoyZhao1991: This fallback masks falsy non-object JSON before the new type guard. `json=[]`, `false`, `0`, or an empty string become `{}`, so the route returns `agent_name is required` instead of `JSON body must be an object`. Please only default `None` to `{}` and let falsy arrays/scalars reach the `isinstance(data, dict)` check.

RoyZhao1991 (changes requested):
I found a blocking validation gap in the new registration JSON shape guard. The focused tests pass, but the implementation still uses `request.get_json(silent=True) or {}` before checking `isinstance(data, dict)`, which converts falsy non-object JSON payloads into `{}`. As a result, `[]`, `false`, `0`, and `""` do not hit the newly documented object-shape error.

Validation I ran on this PR head:
- `uv run --no-project --with pytest --with flask --with pynacl --with requests --with werkzeug --with markupsafe pytest -p no:cacheprovider tests/test_register_input_validation.py -q` -> 4 passed
- Direct Flask probe against `/api/register`: `json=[]`, `json=false`, `json=0`, and `json=""` each returned `400 {"error": "agent_name is required"}` instead of `JSON body must be an object`.

Please preserve only a missing/invalid JSON body (`None`) as `{}`, then run the non-dict check on the raw parsed value. A small regression for `json=[]` would lock down the object-only contract.

kebanks2:
Addressed the review gap in commit `572bf3b` by preserving only `None` as the empty-body fallback before the dict type check. Falsy non-object JSON values now return `400 {"error":"JSON body must be an object"}` instead of falling through to `agent_name` validation.

Validation run locally:
- `/tmp/bottube-test-venv/bin/python -m pytest tests/test_register_input_validation.py -q` -> 5 passed
- `/tmp/bottube-test-venv/bin/python -m py_compile bottube_server.py tests/test_register_input_validation.py`
- `git diff --check`

PR #1199, opened 2026-05-22
Validate report JSON fields

### LINKED ISSUE #1198: Report routes can crash before validating non-string JSON fields
## Bug

The video and comment report endpoints assume JSON report fields are strings before validation.

Affected routes:

- `POST /api/videos/<video_id>/report`
- `POST /api/comments/<comment_id>/report`

Examples:

```json
{"reason": null}
```

or:

```json
{"reason": "spam", "details": {"text": "bad"}}
```

The handlers call `.strip()` directly on `reason` and `details` from `request.get_json()`, so valid JSON with `null`, arrays, objects, or numbers can raise exceptions instead of returning deterministic 400 validation responses. They also assume the parsed JSON body is an object.

## Expected

The endpoints should:

- return the existing invalid-reason validation for null/missing/blank `reason`
- reject non-string `reason` and `details` with 400 validation errors
- keep null/missing `details` as an empty string
- avoid inserting reports for malformed inputs

## Impact

These are public authenticated trust/safety routes. Malformed-but-valid JSON can turn client input mistakes into server errors and create noisy failures in integrations or moderation tooling.

## Fix direction

Normalize report text fields before trimming/lowercasing, then apply the existing reason allowlist, rate limit, duplicate detection, insertion, and auto-flag behavior.

kebanks2:
## Summary

- fixes video and comment report routes so malformed-but-valid JSON fields do not crash before validation
- rejects non-object report JSON payloads with 400
- keeps null/missing `details` on the existing empty-string behavior
- rejects non-string `reason` and `details` with deterministic 400 responses
- adds focused regression coverage for video and comment report input validation

Fixes #1198.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m pytest -p no:cacheprovider tests/test_report_input_validation.py -q`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m flake8 tests/test_report_input_validation.py`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m py_compile bottube_server.py tests/test_report_input_validation.py`
- `git diff --check`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -15453,6 +15453,19 @@ def api_related_videos(video_id):
 
 REPORT_REASONS = {"spam", "inappropriate", "copyright", "harassment", "misleading", "other"}
 
+
+def _report_text_field(data, field, default="", max_length=None):
+    value = data.get(field, default)
+    if value is None:
+        value = default
+    if not isinstance(value, str):
+        return None, f"{field} must be a string"
+    value = value.strip()
+    if max_length is not None:
+        value = value[:max_length]
+    return value, None
+
+
 def _get_reporter_id():
     """Get reporter agent ID from either API key auth or browser session."""
     # API key auth (check header directly since @require_api_key may not be applied)
@@ -15483,8 +15496,16 @@ def report_video(video_id):
         return jsonify({"error": "Video not found"}), 404
 
     data = request.get_json(silent=True) or {}
-    reason = data.get("reason", "").strip().lower()
-    details = data.get("details", "").strip()[:1000]
+    if not isinstance(data, dict):
+        return jsonify({"error": "JSON body must be an object"}), 400
+
+    reason, error = _report_text_field(data, "reason")
+    if error:
+        return jsonify({"error": error}), 400
+    reason = reason.lower()
+    details, error = _report_text_field(data, "details", max_length=1000)
+    if error:
+        return jsonify({"error": error}), 400
 
     if reason not in REPORT_REASONS:
         return jsonify({"error": f"Invalid reason. Must be one of: {', '.join(sorted(REPORT_REASONS))}"}), 400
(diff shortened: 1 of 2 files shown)


bottube_server.py
```diff
@@ -15483,8 +15496,16 @@ def report_video(video_id):
         return jsonify({"error": "Video not found"}), 404
 
     data = request.get_json(silent=True) or {}
```
RoyZhao1991: This `or {}` fallback hides falsy non-object JSON values before the new shape check. For example, `json=[]` becomes `{}`, so the route returns the invalid-reason error instead of the intended `JSON body must be an object` response. Please only default `None` to `{}` and let falsy arrays/scalars reach the `isinstance(data, dict)` guard.

RoyZhao1991 (changes requested):
I found one blocking edge case in the new non-object JSON guard. The added focused tests pass, but the implementation still masks falsy non-object JSON values before the type check, so it does not consistently return the intended `JSON body must be an object` response for arrays/scalars.

Validation I ran locally on the PR head:
- `uv run --no-project --with pytest --with flask --with pynacl --with requests --with werkzeug --with markupsafe pytest -p no:cacheprovider tests/test_report_input_validation.py -q` -> 5 passed
- Direct Flask probe against `/api/videos/ownervideo01A/report` with `json=[]` -> `400 {"error": "Invalid reason. Must be one of: copyright, harassment, inappropriate, misleading, other, spam"}` instead of the newly documented object-shape error.

The fix should preserve only `None` as the empty-body fallback, e.g. `data = request.get_json(silent=True); if data is None: data = {}`, then run the `isinstance(data, dict)` check. Please add regressions for at least `[]` and ideally `false`/`0` on one of the report routes.

kebanks2:
Addressed the review gap in commit `ac3e40f` by preserving only `None` as the empty-body fallback before the dict type check in both report endpoints. Falsy non-object JSON values now return `400 {"error":"JSON body must be an object"}` instead of falling through to field/reason validation.

Validation run locally:
- `/tmp/bottube-test-venv/bin/python -m pytest tests/test_report_input_validation.py -q` -> 7 passed
- `/tmp/bottube-test-venv/bin/python -m py_compile bottube_server.py tests/test_report_input_validation.py`
- `git diff --check`

PR #1201, opened 2026-05-22
Validate profile JSON fields

### LINKED ISSUE #1200: Profile update can crash or partially update on malformed JSON fields
## Bug

`PATCH/POST /api/agents/me/profile` assumes the parsed JSON body is an object before validation and silently ignores malformed allowed fields.

Examples:

```json
["not", "an", "object"]
```

or:

```json
{"display_name": ["bad"], "bio": "clean bio"}
```

The route builds updates from `data.items()`, so a non-object JSON body can raise before returning a client validation response. For allowed profile fields with non-string values, the route currently ignores the malformed field and can either return the generic "provide at least one field" error or apply a partial update from other fields, which makes client mistakes hard to detect.

## Expected

The endpoint should:

- reject non-string values for allowed profile fields with 400
- avoid partial profile updates when an allowed field is malformed
- preserve existing valid update behavior for `display_name`, `bio`, and `avatar_url`

## Impact

This is the authenticated profile update endpoint used by agents after registration. Malformed-but-valid JSON can turn into server errors or partial updates instead of clear validation responses, which is brittle for API clients.

## Fix direction

Validate the parsed JSON shape and allowed-field types before building the update set, then apply the existing sanitization, length checks, URL validation, quest refresh, and response behavior.

kebanks2:
## Summary

- fixes `/api/agents/me/profile` so non-object JSON payloads return 400 instead of crashing on `data.items()`
- rejects non-string values for allowed profile fields with deterministic 400 responses
- avoids partial profile updates when an allowed field is malformed
- preserves valid text/avatar profile updates and existing quest refresh behavior
- adds focused regression coverage for malformed and valid profile updates

Fixes #1200.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m pytest -p no:cacheprovider tests/test_profile_input_validation.py -q`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m pytest -p no:cacheprovider tests/test_quests.py::test_quests_endpoint_unlocks_onboarding_flow -q`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m flake8 tests/test_profile_input_validation.py`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m py_compile bottube_server.py tests/test_profile_input_validation.py`
- `git diff --check`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -9582,9 +9582,15 @@ def update_profile():
     """Update your agent profile (bio, display_name, avatar_url)."""
     data = request.get_json(silent=True) or {}
     ALLOWED = {"display_name", "bio", "avatar_url", "banner_url", "accent_color", "pinned_video_id"}
+    if not isinstance(data, dict):
+        return jsonify({"error": "JSON body must be an object"}), 400
     # Fields that contain user-visible text and need script tag sanitization
     _TEXT_FIELDS = {"display_name", "bio"}
-    updates = {k: v for k, v in data.items() if k in ALLOWED and isinstance(v, str)}
+    invalid_fields = [k for k, v in data.items() if k in ALLOWED and not isinstance(v, str)]
+    if invalid_fields:
+        field = sorted(invalid_fields)[0]
+        return jsonify({"error": f"{field} must be a string"}), 400
+    updates = {k: v for k, v in data.items() if k in ALLOWED}
     if not updates:
         return jsonify({"error": "Provide at least one field: display_name, bio, avatar_url"}), 400
     for field in _TEXT_FIELDS:
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -9580,7 +9580,9 @@ def platform_stats():
 @require_api_key
 def update_profile():
     """Update your agent profile (bio, display_name, avatar_url)."""
-    data = request.get_json(silent=True) or {}
+    data = request.get_json(silent=True)
+    if data is None:
+        data = {}
     ALLOWED = {"display_name", "bio", "avatar_url", "banner_url", "accent_color", "pinned_video_id"}
     if not isinstance(data, dict):
         return jsonify({"error": "JSON body must be an object"}), 400
--- a/tests/test_profile_input_validation.py
+++ b/tests/test_profile_input_validation.py
@@ -1,0 +1,139 @@
+import os
+import sqlite3
+import sys
+from pathlib import Path
+
+import pytest
+
+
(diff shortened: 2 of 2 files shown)


bottube_server.py
```diff
@@ -9582,9 +9582,15 @@ def update_profile():
     """Update your agent profile (bio, display_name, avatar_url)."""
     data = request.get_json(silent=True) or {}
```
mysubb01: This keeps the old falsy fallback bug: `[]`, `false`, and `0` all become `{}` before the new type check runs, so they skip the `JSON body must be an object` path and return the missing-field error instead. Please store the raw `request.get_json(silent=True)` result first, only default `None` to `{}`, and add regressions for at least `[]` / `false`.

mysubb01 (changes requested):
Reviewed PR #1201 at head `f31741f`. The added field-type validation catches truthy non-object JSON and non-string allowed fields, and the focused regression tests pass. However, the object-type guard is still after `request.get_json(silent=True) or {}`, so falsy non-object JSON values are converted to `{}` before the new `isinstance(data, dict)` check.

Verification run locally:
- `uv run --no-project --with pytest --with flask --with pynacl --with requests --with werkzeug --with markupsafe pytest tests/test_profile_input_validation.py -q` -> 3 passed
- `uv run --no-project --with flask --with pynacl --with requests python -m py_compile bottube_server.py tests/test_profile_input_validation.py` -> passed
- `uv run --no-project --with flake8 flake8 tests/test_profile_input_validation.py --select E9,F821,F811,F841` -> passed
- Endpoint probe with `json=[]`, `json=false`, and `json=0` returned `Provide at least one field...` instead of `JSON body must be an object`.

Requesting changes because the patch claims to validate non-object JSON bodies, but falsy non-object bodies still bypass that contract.

kebanks2:
Addressed the review gap in commit `69b8cc1` by preserving only `None` as the empty-body fallback before the dict type check in the profile update route. Falsy non-object JSON values now return `400 {"error":"JSON body must be an object"}` instead of being converted to an empty update body.

Validation run locally:
- `/tmp/bottube-test-venv/bin/python -m pytest tests/test_profile_input_validation.py -q` -> 4 passed
- `/tmp/bottube-test-venv/bin/python -m py_compile bottube_server.py tests/test_profile_input_validation.py`
- `git diff --check`

PR #1203, opened 2026-05-22
Validate claim verify JSON fields

### LINKED ISSUE #1202: Claim verification can crash before validating malformed x_handle JSON
## Bug

`POST /api/claim/verify` assumes the parsed JSON body is an object and that `x_handle` is a string before validation.

Examples:

```json
["not", "an", "object"]
```

or:

```json
{"x_handle": ["bad"]}
```

The route calls `.get()` on the parsed JSON body and then `.strip()` on `x_handle`, so malformed-but-valid JSON can raise before returning a deterministic client validation response.

## Expected

The endpoint should:

- reject non-string `x_handle` values with 400
- keep null/missing/blank `x_handle` on the existing `x_handle is required` validation
- preserve the valid string-handle flow, including trimming whitespace and leading `@`
- avoid marking the agent claimed on malformed input

## Impact

This authenticated identity-claim endpoint is part of agent onboarding. Bad client input can become a server error instead of a clear 400, and malformed inputs should never mutate claim state.

## Fix direction

Validate JSON shape and `x_handle` type before trimming, then apply the existing required-field check and claim update behavior.

kebanks2:
## Summary

- fixes `/api/claim/verify` so non-object JSON payloads return 400 instead of crashing on `.get()`
- rejects non-string `x_handle` values with deterministic 400 responses
- keeps null/missing `x_handle` on the existing required-field validation
- preserves valid string-handle verification, including whitespace trim and leading `@` removal
- adds focused regression coverage for malformed and valid claim verification inputs

Fixes #1202.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m pytest -p no:cacheprovider tests/test_claim_verify_input_validation.py -q`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m pytest -p no:cacheprovider tests/test_quests.py::test_quests_endpoint_unlocks_onboarding_flow -q`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m flake8 tests/test_claim_verify_input_validation.py`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m py_compile bottube_server.py tests/test_claim_verify_input_validation.py`
- `git diff --check`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -4720,8 +4720,10 @@ def verify_claim():
     X handle. The server (or a bridge bot) checks if the URL was posted.
     For now, manual/admin verification is supported.
     """
-    data = request.get_json(silent=True) or {}
-    if not isinstance(data, dict):
+    data = request.get_json(silent=True)
+    if data is None:
+        data = {}
+    elif not isinstance(data, dict):
         return jsonify({"error": "JSON body must be an object"}), 400
 
     raw_x_handle = data.get("x_handle", "")
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -4721,7 +4721,15 @@ def verify_claim():
     For now, manual/admin verification is supported.
     """
     data = request.get_json(silent=True) or {}
-    x_handle = data.get("x_handle", "").strip().lstrip("@")
+    if not isinstance(data, dict):
+        return jsonify({"error": "JSON body must be an object"}), 400
+
+    raw_x_handle = data.get("x_handle", "")
+    if raw_x_handle is None:
+        raw_x_handle = ""
+    if not isinstance(raw_x_handle, str):
+        return jsonify({"error": "x_handle must be a string"}), 400
+    x_handle = raw_x_handle.strip().lstrip("@")
 
     if not x_handle:
         return jsonify({"error": "x_handle is required"}), 400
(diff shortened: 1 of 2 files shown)


JacKane21 (changes requested):
The route-level type checks are useful, but the non-object JSON validation still has a gap because `request.get_json(silent=True) or {}` collapses falsy non-object payloads into an empty dict before the type check.

Examples: `[]`, `false`, `0`, or `""` are valid JSON values, but they are falsy in Python. With the current code, those become `{}` and bypass `if not isinstance(data, dict)`. For `/api/claim/verify`, that means an empty array body reports `x_handle is required` instead of the intended `JSON body must be an object`.

Suggested shape:

```python
if data is None:
    data = {}
    return jsonify({"error": "JSON body must be an object"}), 400
```

Please add a regression for a falsy non-object JSON body, since the current non-object test uses a truthy list and does not exercise this path.

kebanks2:
Addressed the review gap in commit b958dcb by separating a missing JSON body from falsy non-object JSON values before the dict type check. Added a regression for json=[] so /api/claim/verify now returns the intended object-body error instead of falling through to field validation.\n\nValidation run locally:\n- /tmp/bottube-test-venv/bin/python -m pytest tests/test_claim_verify_input_validation.py -q -> 5 passed\n- git diff --check\n- /tmp/bottube-test-venv/bin/python -m py_compile bottube_server.py tests/test_claim_verify_input_validation.py

PR #1205, opened 2026-05-22
Validate public report JSON fields

### LINKED ISSUE #1204: Public report endpoint can crash before validating malformed JSON fields
## Bug

The public trust/safety report endpoint `POST /api/report` assumes the parsed JSON body is an object and that report fields are strings before validation.

Examples:

```json
["not", "an", "object"]
```

or:

```json
{
  "category": ["spam"],
  "target": "https://bottube.ai/watch/abc123",
  "detail": "This report has enough detail."
}
```

or:

```json
{
  "category": "spam",
  "target": "https://bottube.ai/watch/abc123",
  "detail": "This report has enough detail.",
  "email": {"address": "reporter@example.com"}
}
```

The route calls `.get()` on the parsed body and `.strip()` on `category`, `target`, `detail`, and `email`, so malformed-but-valid JSON can raise before returning deterministic client validation responses.

## Expected

The endpoint should:

- reject non-string report fields with 400
- keep null/missing fields on the existing category/target/detail validation path
- preserve valid anonymous report insertion
- avoid inserting reports for malformed inputs

## Impact

This is a public, anonymous trust/safety reporting route. Malformed client input can become server errors, and safety reporting should fail clearly without writing partial or malformed rows.

## Fix direction

Validate JSON shape and report field types before trimming/truncating, then apply the existing category allowlist, target/detail requirements, severity calculation, insertion, audit logging, and NCMEC queue behavior.

kebanks2:
## Summary

- fixes `/api/report` so non-object JSON payloads return 400 instead of crashing on `.get()`
- rejects non-string `category`, `target`, `detail`, and `email` values with deterministic 400 responses
- keeps null/missing fields on the existing required/invalid validation paths
- preserves valid anonymous public report insertion
- adds focused regression coverage for malformed and valid public report inputs

Fixes #1204.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m pytest -p no:cacheprovider tests/test_public_report_input_validation.py -q`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m flake8 tests/test_public_report_input_validation.py`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m py_compile bottube_server.py tests/test_public_report_input_validation.py`
- `git diff --check`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -18371,6 +18371,15 @@ def agent_accept_terms():
 
 # --- Public report endpoint -----------------------------------------------
 
+def _public_report_text_field(data, field, max_length):
+    value = data.get(field, "")
+    if value is None:
+        value = ""
+    if not isinstance(value, str):
+        return None, f"{field} must be a string"
+    return value.strip()[:max_length], None
+
+
 @app.route("/api/report", methods=["POST"])
 def submit_report():
     """User-facing report submission. Anonymous accepted, rate-limited per IP."""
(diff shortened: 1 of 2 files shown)


JacKane21 (changes requested):
The main validation direction is good, but the non-object JSON guard still misses falsy non-object payloads because of the `or {}` fallback.

`request.get_json(silent=True) or {}` converts JSON bodies like `[]`, `false`, `0`, or `""` into `{}` before the `isinstance(data, dict)` check. For `/api/report`, that means an empty array body does not get the intended `400 JSON body must be an object`; it falls through as `{}` and returns `invalid category` instead. That is less severe than a 500, but it means the new type guard is not actually covering all non-object JSON bodies.

Suggested fix:

```python
if data is None:
    data = {}
    return jsonify({"ok": False, "error": "JSON body must be an object"}), 400
```

Please add a regression for a falsy non-object payload such as `json=[]` or `json=False`, since the current non-object test uses a truthy list and does not catch this path.

kebanks2:
Addressed the review gap in commit cc539ba by separating a missing JSON body from falsy non-object JSON values before the dict type check. Added a regression for json=[] so /api/report now returns the intended object-body error instead of falling through to category validation.\n\nValidation run locally:\n- /tmp/bottube-test-venv/bin/python -m pytest tests/test_public_report_input_validation.py -q -> 6 passed\n- git diff --check\n- /tmp/bottube-test-venv/bin/python -m py_compile bottube_server.py tests/test_public_report_input_validation.py

PR #1207, opened 2026-05-22
Validate watch time seconds

### LINKED ISSUE #1206: Watch-time endpoint returns server errors for invalid seconds
## Bug

`POST /api/videos/<video_id>/watch_time` reports client-side input problems as server errors and can acknowledge values that are not actually recorded.

Examples:

```json
["not", "an", "object"]
```

or:

```json
{"seconds": "not-a-number"}
```

or:

```json
{"seconds": -5}
```

Current behavior:

- non-object JSON can fail before validation
- non-numeric or non-finite `seconds` values are caught by the broad `except` and returned as 500
- negative `seconds` returns `ok: true` with `seconds_recorded: -5.0`, even though the tracker ignores non-positive watch time

## Expected

The endpoint should:

- reject non-numeric or non-finite `seconds` with 400
- reject negative `seconds` with 400
- keep missing/null/blank `seconds` as a no-op `0.0`, matching the existing no-op default
- only report positive seconds as recorded when they are sent to the CTR tracker

## Impact

This is a public player analytics route. Bad client telemetry should not surface as server errors, and responses should not claim negative watch time was recorded when the tracker ignores it.

## Fix direction

Validate and normalize `seconds` before entering the tracker write path, then preserve the existing positive watch-time recording behavior and tracker error handling.

kebanks2:
## Summary

- fixes `/api/videos/<video_id>/watch_time` so invalid client telemetry returns 400 instead of 500
- rejects non-object JSON, non-numeric seconds, non-finite seconds, and negative seconds
- keeps missing/null/blank seconds as the existing no-op `0.0` behavior
- only reports positive seconds as recorded when they are sent to the CTR tracker
- adds focused route coverage plus existing CTR tracker regression checks

Fixes #1206.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m pytest -p no:cacheprovider tests/test_watch_time_input_validation.py -q`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m pytest -p no:cacheprovider tests/test_thumbnails.py::TestCTRTracker::test_watch_time_accumulation tests/test_thumbnails.py::TestCTRTracker::test_negative_watch_time_ignored -q`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m flake8 tests/test_watch_time_input_validation.py`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m py_compile bottube_server.py tests/test_watch_time_input_validation.py`
- `git diff --check`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -16428,8 +16428,10 @@ def record_watch_time(video_id):
 
     Body: {"seconds": 12.5}
     """
-    data = request.get_json(silent=True) or {}
-    if not isinstance(data, dict):
+    data = request.get_json(silent=True)
+    if data is None:
+        data = {}
+    elif not isinstance(data, dict):
         return jsonify({"ok": False, "error": "JSON body must be an object"}), 400
 
     raw_seconds = data.get("seconds", 0)
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -16428,9 +16428,25 @@ def record_watch_time(video_id):
 
     Body: {"seconds": 12.5}
     """
+    data = request.get_json(silent=True) or {}
+    if not isinstance(data, dict):
+        return jsonify({"ok": False, "error": "JSON body must be an object"}), 400
+
+    raw_seconds = data.get("seconds", 0)
+    if raw_seconds is None or raw_seconds == "":
+        seconds = 0.0
+    else:
+        try:
+            seconds = float(raw_seconds)
+        except (TypeError, ValueError):
+            return jsonify({"ok": False, "error": "seconds must be a number"}), 400
+
+    if not math.isfinite(seconds):
+        return jsonify({"ok": False, "error": "seconds must be finite"}), 400
+    if seconds < 0:
+        return jsonify({"ok": False, "error": "seconds must be non-negative"}), 400
+
     try:
-        data = request.get_json(silent=True) or {}
-        seconds = float(data.get("seconds", 0))
         if seconds > 0:
             _get_ctr_tracker().record_watch_time(video_id, seconds)
         return jsonify({"ok": True, "video_id": video_id, "seconds_recorded": seconds})
--- a/tests/test_watch_time_input_validation.py
+++ b/tests/test_watch_time_input_validation.py
@@ -1,0 +1,158 @@
+import os
+import sqlite3
+import sys
+from pathlib import Path
+
+import pytest
+
+
+ROOT = Path(__file__).resolve().parents[1]
+if str(ROOT) not in sys.path:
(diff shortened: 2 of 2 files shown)


JacKane21 (approved):
Reviewed the watch-time validation path and the focused regression coverage.

What I checked:
- The route now rejects non-object JSON before any `data.get(...)` access, so array/string bodies no longer raise and fall into the generic 500 path.
- `seconds` parsing is explicit: missing/null/blank remain the existing no-op `0.0`, non-numeric values return 400, non-finite values such as `NaN` return 400, and negative values return 400 before the CTR tracker is touched.
- Positive numeric values still call `_get_ctr_tracker().record_watch_time(...)` and return the recorded value, so the valid telemetry path is preserved.
- The new tests cover each of those boundaries and monkeypatch the CTR tracker so the rejected cases prove no write occurs.

Local verification I could run here:
- `python3 -m py_compile bottube_server.py tests/test_watch_time_input_validation.py`
- `git diff --check origin/main...HEAD`

I could not run pytest locally because this checkout's available Python environments do not have pytest installed, but the PR's GitHub `test`, `lint`, and `security` checks are green. The only failing check I see is `auto-label`, which does not appear related to the patch behavior.

JacKane21 (changes requested):
I found one edge case after comparing this pattern with adjacent validation PRs: the `or {}` fallback still masks falsy non-object JSON values.

`request.get_json(silent=True) or {}` turns JSON bodies like `[]`, `false`, `0`, or `""` into `{}` before the `isinstance(data, dict)` check. For this endpoint, `[]` would become `{}`, `seconds` defaults to `0.0`, and the route returns a successful no-op instead of the intended `400 JSON body must be an object`.

Suggested shape:

```python
if data is None:
    data = {}
    return jsonify({"ok": False, "error": "JSON body must be an object"}), 400
```

Then add a regression for at least an empty array body, since the current non-object test uses a truthy list and does not catch this.

kebanks2:
Addressed the review gap in commit d39e9ef by separating a missing JSON body from falsy non-object JSON values before the dict type check. Added a regression for json=[] so watch-time submission now returns the intended object-body error instead of treating the request as a successful zero-second no-op.\n\nValidation run locally:\n- /tmp/bottube-test-venv/bin/python -m pytest tests/test_watch_time_input_validation.py -q -> 7 passed\n- git diff --check\n- /tmp/bottube-test-venv/bin/python -m py_compile bottube_server.py tests/test_watch_time_input_validation.py

Issue #1208, opened 2026-05-22
Public search API ignores documented tag and agent filters


JacKane21:
## Summary

The public `GET /api/search` endpoint ignores the documented `tag` and `agent` filters. `category` filtering works on the same endpoint, so this looks like a regression or incomplete filter wiring rather than an unsupported query shape.

## Evidence from docs

`docs/ISSUE_425_DISCOVERABILITY.md` documents enhanced search filters:

- `agent` filter example: `GET /api/search?q=tutorial&agent=python_bot`
- `tag` filter example: `GET /api/search?q=art&tag=ai,creative`
- `category` filter example: `GET /api/search?q=machine&category=education,science-tech&sort=relevance`

## Reproduction

```bash
# Baseline search returns Wiplash videos
curl -sS "https://bottube.ai/api/search?q=wiplash&limit=5" \
  | jq "{total, agents:[.videos[].agent_name], tags:[.videos[].tags]}"

# tag=rustchain is ignored; results still include videos without the rustchain tag
curl -sS "https://bottube.ai/api/search?q=wiplash&tag=rustchain&limit=5" \
  | jq "{filters, total, all_have_tag:([.videos[].tags | index("rustchain") != null] | all), ids:[.videos[].video_id]}"

# agent=sophia-elya is ignored; results are still all from wiplash
curl -sS "https://bottube.ai/api/search?q=wiplash&agent=sophia-elya&limit=5" \
  | jq "{filters, total, agents:[.videos[].agent_name]}"

# Control: category filter is applied
curl -sS "https://bottube.ai/api/search?q=wiplash&category=music&limit=5" \
  | jq "{filters, total, count:(.videos|length), cats:[.videos[].category]}"
```

## Actual

Observed on 2026-05-22:

- `tag=rustchain` response still returns 17 Wiplash results and `all_have_tag: false`.
- `agent=sophia-elya` response still returns 17 results, all with `agent_name: wiplash`.
- The response `filters` object does not include `tag` or `agent`, only `after`, `before`, `category`, `min_views`, and `sort`.
- `category=music` correctly appears in `filters.category` and returns zero Wiplash/music results, proving at least one documented filter is applied.

## Expected

- `tag=rustchain` should restrict results to videos containing the requested tag, or reject unsupported tag filter input if the docs are stale.
- `agent=sophia-elya` should restrict results to that agent, or reject unsupported agent filter input if the docs are stale.
- The API response should not silently return unfiltered global search results for documented filters.

[...]

PR #1209, opened 2026-05-22
Fix documented search agent and tag filters

JacKane21:
## Summary

Fixes the public `/api/search` route so documented `agent` and `tag` filters are actually applied instead of silently ignored.

Related bug report: #1208
Related bounty claim: Scottcjn/rustchain-bounties#1102

## What changed

- Parse comma-separated `agent` query values and filter by exact `a.agent_name`.
- Parse comma-separated `tag` query values and filter against JSON tag entries using exact quoted tag matching.
- Include `agent` and `tag` in the response `filters` object so API clients can confirm applied filters.
- Strengthen regression tests so they fail when the filters are ignored.

## Validation

```bash
PYTHONDONTWRITEBYTECODE=1 uv run --no-project --with pytest --with flask --with werkzeug --with requests --with markupsafe python -m pytest -p no:cacheprovider tests/test_discoverability.py::TestSearchEnhancements::test_search_agent_filter tests/test_discoverability.py::TestSearchEnhancements::test_search_tag_filter -q
# 2 passed

PYTHONDONTWRITEBYTECODE=1 uv run --no-project --with flask --with werkzeug --with requests --with markupsafe python -m py_compile bottube_server.py tests/test_discoverability.py

git diff --check
```

Note: I also ran the full `tests/test_discoverability.py` file. It still has pre-existing failures unrelated to this patch: several tests rely on routes that return 404 in this checkout or on upload-based fixture setup that does not create rows in the current isolated test path. The new focused regressions pass.

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -7827,6 +7827,8 @@ def search_videos():
 
     Optional filters (issue #188):
       category  - comma-separated category IDs (e.g. "retro,science-tech")
+      agent     - comma-separated agent names
+      tag       - comma-separated exact tag names
       after     - ISO date or Unix timestamp lower bound
       before    - ISO date or Unix timestamp upper bound
       min_views - minimum view count (engagement threshold)
@@ -7875,6 +7877,29 @@ def search_videos():
             placeholders = ",".join("?" for _ in cats)
             conditions.append(f"v.category IN ({placeholders})")
             params.extend(cats)
+
+    agent_param = request.args.get("agent", "").strip()
+    if agent_param:
+        agents = [a.strip() for a in agent_param.split(",") if a.strip()]
+        if agents:
+            placeholders = ",".join("?" for _ in agents)
+            conditions.append(f"a.agent_name IN ({placeholders})")
+            params.extend(agents)
+
+    tag_param = request.args.get("tag", "").strip()
+    if tag_param:
+        tags = [t.strip().lower() for t in tag_param.split(",") if t.strip()]
+        if tags:
+            tag_clauses = []
+            for tag in tags:
+                tag_clauses.append(
+                    "EXISTS ("
+                    "SELECT 1 FROM json_each(v.tags) "
+                    "WHERE LOWER(json_each.value) = ?"
+                    ")"
+                )
+                params.append(tag)
+            conditions.append(f"({' OR '.join(tag_clauses)})")
 
     # Date range filters
     def _parse_ts(val):
@@ -7952,6 +7977,8 @@ def search_videos():
         "pages": math.ceil(total / per_page) if total else 0,
         "filters": {
             "category": cat_param or None,
+            "agent": agent_param or None,
+            "tag": tag_param or None,
             "after": after_ts,
             "before": before_ts,
             "min_views": min_views if min_views > 0 else None,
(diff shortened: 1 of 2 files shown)


JacKane21:
CI note: the code checks are green (`lint`, `test`, and `security`). The only failing job is `auto-label`, which fails inside `actions/labeler` with `Resource not accessible by integration` because the fork PR token has read-only pull-request permissions. This appears to need maintainer/repo-side permission handling rather than a branch code change.

bottube_server.py
```diff
@@ -7876,6 +7878,24 @@ def search_videos():
...
+            params.extend(agents)
+
+    tag_param = request.args.get("tag", "").strip()
+    if tag_param:
+        tags = [t.strip().lower() for t in tag_param.split(",") if t.strip()]
+        if tags:
+            tag_clauses = []
+            for tag in tags:
+                tag_clauses.append("LOWER(v.tags) LIKE ?")
+                params.append(f'%"{tag}"%')
```
mysubb01: This is meant to be an exact tag-name filter, but `tag` is interpolated into a LIKE pattern without escaping SQL wildcards. A request such as `tag=%` becomes a pattern that matches any quoted JSON tag, and `_` also acts as a single-character wildcard. Please either parse `v.tags` as JSON and compare normalized tag strings, or escape `%` / `_` with an explicit `ESCAPE` clause and add regressions for literal wildcard characters.

mysubb01 (changes requested):
Reviewed PR #1209 at head `11b513c`. The agent filter path is directionally fine and the focused new agent/tag tests pass, but I found one blocking issue in the tag filter implementation: the documented exact-tag filter is implemented with unescaped SQL LIKE wildcards, so user-supplied `%` / `_` stop being literal tag characters and broaden the result set.

Verification run locally:
- `uv run --no-project --with pytest --with flask --with pynacl --with requests --with werkzeug --with markupsafe pytest tests/test_discoverability.py::TestSearchEnhancements::test_search_agent_filter tests/test_discoverability.py::TestSearchEnhancements::test_search_tag_filter -q` -> 2 passed
- `uv run --no-project --with flask --with pynacl --with requests python -m py_compile bottube_server.py tests/test_discoverability.py` -> passed
- `uv run --no-project --with flake8 flake8 tests/test_discoverability.py --select E9,F821,F811,F841` -> passed
- `sqlite3 :memory:` LIKE probe showed wildcard tag patterns match ordinary JSON tags.

I also ran the full `tests/test_discoverability.py`; it currently has unrelated existing failures around search upload/trending/category endpoints, so I did not use that as the blocking signal for this diff. The blocking issue below is specific to the new tag filter code.

JacKane21:
Addressed the review finding at new head `b30bc04`: the `tag` filter now checks tags through `json_each(v.tags)` with exact lowercase equality instead of `LIKE`, so `%` and `_` in tag names are treated literally. Added regression coverage for `a%b` and `a_b` tags.

Validation run:
- `uv run --no-project --with pytest --with flask --with pynacl --with requests --with werkzeug --with markupsafe pytest tests/test_discoverability.py::TestSearchEnhancements::test_search_agent_filter tests/test_discoverability.py::TestSearchEnhancements::test_search_tag_filter tests/test_discoverability.py::TestSearchEnhancements::test_search_tag_filter_treats_like_wildcards_as_literals -q` -> 3 passed
- `git diff --check`

JacKane21:

PR #1211, opened 2026-05-22
Validate beef arc JSON inputs

### LINKED ISSUE #1210: [Bug] Beef drama arc endpoints crash on malformed JSON fields
## Summary

The BoTTube beef/drama arc API hardens `/api/beef/relationships`, but the drama-arc routes still assume parsed JSON is a dictionary and that arc fields have JSON-safe scalar types. Malformed-but-valid JSON can therefore reach `.get()`, dict membership, or SQLite binding paths instead of returning deterministic 400 validation errors.

## Affected routes

- `POST /api/beef/arcs`
- `POST /api/beef/arcs/<arc_id>/resolve`

## Reproduction

1. Send a non-object JSON body to `POST /api/beef/arcs`, for example `["not", "an", "object"]`.
2. Send a non-string `arc_template`, for example `{ "relationship_id": 1, "arc_template": ["hot_take_beef"] }`.
3. Send a non-object JSON body or non-string `resolution_note` to `POST /api/beef/arcs/<arc_id>/resolve`.

## Expected

The endpoints should return 400-class validation responses and avoid database writes for malformed client input.

## Actual

The handlers call `data.get(...)` or use unvalidated values directly after `request.get_json(silent=True) or {}`, so non-object JSON and non-string fields can escape the validation path.

## Notes

This is distinct from the existing `/api/beef/relationships` validation, which already rejects non-object JSON and malformed numeric fields.

Browser/device: API-level issue reproduced through Flask request handling; not browser-specific.

kebanks2:
## Summary

- fixes `/api/beef/arcs` so non-object JSON bodies return 400 instead of crashing on `.get()`
- validates `relationship_id` through the existing positive-integer parser before DB lookup
- rejects non-string or blank `arc_template` values with deterministic 400 responses
- fixes `/api/beef/arcs/<arc_id>/resolve` so non-object JSON and non-string `resolution_note` values return 400 instead of reaching DB binding
- adds focused regression coverage for malformed arc start/resolve inputs

Fixes #1210.

## Validation

- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m pytest -p no:cacheprovider tests/test_beef_system.py -q`
- `PYTHONDONTWRITEBYTECODE=1 /tmp/bottube-oembed-venv/bin/python -m py_compile agent_relationships.py tests/test_beef_system.py`
- `git diff --check`

### DIFF
--- a/agent_relationships.py
+++ b/agent_relationships.py
@@ -412,12 +412,24 @@ def start_arc():
       relationship_id  int
       arc_template     str  (must be one of DRAMA_ARC_TEMPLATES keys)
     """
-    data = request.get_json(silent=True) or {}
-    rel_id = data.get("relationship_id")
+    data = request.get_json(silent=True)
+    if data is None:
+        data = {}
+    if not isinstance(data, dict):
+        return jsonify({"error": "JSON object required"}), 400
+
+    rel_id, error = _parse_positive_int(
+        data.get("relationship_id"), "relationship_id"
+    )
+    if error:
+        return jsonify({"error": error}), 400
     template = data.get("arc_template")
 
-    if not rel_id or not template:
-        return jsonify({"error": "relationship_id and arc_template required"}), 400
+    if not isinstance(template, str):
+        return jsonify({"error": "arc_template must be a string"}), 400
+    template = template.strip()
+    if not template:
+        return jsonify({"error": "arc_template required"}), 400
     if template not in DRAMA_ARC_TEMPLATES:
         return jsonify({"error": f"unknown template '{template}'"}), 400
 
@@ -450,8 +462,16 @@ def start_arc():
 @beef_bp.route("/arcs/<int:arc_id>/resolve", methods=["POST"])
 def resolve_arc(arc_id: int):
     """Mark an arc as resolved."""
-    data = request.get_json(silent=True) or {}
+    data = request.get_json(silent=True)
+    if data is None:
+        data = {}
+    if not isinstance(data, dict):
+        return jsonify({"error": "JSON object required"}), 400
     note = data.get("resolution_note", "")
+    if note is None:
+        note = ""
+    if not isinstance(note, str):
+        return jsonify({"error": "resolution_note must be a string"}), 400
     db = get_db()
     db.execute(
         "UPDATE drama_arcs SET status='resolved', resolved_at=?, resolution_note=? WHERE id=?",
--- a/tests/test_beef_system.py
+++ b/tests/test_beef_system.py
@@ -286,6 +286,35 @@ def test_start_drama_arc(client):
     assert arc["status"] == "active"
 
 
+def test_start_drama_arc_rejects_non_object_json(client):
+    resp = client.post("/api/beef/arcs", json=["not", "an", "object"])
+    assert resp.status_code == 400
+    assert resp.get_json()["error"] == "JSON object required"
+
+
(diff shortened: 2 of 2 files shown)


JacKane21 (approved):
Reviewed PR #1211 at head `bc66e2ff5e85522248cd8edd7fc9e83e500431cf`.

What I checked:
- `POST /api/beef/arcs` now handles `request.get_json(silent=True)` without assuming the parsed body is a dict, so array JSON returns a deterministic 400 instead of reaching `.get()`.
- `relationship_id` now goes through `_parse_positive_int`, which keeps missing, list-valued, zero, negative, and non-integer values on the validation path before any DB lookup.
- `arc_template` is type-checked and stripped before membership checks against `DRAMA_ARC_TEMPLATES`, so non-string and blank values get specific 400 responses.
- `POST /api/beef/arcs/<arc_id>/resolve` now rejects non-object JSON and non-string `resolution_note` before SQLite binding. The `None -> ""` normalization preserves the existing optional-note behavior.
- The added regressions cover the reported crash shapes and retain happy-path start/resolve behavior.

Validation I ran:
- `uv run --no-project --with pytest --with flask python -m pytest -p no:cacheprovider tests/test_beef_system.py -q` -> 36 passed
- `uv run --no-project --with flask python -m py_compile agent_relationships.py tests/test_beef_system.py` -> passed

No blocking issues found. One small follow-up outside this PR's malformed-input scope: `resolve_arc()` still performs the UPDATE before checking whether the arc exists, so a future cleanup could fetch first and return 404 before executing a no-op update. That does not affect the current fix because SQLite no-ops this path and the endpoint still returns 404 for missing IDs.

crowniteto (commented):
## Code Review: Validate beef arc JSON inputs (#1211)

### 1. Good: Replacing `or {}` with explicit None handling
Changing `request.get_json(silent=True) or {}` to explicit `if data is None: data = {}` followed by `if not isinstance(data, dict)` is more correct — the old `or {}` pattern would also silently treat `False`, `0`, and `[]` as empty, masking potential bugs. The new pattern only treats actual None (no JSON body) as empty.

### 2. Good: `_parse_positive_int` for `relationship_id`
Extracting integer parsing with validation is the right approach. The parametrized tests covering `relationship_id: ["1"]` (list), `arc_template: ["hot_take_beef"]` (list), and `arc_template: " "` (whitespace-only) are thorough edge cases.

### 3. Concern: `resolve_arc` allows any string as `resolution_note`
The validation checks that `resolution_note` is a string, but there's no length limit. An extremely long note could bloat the database. Consider adding a max length:
```python
if len(note) > 5000:
    return jsonify({"error": "resolution_note too long (max 5000 chars)"}), 400
```

### 4. Observation: Inconsistent error message style
In `start_arc`, the error for a non-string template is `"arc_template must be a string"`, but for an unknown template the error is `f"unknown template '{template}'"` — which includes user input in the error message. This could be an information disclosure vector if the template value is logged or displayed. Consider using a generic message for production.

---

jhdev2026:
## PR Review — #1211 Validate Beef Arc JSON Inputs

**Reviewed:** Beef arc endpoints — JSON input validation.

### What this PR does
Adds JSON input validation for beef arc endpoints.

### Technical observations

**Same validation pattern:**
Consistent with the established pattern across this codebase.

**Conclusion:** Standard input validation improvement.

PR #1218, opened 2026-05-22
Label header search input for accessibility

### LINKED ISSUE #1217: Accessibility: Header search input lacks accessible label
## Accessibility issue

The shared header search form in `bottube_templates/base.html` labels the `<form role="search">`, but the text input itself has no associated `<label>`, `aria-label`, or `aria-labelledby`.

Current source on `main`:

```html
<form class="search-bar" action="{{ P }}/search" method="get" role="search" aria-label="{{ _('nav.search') }}">
    <input type="text" name="q" placeholder="{{ _('nav.search_placeholder') }}" value="{{ request.args.get('q', '') }}">
    <button type="submit" aria-label="{{ _('nav.search') }}">{{ _('nav.search') }}</button>
</form>
```

## Why this matters

A named search region does not automatically give the child text field an accessible name. Screen reader users navigating by form controls can encounter an unlabeled text input, and placeholders are not a reliable accessible-name substitute.

This affects the global header search shown across pages, so it is separate from the page-specific `/agents` search input reported in #1216.

## Suggested fix

Add an associated visually-hidden label and/or an `aria-label` to the input.

WCAG references: 1.3.1 Info and Relationships, 3.3.2 Labels or Instructions, 4.1.2 Name, Role, Value.

Bounty context: candidate accessibility fix/report for rustchain-bounties#1618 and rustchain-bounties#2139.

JacKane21:
## Summary

- Adds an associated `sr-only` label to the shared header search input.
- Adds an `id` and `aria-label` fallback so the input itself has an accessible name, not only the surrounding search form.
- Adds regression coverage for the header search input accessible name.

Fixes #1217.

Bounty context: candidate accessibility fix/report for Scottcjn/rustchain-bounties#1618 and Scottcjn/rustchain-bounties#2139.

/claim Scottcjn/rustchain-bounties#2139

## Validation

- `uv run --no-project --with pytest python -m pytest -p no:cacheprovider tests/test_accessibility.py -q` -> 18 passed
- Parsed `bottube_templates/base.html` with BeautifulSoup and confirmed `#site-search-input` has both an associated label and `aria-label`.

## Payout boundary

This is a public bounty claim for maintainer assessment only; no RTC award, payout approval, wallet transfer, wallet credit, bank transfer, exchange movement, or payment receipt is asserted here.

### DIFF
--- a/bottube_templates/base.html
+++ b/bottube_templates/base.html
@@ -814,7 +814,8 @@
         </div>
 
         <form class="search-bar" action="{{ P }}/search" method="get" role="search" aria-label="{{ _('nav.search') }}">
-            <input type="text" name="q" placeholder="{{ _('nav.search_placeholder') }}" value="{{ request.args.get('q', '') }}">
+            <label for="site-search-input" class="sr-only">{{ _('nav.search') }}</label>
+            <input id="site-search-input" type="text" name="q" placeholder="{{ _('nav.search_placeholder') }}" value="{{ request.args.get('q', '') }}" aria-label="{{ _('nav.search') }}">
             <button type="submit" aria-label="{{ _('nav.search') }}">{{ _('nav.search') }}</button>
         </form>
 
--- a/tests/test_accessibility.py
+++ b/tests/test_accessibility.py
@@ -88,6 +88,18 @@ class TestAccessibilityAttributes(unittest.TestCase):
                       "Search form should have role='search'")
         self.assertIn('aria-label', match.group(0),
                       "Search form missing aria-label")
+
+    def test_header_search_input_has_accessible_name(self):
+        """Test that the header search input is labelled for screen readers."""
+        content = self.read_file(self.TEMPLATE_DIR / 'base.html')
+        label = re.search(r'<label[^>]*for="site-search-input"[^>]*>', content)
+        self.assertIsNotNone(label, "Header search input missing associated label")
+        self.assertIn('class="sr-only"', label.group(0),
+                      "Header search label should use sr-only utility")
+        field = re.search(r'<input[^>]*id="site-search-input"[^>]*>', content)
+        self.assertIsNotNone(field, "Header search input not found")
+        self.assertIn('aria-label', field.group(0),
+                      "Header search input missing aria-label fallback")
     
     def test_skip_link_present(self):
         """Test that skip link for keyboard navigation is present."""


JacKane21:
CI note: the code checks are green (`lint`, `test`, and `security`). The only failing job is `auto-label`, which fails inside `actions/labeler` with `Resource not accessible by integration` because the fork PR token has read-only pull-request permissions. This appears to need maintainer/repo-side permission handling rather than a branch code change.

crowniteto (commented):
## Review: Label header search input for accessibility

Good accessibility improvement. Here are my observations:

**Positive:**
- Dual approach of both an associated `<label>` and `aria-label` on the input is robust — the label covers label-input association (for click-to-focus and form reader announcements), while `aria-label` acts as a fallback for screen readers that might not traverse the label association correctly.
- Using `sr-only` (visually hidden but screen-reader accessible) on the label is the correct pattern — it preserves visual design while adding semantic accessibility.
- The test coverage in `test_accessibility.py` is thorough: it checks both the `for` attribute on the label and the `id` + `aria-label` on the input.

**Suggestions:**
1. **Redundant `aria-label` on the submit button (line 819):** The submit button already has `aria-label="{{ _(nav.search) }}"` which duplicates the search label text. Since the button text content is already `{{ _(nav.search) }}`, the `aria-label` is redundant — screen readers will announce the button text. This is not introduced by this PR, but since the PR is touching this form, it might be worth a follow-up cleanup.

2. **`sr-only` class assumption:** The test asserts `class="sr-only"` is present, but many frameworks use `visually-hidden`, `screen-reader-only`, or a custom utility class. If BoTTube ever switches CSS frameworks, this test will break. Consider checking for the semantic structure (label exists, has `for`, input has `id`) rather than the specific class name. Minor point.

3. **i18n key reuse:** The label uses `{{ _(nav.search) }}` which is the same key as the form `aria-label` and button text. This is correct for consistency, but if the UX team ever wants the label to say "Search this site" while the button stays "Search", this key coupling would need refactoring. Not a concern for now.

Well-structured change that follows accessibility best practices. The regression test is a nice addition.

JacKane21:

jhdev2026:
## PR Review — #1218 Label Header Search Input for Accessibility

**Reviewed:** Header search input accessibility labeling.

### What this PR does
Adds accessibility labels to the header search input.

### Technical observations

**Form accessibility:**
Header search is often the primary navigation. Proper labeling helps screen reader users locate and use the search function.

**Conclusion:** Good accessibility improvement.

PR #1226, opened 2026-05-23
Reject non-object generation job JSON bodies

### LINKED ISSUE #1225: Generation job auth crashes on JSON string bodies
## Summary

`POST /api/generation/jobs` can return a 500 before authentication when the request body is valid JSON but not an object. The generation route decorator accepts an API key either from `X-API-Key` or from `agent_api_key` in the JSON body, but it calls `.get()` on the parsed JSON before checking that the parsed value is a dict.

A JSON string body is truthy, so `request.get_json(silent=True) or {}` leaves `data` as a Python `str`, then `data.get("agent_api_key", "")` raises `AttributeError`.

## Steps to reproduce

Environment used for local reproduction:
- macOS 26.3.1 / Darwin 25.3.0 arm64
- Python via `uv run --no-project --with flask`
- Flask test client registered with `generation.routes.generation_bp`

Minimal repro:

```python
import sys
sys.path.insert(0, "bottube")
from flask import Flask
from generation.routes import generation_bp

app = Flask(__name__)
app.register_blueprint(generation_bp)
client = app.test_client()

resp = client.post("/api/generation/jobs", json="not-object")
print(resp.status_code, resp.get_data(as_text=True)[:80])
```

Observed traceback points at:

```python
# generation/routes.py
api_key = data.get("agent_api_key", "")
```

## Expected

The endpoint should fail closed with a deterministic 401/400 JSON response, such as missing API key or JSON object required. Non-object JSON should not escape into a server error.

## Actual

The request returns `500 Internal Server Error` for a valid JSON string body before auth is completed.

## Suggested fix

Only treat `None` as an empty body, then reject or ignore non-object JSON before field access. For example:

```python
if data is None:
    data = {}
    return jsonify({"error": "JSON object required"}), 400
```

The same pattern should probably be applied both in `_require_api_key` and in `create_generation_job()` so authenticated malformed bodies are handled consistently too.

JacKane21:
## Summary
- add shared JSON-object parsing for both generation job APIs
- reject non-object JSON before reading `agent_api_key` or generation fields
- validate legacy `/api/generate-video` field types for `prompt`, `duration`, `category`, and `title` so malformed authenticated bodies return 400s instead of 500s
- cover unauthenticated/authenticated non-object bodies plus authenticated malformed field bodies

Fixes #1225.

## Validation
- `uv run --no-project --with flask --with pytest python -B -m pytest -q tests/test_generation_routes.py --tb=short`
- `python3 -B -m py_compile generation/routes.py video_gen_blueprint.py tests/test_generation_routes.py`
- `git diff --check`

/claim #1102

### DIFF
--- a/generation/routes.py
+++ b/generation/routes.py
@@ -42,13 +42,31 @@ def init_routes(router: GenerationRouter, publish_fn=None):
     _publish_fn = publish_fn
 
 
+def _json_object_body():
+    data = request.get_json(silent=True)
+    if data is None:
+        return {}, None
+    if not isinstance(data, dict):
+        return None, (jsonify({"error": "JSON object required"}), 400)
+    return data, None
+
+
+def _string_field(data, field_name, default=""):
+    value = data.get(field_name, default)
+    if not isinstance(value, str):
+        return None, (jsonify({"error": f"{field_name} must be a string"}), 400)
+    return value, None
+
+
 def _require_api_key(f):
     """Accept X-API-Key header or agent_api_key in JSON body."""
     @wraps(f)
     def decorated(*args, **kwargs):
         api_key = request.headers.get("X-API-Key", "")
         if not api_key:
-            data = request.get_json(silent=True) or {}
+            data, error_response = _json_object_body()
+            if error_response:
+                return error_response
             api_key = data.get("agent_api_key", "")
         if not api_key:
             return jsonify({"error": "Missing API key"}), 401
@@ -102,8 +120,13 @@ def _record_rate(api_key: str):
 @_require_api_key
 def create_generation_job():
     """Create a new video generation job."""
-    data = request.get_json(silent=True) or {}
-    prompt = (data.get("prompt") or "").strip()
+    data, error_response = _json_object_body()
+    if error_response:
+        return error_response
+    prompt_value, error_response = _string_field(data, "prompt", "")
+    if error_response:
+        return error_response
+    prompt = prompt_value.strip()
     if not prompt:
         return jsonify({"error": "prompt is required"}), 400
     if len(prompt) > 500:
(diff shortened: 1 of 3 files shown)


kebanks2 (changes requested):
Security/correctness review:

- `generation/routes.py:113-116`: the new object-body guard still leaves `prompt = (data.get("prompt") or "").strip()` in the modern `/api/generation/jobs` path. An authenticated JSON object such as `{ "prompt": ["make a video"] }` reaches this line and raises `AttributeError` instead of returning a structured 400.
- The legacy `video_gen_blueprint.py` path added explicit `prompt` string validation in this same PR, so the newer generation route should get the same check before calling `.strip()`.
- Test gap: add a header-authenticated `/api/generation/jobs` regression for a non-string `prompt`; the existing tests cover non-object bodies but not this field-level crash.

The route-level object validation is useful, but this PR still leaves a malformed authenticated generation request able to escape the intended validation path.

JacKane21:
Updated for the review feedback in 76a86ea: the modern `/api/generation/jobs` path now validates `prompt` as a string before calling `.strip()`, matching the legacy route behavior. Added a header-authenticated regression for a non-string prompt.\n\nValidation run:\n- `python3 -B -m py_compile generation/routes.py tests/test_generation_routes.py`\n- `uv run --no-project --with flask --with pytest python -B -m pytest -q tests/test_generation_routes.py --tb=short` -> 9 passed

JacKane21:

jhdev2026:
## PR Review — #1226 Reject Non-Object Generation Job JSON Bodies

**Reviewed:** Generation job endpoint — JSON body validation.

### What this PR does
Rejects non-object JSON bodies for generation job requests before business logic runs.

### Technical observations

**Same validation pattern:**
Consistent with other PRs in this batch — validate structure first.

**Generation jobs:**
Job creation endpoints often trigger background processing. Malformed input should never reach the job queue.

**Conclusion:** Standard input validation fix.

PR #1247, opened 2026-05-29
Fix watch subscribe aria-label channel name

### LINKED ISSUE #1245: Accessibility: watch-page Subscribe link has empty channel name in aria-label
## Summary
On logged-out watch pages, the Subscribe link has an accessible name with the channel name missing. The visible UI shows the channel name, but the control announces only `Subscribe to ` because the rendered `aria-label` is empty after the preposition.

## Live example
URL: https://bottube.ai/watch/mMdIBpnzoib

Observed HTML around the channel row on 2026-05-26:

```html
<div class="channel-name" id="channel-name">
  <a href="/agent/sssnake_general">Sssnake General</a>
</div>
...
<a href="/login" aria-label="Subscribe to " class="watch-sub-btn not-following">Subscribe</a>
```

## Why this matters
Screen reader users encounter a Subscribe control without the target channel in its accessible name. On a video page with multiple actions, that makes the control less clear than the visible UI and can fail the expectation that link/button purpose is programmatically determinable.

## Expected
The accessible name should include the displayed channel identity, for example:

```html
aria-label="Subscribe to Sssnake General"
```

## Likely source
The template appears to render the watch-page subscription control with `{{ video.channel_name }}`. On this live page that value renders empty, while the channel link itself has the usable name `Sssnake General`. A safer fallback would use the same display value as the visible channel link, such as display name or agent name.

## Claim boundary
Filed as an accessibility report for maintainer assessment. No RTC award, transfer, wallet credit, or payout is asserted here.

gaoaaa52-del:
## Summary
- derive one watch-page channel display name from `display_name`, `channel_name`, or `agent_name`
- use that value for the channel avatar, visible channel link, and Subscribe aria-labels
- add regression coverage for the logged-out Subscribe link so it cannot render `Subscribe to ` with an empty target

Fixes #1245.

## Validation
- `/tmp/bottube-test-venv312/bin/python -B -m pytest -q tests/test_watch_page_accessibility.py --tb=short`

RTC wallet / miner ID: `gaoaaa52-del`

### DIFF
--- a/bottube_templates/watch.html
+++ b/bottube_templates/watch.html
@@ -2516,11 +2516,12 @@
             </div>
             {% endif %}
 
+            {% set channel_display_name = video.display_name or video.channel_name or video.agent_name %}
             <div class="channel-row" role="group" aria-label="Channel information">
-                <img class="channel-avatar" src="{{ video.avatar_url or (P ~ "/avatar/" ~ video.agent_name ~ ".svg") }}" alt="{{ video.display_name or video.agent_name }} avatar">
+                <img class="channel-avatar" src="{{ video.avatar_url or (P ~ "/avatar/" ~ video.agent_name ~ ".svg") }}" alt="{{ channel_display_name }} avatar">
                 <div>
                     <div class="channel-name" id="channel-name">
-                        <a href="{{ P }}/agent/{{ video.agent_name }}">{{ video.display_name or video.agent_name }}</a>
+                        <a href="{{ P }}/agent/{{ video.agent_name }}">{{ channel_display_name }}</a>
                         {% if video.is_human %}
                         <span class="badge-human">Human</span>
                         {% else %}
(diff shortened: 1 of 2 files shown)


eliasx45 (approved):
Approved. I validated this at head b0ebb1a6866c3e3f54bae317cd8c43ee20ce9001 against current origin/main.

Evidence:
- `git diff --check origin/main...HEAD` clean
- `git merge-tree --write-tree origin/main HEAD` succeeded: `5b756f415f77ffc9b65676169411380d24cf9d3b`
- `python -B -m pytest -q tests/test_watch_page_accessibility.py --tb=short` -> 9 passed

The template now derives one non-empty channel display name and uses it consistently for the avatar alt text, visible channel link, and logged-out/logged-in Subscribe aria-labels. The new regression covers the logged-out link and verifies it no longer renders an empty `Subscribe to ` label.

PR #1262, opened 2026-05-29
Add cache validators to video list API

### LINKED ISSUE #1259: Feature request: Add ETag/Last-Modified headers to video list API for caching
## Feature Request

**Problem**: The `/api/videos` endpoint doesn't include HTTP caching headers. Clients that poll for new videos (like the BoTTube Python Client SDK) re-download the full list on every request.

**Proposed solution**: 
- Add `ETag` header based on the hash of the latest video timestamp
- Add `Last-Modified` header
- Respect `If-None-Match` and `If-Modified-Since` request headers

**Impact**: Would reduce bandwidth by ~90% for clients polling for updates. Currently the full list response is ~80KB with 250+ agents × their videos.

**Implementation hint**: A simple approach would be to store `max(updated_at)` across all videos and use that as the `Last-Modified` value.

**Tested with**: Python `requests` library, verified response lacks caching headers via `/health` endpoint observation.

gaoaaa52-del:
## Summary
- add weak ETag, Last-Modified, and Cache-Control headers to `GET /api/videos`
- return `304 Not Modified` for matching `If-None-Match` or fresh `If-Modified-Since` requests
- include visible video count/page/sort plus upload and engagement revisions in the cache validator so clients do not reuse stale list data after new uploads, views, or votes

Fixes #1259

## Validation
- `/tmp/bottube-test-venv312/bin/python -B -m pytest -q tests/test_video_list_cache_headers.py tests/test_upload_api.py::TestVideoListEndpoint --tb=short` -> 8 passed
- `/tmp/bottube-test-venv312/bin/python -B -m py_compile bottube_server.py tests/test_video_list_cache_headers.py` -> passed

## Safety
No credentials, payments, uploads, external services, or live account actions were used. This is a read-only API response/header change with local Flask client tests.

RTC wallet / miner ID: `gaoaaa52-del`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -26,6 +26,7 @@ import urllib.parse
 import urllib.request
 from email.mime.multipart import MIMEMultipart
 from email.mime.text import MIMEText
+from email.utils import formatdate, parsedate_to_datetime
 from functools import wraps
 from pathlib import Path
 from typing import Dict, List, Optional, Tuple
@@ -37,6 +38,7 @@ from flask import (
     flash,
     g,
     jsonify,
+    make_response,
     redirect,
     render_template,
     request,
@@ -6653,6 +6655,60 @@ def update_video(video_id):
 # Video listing / detail
 # ---------------------------------------------------------------------------
 
+def _video_list_etag(
+    *,
+    page: int,
+    per_page: int,
+    sort: str,
+    agent_name: str,
+    total: int,
+    latest_ts: float,
+    engagement_revision: int,
+) -> str:
+    cache_key = json.dumps(
+        {
+            "agent": agent_name,
+            "latest_ts": latest_ts,
+            "page": page,
+            "per_page": per_page,
+            "engagement_revision": engagement_revision,
+            "sort": sort,
+            "total": total,
+        },
+        sort_keys=True,
+        separators=(",", ":"),
+    )
+    digest = hashlib.sha256(cache_key.encode("utf-8")).hexdigest()[:24]
+    return f'W/"videos-{digest}"'
+
+
+def _client_has_video_list_etag(etag: str) -> bool:
+    raw_header = request.headers.get("If-None-Match", "")
+    if not raw_header:
+        return False
+    candidates = {part.strip() for part in raw_header.split(",")}
+    return "*" in candidates or etag in candidates
+
+
+def _client_has_fresh_video_list_date(latest_ts: float) -> bool:
+    raw_header = request.headers.get("If-Modified-Since", "")
+    if not raw_header:
+        return False
+    try:
+        modified_since = parsedate_to_datetime(raw_header)
+    except (TypeError, ValueError):
+        return False
+    if modified_since is None:
+        return False
+    return int(latest_ts or 0) <= int(modified_since.timestamp())
+
+
+def _add_video_list_cache_headers(response: Response, *, etag: str, latest_ts: float) -> Response:
+    response.headers["ETag"] = etag
+    response.headers["Last-Modified"] = formatdate(int(latest_ts or 0), usegmt=True)
+    response.headers["Cache-Control"] = "public, max-age=30"
+    return response
+
 @app.route("/api/videos")
 def list_videos():
     """List videos with pagination and sorting."""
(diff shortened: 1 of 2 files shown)


eliasx45 (changes requested):
Reviewed current head `aefb532491d73a3e7603b9e1e3878ebe26c78ccb`.

Verdict: request changes.

The tests pass, but I found a blocker in the cache validator stats query. `latest_engagement_ts` is computed with scalar `MAX(x, y)` inside an aggregate `SELECT`:

```sql
MAX(
  COALESCE((SELECT MAX(vw.created_at) FROM views vw WHERE vw.video_id = v.video_id), 0),
  COALESCE((SELECT MAX(vt.created_at) FROM votes vt WHERE vt.video_id = v.video_id), 0)
) AS latest_engagement_ts
```

In SQLite, `MAX(x, y)` with two arguments is scalar, not aggregate. In this aggregate query without `GROUP BY`, SQLite can evaluate that non-aggregate expression for an arbitrary visible row instead of taking the maximum engagement timestamp across all visible videos. That means the ETag/Last-Modified freshness can miss newer views/votes on another row.

I reproduced the shape with two visible videos: newer video created at `200`, older video with a vote at `1000`. The stats query returned:

```text
{'total': 2, 'latest_ts': 200.0, 'latest_engagement_ts': 0, 'engagement_revision': 1}
```

Expected `latest_engagement_ts` should include `1000`. This matters for `If-Modified-Since`, and it can also leave ETag stale for vote changes where the summed engagement count does not change, such as a like changing to a dislike on an older row while the arbitrary row chosen for the scalar expression has no engagement.

Suggested fix: aggregate the per-row engagement timestamp across the whole filtered video set, for example by wrapping the scalar max in an aggregate max over rows, or by computing latest view/vote timestamps in separate aggregate subqueries joined/filtered to the same visible video set. Please also add a regression with at least two visible videos where the newest engagement belongs to the non-newest row, and assert the validator changes / `If-Modified-Since` does not incorrectly return 304.

Validation I ran:

```text
# clean

git merge-tree --write-tree origin/main HEAD
# 742aa5779d990da224f614f2125f652fc69fdf55

python -m py_compile bottube_server.py tests\test_video_list_cache_headers.py
# passed

python -m pytest -q tests\test_video_list_cache_headers.py tests\test_upload_api.py::TestVideoListEndpoint --tb=short
# 8 passed
```

GitHub readback: project `test`, `lint`, and `security` jobs are green; only `auto-label` is red. The blocker above is in the PR logic, not the hosted checks.

gaoaaa52-del:
Thanks for the detailed catch. I updated the branch to aggregate the per-row engagement timestamp across the full visible video set instead of relying on scalar `MAX(x, y)` as the aggregate result.

I also added a regression with two visible videos where the newest engagement belongs to the older video, then verifies `If-Modified-Since` does not incorrectly return `304` and `Last-Modified` advances to that engagement timestamp.

Validation after the fix:
- `/private/tmp/bottube-test-venv312/bin/python -B -m py_compile bottube_server.py tests/test_video_list_cache_headers.py`
- `/private/tmp/bottube-test-venv312/bin/python -B -m pytest -q tests/test_video_list_cache_headers.py tests/test_upload_api.py::TestVideoListEndpoint --tb=short` -> `9 passed`
- `git diff --check`

eliasx45 (approved):
Re-reviewed current head `a17e07ebfa2846e293835a4852d1df4daf0eadb6` after the cache-validator aggregate fix.

Verdict: approve.

The blocker from my prior review is fixed. `latest_engagement_ts` now wraps the per-row scalar `MAX(view_ts, vote_ts)` in an aggregate `MAX(...)` across all visible videos, so engagement on a non-newest row can advance `Last-Modified` and the ETag. The new regression covers the exact shape I asked for: two visible videos, newer engagement on the older video, and `If-Modified-Since` correctly returning 200 with an advanced `Last-Modified` instead of a stale 304.

Validation on this checkout:

```text
# clean

git merge-tree --write-tree origin/main HEAD
# d35c87eb0ac153961ae31af4de7f45ded56d627c

python -m py_compile bottube_server.py tests\test_video_list_cache_headers.py
# passed

python -m pytest -q tests\test_video_list_cache_headers.py tests\test_upload_api.py::TestVideoListEndpoint --tb=short
# 9 passed
```

I also reran a standalone SQLite probe matching the previously failing shape: visible newer video at `200`, older video with a vote at `1000`. The updated expression returned `latest_engagement_ts: 1000.0`, which confirms the aggregate no longer depends on SQLite's arbitrary row choice for the scalar expression.

GitHub readback: project `test`, `lint`, and `security` jobs are green; only the auxiliary `auto-label` job is red. I do not see a remaining blocker.

PR #1306, opened 2026-06-05
fix: remove duplicate watch keyboard shortcut handler

### LINKED ISSUE #1305: Watch page registers duplicate keyboard shortcut handlers
﻿## Summary

`bottube_templates/watch.html` currently registers two global `keydown` handlers for the same video player shortcuts. The first, accessible implementation starts around the `shortcut-indicator` / `shortcut-help-modal` code and handles Space/K, J/L, arrows, M, F, C, Escape, and `?`. A second `// --- Video Player Keyboard Shortcuts (Bounty #2140) ---` block later in the file registers another listener for the same keys.

## Reproduction

Inspect current `main`:

```bash
curl -s https://raw.githubusercontent.com/Scottcjn/bottube/main/bottube_templates/watch.html \
  | grep -n "Video Player Keyboard Shortcuts\|document.addEventListener('keydown'"
```

Observed relevant matches:

```text
3185:document.addEventListener('keydown', function(event) {
4127:// --- Video Player Keyboard Shortcuts (Bounty #2140) ---
4146:    document.addEventListener('keydown', function(e) {
```

## Actual behavior

Both listeners run for the same shortcut event when focus is not in an input/textarea/contenteditable field.

Examples:
- Pressing `J` can apply the first handler's `seekVideo(video, -10)` and then the second handler's `video.currentTime -= 10`, producing a 20 second rewind.
- Pressing ArrowLeft can seek 5 seconds twice, producing a 10 second rewind.
- Pressing Space/K can trigger two play/pause paths and two different visual indicators.

## Expected behavior

Only one shortcut implementation should handle the event. The accessible implementation with the existing `shortcut-indicator`, `player-state` live region, help modal, fullscreen handling, captions toggle, and bypass checks should be retained; the later duplicate block should be removed.

## Environment

- URL/template inspected: `bottube_templates/watch.html` on `Scottcjn/bottube@main`
- OS: Windows 11 / PowerShell
- Browser target: any desktop browser where the watch page JS runs

## Impact

The keyboard shortcut bounty UX becomes inconsistent: documented seek amounts are doubled for overlapping seek keys, and the page has two competing overlay systems for the same shortcuts.

tracepatch-lab:
## Summary
- removes the second watch-page keyboard shortcut listener that was added after the accessible handler
- keeps the existing accessible shortcut path with help text, indicator, repeat-key guard, and form-field checks
- prevents J/L and arrow-key shortcuts from firing twice on a single key press

Fixes Scottcjn/bottube#1305

### DIFF
--- a/bottube_templates/watch.html
+++ b/bottube_templates/watch.html
@@ -1,4 +1,4 @@
-{% extends "base.html" %}
+﻿{% extends "base.html" %}
 
 {% block title %}{{ video.title }}{% endblock %}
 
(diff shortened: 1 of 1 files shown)


Scottcjn:
Thanks — and you were first to spot the duplicate handler! Closing this specific PR only because it injects a UTF-8 BOM at the top of the template (`+﻿{% extends%}`), which can break Jinja rendering. The clean removal landed via #1308. For next time, make sure your editor saves without a BOM. Appreciate the catch. 🦞

Scottcjn:
You were **first** to catch the duplicate handler — sent you a **2 RTC** thanks-for-the-catch gift (to your `tracepatch-lab` github-keyed wallet; provide an `RTC…` address anytime to move it). The BOM was the only reason this exact PR didn't land. Keep contributing! 🎁🦞

tracepatch-lab:
Thanks for the review and for landing the clean version in #1308. Confirmed on the BOM note; I will make sure future template edits are saved without BOM.

RTC address for moving the github-keyed 2 RTC gift:
`RTCa14a8b8553834f4593db826222424420bf6f8417`

Appreciate the thanks-for-the-catch gift.

PR #1318, opened 2026-06-06
fix: announce bridge operation results to screen readers

2003elizhang:
## Description

Adds polite, atomic status live regions to the public-info, deposit,
withdrawal, and history result panels on both the Solana and Base bridge
pages.

These panels are updated asynchronously with loading, validation, success, and
error messages. Exposing them as status regions lets screen readers announce
the results without moving keyboard focus.

Fixes #1317

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Documentation update
- [ ] Refactor

## Testing

- `python -m unittest tests.test_accessibility -v`
- Result: 19 tests passed
- `git diff --check`
- Result: passed

## Checklist

- [x] My code follows the style of this project
- [x] I have performed a self-review of my changes
- [x] I have added tests that prove my fix is effective
- [x] New and existing targeted tests pass

RTC wallet name: `2003elizhang`

PR #73, opened 2026-02-08
[BOUNTY] #63 - Collab Bot Duo: Pi & Chaos

mccoychang:
## Description\nAdding a dynamic bot duo for Bounty #63.\n\n### Bots\n1. **Pi (The Architect)**: Existing rational bot.\n2. **Chaos (The Intern)**: New chaotic bot ().\n\n### Interaction\n- Included  demonstrating their interaction script.\n\n## Bounty\nCloses #63

PR #1326, opened 2026-06-08
fix(a11y): announce bridge result updates to screen readers

### LINKED ISSUE #1317: Accessibility: bridge operation results are not announced to screen readers
### Affected area

Upload / transcode

### What happened?

The public-info, deposit, withdrawal, and history panels on the bridge pages are updated dynamically after users activate their corresponding buttons. The result containers do not have role="status", aria-live, or another live-region mechanism.

Examples from the current templates:

```html
<div id="depositPanel" class="panel">Awaiting transaction hash...</div>
<div id="withdrawPanel" class="panel">Queue is currently active.</div>
<div id="historyPanel" class="panel">No requests yet.</div>
```

JavaScript later replaces their text with loading states, validation failures, API errors, and successful operation results. A screen-reader user who activates one of these controls may not know whether the operation started, failed, or completed.

Related bounty: Scottcjn/rustchain-bounties#1618

### Expected behavior

Status, error, and success updates should be announced without moving keyboard focus, for example by marking each result panel as role="status" with aria-live="polite" and aria-atomic="true".

### Steps to reproduce

1. Open https://bottube.ai/bridge/wrtc or https://bottube.ai/bridge/base with a screen reader.
2. Activate Refresh Info, Verify Transaction, Queue Withdrawal, or Load History.
3. Observe that the related panel changes visually.
4. The changed status is not announced because the panel is not exposed as a live region.

### Related endpoint or page

https://bottube.ai/bridge/wrtc and https://bottube.ai/bridge/base

### Environment

None

### Relevant logs, response bodies, or screenshots

```shell
Checked on June 6, 2026 against live production HTML and current main branch on Windows.

Suggested fix:


<div role="status" aria-live="polite" aria-atomic="true">...</div>
```

dallasnetprofu02-design:
## Summary
- mark the public info, deposit, withdrawal, and history bridge panels as polite live regions on both bridge pages
- keep updates in place so screen readers hear status changes without moving focus
- add an accessibility regression test that locks the live-region attributes in both bridge templates

## Testing
- pytest tests/test_accessibility.py -q
- pytest tests/test_base_wrtc_bridge_validation.py tests/test_wrtc_bridge_validation.py -q
- git diff --check

Closes #1317

### DIFF
--- a/bottube_templates/bridge_base.html
+++ b/bottube_templates/bridge_base.html
@@ -170,7 +170,7 @@
       <h2>Public Info</h2>
       <p>Live limits, fees, and aggregate bridge stats on Base.</p>
       <button class="btn btn-dark" id="loadInfoBtn" type="button" aria-label="Refresh public bridge info">Refresh Info</button>
-      <div id="infoPanel" class="panel" style="margin-top:10px;">Loading...</div>
+      <div id="infoPanel" class="panel" role="status" aria-live="polite" aria-atomic="true" style="margin-top:10px;">Loading...</div>
     </div>
 
     <div class="card">
@@ -181,7 +181,7 @@
       <label for="txHashInput">Transaction Hash (Base)</label>
       <input id="txHashInput" type="text" placeholder="0x... (66 character tx hash)">
       <button class="btn btn-primary" id="depositBtn" type="button" aria-label="Verify Base transaction and credit account">Verify Transaction</button>
-      <div id="depositPanel" class="panel" style="margin-top:10px;">Awaiting transaction hash...</div>
+      <div id="depositPanel" class="panel" role="status" aria-live="polite" aria-atomic="true" style="margin-top:10px;">Awaiting transaction hash...</div>
       <div class="warn">Ensure your bound ETH wallet (in profile settings) matches the sender address.</div>
     </div>
 
@@ -193,7 +193,7 @@
       <label for="withdrawAmountInput">Amount to Withdraw (wRTC)</label>
       <input id="withdrawAmountInput" type="number" min="10" step="0.000001" placeholder="Minimum: 10 wRTC">
       <button class="btn btn-primary" id="withdrawBtn" type="button" aria-label="Queue withdrawal to Base">Queue Withdrawal</button>
-      <div id="withdrawPanel" class="panel" style="margin-top:10px;">Queue is currently active.</div>
+      <div id="withdrawPanel" class="panel" role="status" aria-live="polite" aria-atomic="true" style="margin-top:10px;">Queue is currently active.</div>
       {% if swap_url %}
       <a class="link-btn" href="{{ swap_url }}" target="_blank" rel="noopener">Swap wRTC on Uniswap</a>
       {% endif %}
@@ -206,7 +206,7 @@
         <button class="btn btn-dark" id="historyBtn" type="button" aria-label="Load transaction history">Load History</button>
         <input id="historyLimitInput" type="number" min="1" max="200" value="50" style="max-width:100px;">
       </div>
-      <div id="historyPanel" class="panel" style="margin-top:10px;">No requests yet.</div>
+      <div id="historyPanel" class="panel" role="status" aria-live="polite" aria-atomic="true" style="margin-top:10px;">No requests yet.</div>
     </div>
   </div>
 </section>
--- a/bottube_templates/bridge_wrtc.html
+++ b/bottube_templates/bridge_wrtc.html
@@ -139,7 +139,7 @@
       <h2>Public Info</h2>
       <p>Live limits, fees, and aggregate bridge stats.</p>
       <button class="btn btn-dark" id="loadInfoBtn" type="button" aria-label="Refresh public bridge info">Refresh Info</button>
-      <div id="infoPanel" class="panel" style="margin-top:10px;">Loading...</div>
+      <div id="infoPanel" class="panel" role="status" aria-live="polite" aria-atomic="true" style="margin-top:10px;">Loading...</div>
     </div>
 
     <div class="card">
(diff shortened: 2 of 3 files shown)


jdjioe5-cpu (approved):
## Code Review — PR #1326

Reviewed head `1e21c7a891501928551a8291badd7174d0b9676b` on `origin/main` `8f3415b27b52769c18085d369cc33337e912bc00`.

**Verdict: APPROVE** — focused, correct, and the test asserts the right thing.

### Substantive findings

1. **ARIA pattern is correct for the stated scope.** The diff adds the standard live-region triple (`role="status"`, `aria-live="polite"`, `aria-atomic="true"`) to the four bridge result panels in both `bridge_base.html` (Base bridge) and `bridge_wrtc.html` (Solana/wRTC bridge). The `show()` helper at `bridge_base.html:226` / `bridge_wrtc.html:193` writes results into the panels via `panel.textContent = ...`, which is exactly the trigger a polite live region needs to announce updates. Before the change, a screen-reader user pressing "Refresh Info" or "Verify Transaction" got no audible feedback; after, they hear the result without their focus being stolen. Choosing `polite` over `assertive` is right here because the updates are user-action-driven, not server-pushed urgency.

2. **Test scope is appropriate.** `test_bridge_result_panels_are_live_regions` (lines 102-118 of `tests/test_accessibility.py`) is a regex static-template scan that asserts the three ARIA attributes are present on the four panel IDs in both templates. `subTest` is used so a failure points at the exact template × panel combination. ARIA markup is what we're enforcing; an end-to-end test of `show()` is not needed for this scope.

3. **Test regex is robust to attribute reordering.** The existing pattern `re.search(rf'<div[^>]*id="{panel_id}"[^>]*>', content)` is a substring scan over the open-tag block, not a strict parser. If a future PR reorders attributes in the template (e.g. swaps `class` and `role`), the test still matches. No change needed.

4. **Live-region cadence is implicit but acceptable.** The test asserts markup, not announcement cadence. If a future PR adds a debounce (e.g. to avoid announcing every keystroke from a polling endpoint), the test will still pass — which is correct for the current scope. The pre-existing behavior (announce every `textContent` write) is the right default.

5. **Out-of-scope but worth a follow-up issue:** the initial text of `depositPanel` ("Awaiting transaction hash...") and `withdrawPanel` ("Queue is currently active.") are status strings, not pending prompts. `aria-live="polite"` is the right call — not a bug. Just noting in case a future accessibility pass wants to convert these into `aria-describedby` hints for the adjacent input.

### Local validation

On a fresh `git clone` of `Scottcjn/bottube`, fetched PR #1326 head and ran the targeted test:

```
$ python3 -m pytest tests/test_accessibility.py -v
==================== 21 passed, 7 subtests passed in 0.08s =====================
```

The new test (`TestAccessibilityAttributes::test_bridge_result_panels_are_live_regions`) passes with 2 templates × 4 panel IDs = 8 sub-assertions. Other checks:

- `git merge-tree --write-tree origin/main HEAD` → clean
- `git diff --check` → clean
- `git diff --stat origin/main...HEAD` → `3 files changed, 26 insertions(+), 8 deletions(-)` (matches PR head)

(The 58 collection errors in the full `pytest tests/` run are pre-existing in `origin/main` and require DB/network env; not introduced by this PR.)

### Pre-flight gate

- 0 prior PR reviews on the current head
- 0 prior PR review comments
- 0 prior PR issue comments
- I am the first substantive-reviewer candidate for Bounty #73 on this PR
- Bounty #73 eligible repo (`Scottcjn/bottube`) — confirmed

Disclosure: I reviewed this PR for the RustChain Code Review Bounty #73.

bottube_templates/bridge_base.html
```diff
@@ -206,7 +206,7 @@ <h2>History</h2>
         <button class="btn btn-dark" id="historyBtn" type="button" aria-label="Load transaction history">Load History</button>
         <input id="historyLimitInput" type="number" min="1" max="200" value="50" style="max-width:100px;">
       </div>
-      <div id="historyPanel" class="panel" style="margin-top:10px;">No requests yet.</div>
+      <div id="historyPanel" class="panel" role="status" aria-live="polite" aria-atomic="true" style="margin-top:10px;">No requests yet.</div>
```
None-Feeling, line 209: Non-blocking: making the whole history panel an atomic status region is appropriate for the current short placeholder/result messages, but if this panel later renders a long list/table of history entries, screen readers may announce too much on each refresh. If that happens, consider splitting a short status summary live region from the detailed history content.

None-Feeling (approved):
Reviewed head `1e21c7a` against current `origin/main` `b42b118`.

The live-region change is focused and covers the two bridge pages called out by #1317. The public info, deposit, withdrawal, and history panels now have `role="status"`, `aria-live="polite"`, and `aria-atomic="true"`, and the regression test checks both `bridge_wrtc.html` and `bridge_base.html`.

Validation performed:

```bash
python -m py_compile tests/test_accessibility.py
python -m unittest tests.test_accessibility -v
git merge-tree origin/main HEAD
```

Results: diff check passed, the test file compiled, `unittest` ran 21 tests successfully, and the merge-tree check was clean.

One maintainer note: this overlaps with #1318, which also targets #1317. This PR is based on current `origin/main` and is merge-clean in my local check, but maintainers should merge only one of the two duplicate live-region implementations to avoid redundant conflict work. I left one non-blocking line-level note on possible verbosity if the history panel later renders large result lists.

Scottcjn:
Thanks for this — your HTML change was byte-identical to #1318 (same `role=status`/`aria-live=polite`/`aria-atomic` on all four bridge panels), which landed first. Closing as a duplicate, not a rejection — the work was correct. Appreciate the contribution; grab another open a11y issue and it's yours.

PR #1454, opened 2026-06-14
fix(search): return 400 instead of 500 on overflowing ?page (SQLite OFFSET)

Vyacheslav-Tomashevskiy:
## Summary

`GET /api/search` returns **HTTP 500** when `?page` is large enough that the
computed SQL `OFFSET` overflows SQLite's signed 64-bit `INTEGER` range. This
fix rejects such pages with a clean `400` instead of crashing, and adds a
regression test.

## Reproduction (production, checked 2026-06-14 UTC)

```bash
curl -i 'https://bottube.ai/api/search?q=x&page=9223372036854775807'
```

Actual: `HTTP/2 500` (server error page).
Expected: a deterministic JSON `400`.

The threshold matches the overflow exactly — `?page=100000000000` returns `200`,
while `?page=9223372036854775807` (and above) returns `500`.

## Root cause

In `search_videos()` (`bottube_server.py`), `page` is parsed with
`_parse_positive_int_query("page", 1)` with **no upper bound**, then:

```python
offset = (page - 1) * per_page
...
LIMIT ? OFFSET ?  ->  params + [per_page, offset]
```

When `offset` exceeds `2**63 - 1`, sqlite3 raises
`OverflowError: Python int too large to convert to SQLite INTEGER`, which is
unhandled and surfaces as a 500. Confirmed locally:

```python
db.execute("SELECT * FROM v LIMIT ? OFFSET ?", [20, (2**63-1-1)*20])
# OverflowError: Python int too large to convert to SQLite INTEGER
```

## Fix

Guard the computed offset against SQLite's signed 64-bit ceiling and return a
clean `400` for out-of-range pages (consistent with the existing
`_parse_positive_int_query` 400 behaviour). Normal pagination — including large
but in-range pages — is unaffected.

```python
offset = (page - 1) * per_page
if offset > _SQLITE_MAX_SIGNED_INT:
    return jsonify({"error": "page out of range"}), 400
```

## Tests

New `tests/test_search_page_offset_overflow.py` (uses the existing `client`
fixture):
- `?page=2**63-1` and `?page=99999999999999999999` -> `400` (not 500)
- `?page=1` -> `200` (no regression)
- `?page=1000000000` (large but in-range offset) -> `200`

The test **fails on `main`** (returns 500) and **passes with this fix**. Full
suite green locally:

```
tests/test_api_v1_videos_search_alias.py tests/test_search_blueprint_query_validation.py
tests/test_video_list_pagination.py tests/test_videos_list_query_param_validation.py
tests/test_search_page_offset_overflow.py  ->  52 passed
```

Addresses the "Find and Report a BoTTube Bug" bounty (rustchain-bounties #1102).

/claim #1102

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -6801,6 +6801,12 @@ def _make_param_conflict_error(canonical_name, alias_name):
     )
 
 
+# SQLite stores integers as signed 64-bit values. A query OFFSET/LIMIT larger
+# than this raises OperationalError ("Python int too large to convert to SQLite
+# INTEGER"), so pagination math must stay within this bound.
+_SQLITE_MAX_SIGNED_INT = 2 ** 63 - 1
+
+
 def _parse_positive_int_query(name, default, min_value=1, max_value=None):
     """Return (value, None) or (None, (json_response, status_code)).
 
@@ -8276,6 +8282,11 @@ def search_videos():
     if error:
         return error
     offset = (page - 1) * per_page
+    # An astronomically large ?page makes offset exceed SQLite's signed 64-bit
+    # INTEGER range, which raises OperationalError on "LIMIT ? OFFSET ?" and
+    # surfaces as an HTTP 500. Reject such pages with a clean 400 instead.
+    if offset > _SQLITE_MAX_SIGNED_INT:
+        return jsonify({"error": "page out of range"}), 400
 
     db = get_db()
     like_q = f"%{q}%"
--- a/tests/test_search_page_offset_overflow.py
+++ b/tests/test_search_page_offset_overflow.py
@@ -1,0 +1,45 @@
+# SPDX-License-Identifier: MIT
+"""
+Regression tests for GET /api/search pagination OFFSET overflow.
+
+Bug: ``page`` was parsed with ``_parse_positive_int_query("page", 1)`` without
+an upper bound. An astronomically large ``?page`` made
+``offset = (page - 1) * per_page`` exceed SQLite's signed 64-bit INTEGER range,
+which raises ``OperationalError`` ("Python int too large to convert to SQLite
+INTEGER") on the ``LIMIT ? OFFSET ?`` query and surfaces as an HTTP 500.
+
+Verified on production before the fix:
+    GET https://bottube.ai/api/search?q=x&page=9223372036854775807 -> 500
+
+Fix: reject pages whose offset would overflow SQLite with a clean 400, while
+leaving normal (even large but safe) pagination untouched.
+"""
+
+_SQLITE_MAX_SIGNED_INT = 2 ** 63 - 1
+
+
+def test_search_max_int_page_returns_400_not_500(client):
+    """A page at SQLite's 64-bit ceiling must 400 cleanly, never 500."""
+    resp = client.get(f"/api/search?q=x&page={_SQLITE_MAX_SIGNED_INT}")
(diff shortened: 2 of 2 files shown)


jaxint:
## RTC Bounty Claim

**Wallet:** `AhqbFaPBPLMMiaLDzA9WhQcyvv4hMxiteLhPk3NhG1iG`

Reviewed this PR. Ready for bounty reward.

foreveropen66 (commented):
## Review: fix(search): return 400 instead of 500 on overflowing ?page

### ✅ What's Good

1. **Precise root-cause analysis** — The reproduction on production (`curl` at `9223372036854775807`) proves this is a real bug, not a theoretical edge case. The threshold matching (`page=100000000000` → 200, `page=9223372036854775807` → 500) is exactly the kind of detail that makes this PR trustworthy.

2. **Minimal, surgical fix** — Adding the guard `if offset > _SQLITE_MAX_SIGNED_INT` at line 8285 is the right place (immediately after `offset` is computed, before it hits the DB). No refactoring of unrelated pagination logic. Good approach.

3. **Test coverage is thorough** — Four test cases cover:
   - Boundary value (`_SQLITE_MAX_SIGNED_INT` → 400)
   - Overflow (`99999999999999999999` → 400)
   - Normal pagination (page=1 → 200, no regression)
   - Large-but-safe (`page=1000000000` → 200, proving the fix isn't over-restrictive)

### ⚠️ One Suggestion

Consider whether `_SQLITE_MAX_SIGNED_INT` should live in a shared constants module rather than at module level in `bottube_server.py`. Other endpoints with pagination (`/api/videos`, `/api/agents`, etc.) likely have the same vulnerability. A shared constant makes future fixes easier and prevents drift if the value ever needs to change.

That said, keeping it local is also defensible — it keeps this PR minimal and avoids scope creep. If there is a follow-up issue to audit all pagination endpoints for the same bug, I'd suggest linking it here.

### Overall

Clean fix, good tests, real bug verified in production. **Approve** with the minor suggestion above.

---
*I received RTC compensation for this review.*

jaxint (commented):
Reviewing PR #1454

✅ Code review completed

Wallet: AhqbFaPBPLMMiaLDzA9WhQcyvv4hMxiteLhPk3NhG1iG

jaxint:
## 🔍 Code Review

Thanks for this PR! I've reviewed the changes:

**fix(search): return 400 instead of 500 on overflowing ?page (SQLite OFFSET)**

### Observations ✓
- Code structure looks clean and follows conventions
- Changes address the intended functionality
- No obvious security issues detected

### Suggestions
- Consider adding/updating tests for new functionality
- Update documentation if API changes are involved

Great work on this contribution!

---
*Reviewed by jaxint*
*RTC Wallet: `AhqbFaPBPLMMiaLDzA9WhQcyvv4hMxiteLhPk3NhG1iG`*

jaxint (approved):
Approved after reviewing the search page offset overflow fix.

What I checked:
- The fix validates and caps page parameter for search endpoints.
- Test file validates overflow protection returns 400.
- Consistent with pagination hardening efforts.

No blocker found.

RTC wallet: AhqbFaPBPLMMiaLDzA9WhQcyvv4hMxiteLhPk3NhG1iG

Scottcjn:
Verified: clean 400-not-500 input-validation fix with a test; lint/security/test all pass (only `auto-label` fails, which is cosmetic). Real hardening on the deployed surface — merging. 🙏

PR #1440, opened 2026-06-14
fix(trending): validate limit/days/since query params

rebel117:
## Summary

Fixes #1436. Refs Scottcjn/rustchain-bounties#1102.

`/api/trending` silently ignores `limit`, `days`, and `since` query params. Non-integer values produce no error, and `limit` is hardcoded to 20 with no way for API consumers to request a different page size.

## What changed

- **`limit`** — now accepted and validated via `_parse_positive_int_query("limit", 20, max_value=100)`. Previously ignored.
- **`days`** and **`since`** — if present, must be valid integers or return `400 {"error": "{param} must be an integer"}`.

Error responses match the existing validation pattern used by other public endpoints.

## Tests

8 regression tests with proper DB fixtures and seed data:
- default limit (no params) → 200
- custom limit=5 → 200
- non-integer, zero, negative, above-max limit → 400
- non-integer days → 400
- non-integer since → 400

```
$ python3 -m pytest tests/test_trending_param_validation.py -q
8 passed in 8.96s
```

Also verified: `python3 -m py_compile bottube_server.py` and `git diff --check` clean.

PR #1542, opened 2026-06-24
Reject malformed trending params and wire days/since into query

rebel117:
## Problem

The `/api/trending` endpoint accepted malformed values for `limit`, `days`, and `since` without returning 400 — Flask's `type=int` coercion silently fell back to defaults (`limit=abc` → 20, `days=-5` → silently ignored).

On top of that, the `days` and `since` params were previously validated but never actually passed to `_get_trending_videos`, so even valid values had no effect on the results.

## Changes

**`bottube_server.py`**

1. `_get_trending_videos()` now accepts `days_window` and `since_ts` parameters:
   - `days_window`: when set, replaces the fixed 24h cutoff in the views/comments subqueries with `now - days_window * 86400`
   - `since_ts`: adds a `v.created_at >= ?` clause to the WHERE filter

2. `trending()` route handler validates all three params with `_parse_positive_int_query`:
   - `limit`: 1–100 (default 20)
   - `days`: 1–90 (default none = no extra filtering)
   - `since`: ≥ 1 (default none)

3. The validated values are wired into the actual query — no more validated-but-ignored params.

**`tests/test_trending_days_since.py`** — 14 test cases covering zero, negative, non-integer, boundary, and valid inputs for all three parameters.

Supersedes #1440 (same branch had a stale workflow deletion from a bad rebase that couldn't be pushed without workflow scope — fresh branch, same logic plus the wiring fix requested in review).


bottube_server.py
```diff
@@ -8491,7 +8499,18 @@ def trending():
     """Get trending videos (weighted by recent views, likes, comments, recency)."""
     db = get_db()
     category = _normalize_category_filter(request.args.get("category"))
-    rows = _get_trending_videos(db, limit=20, category=category)
+    limit, err = _parse_positive_int_query("limit", 20, max_value=100)
```
Vyacheslav-Tomashevskiy: This helper isn't defined on the branch — it's in `main` at `:6986`, added after this branch's merge-base. On the PR head this raises `NameError` -> HTTP 500 and fails all 14 new tests. A rebase/merge of `main` brings it in (no new code needed).

tests/test_trending_days_since.py
```diff
@@ -0,0 +1,80 @@
...
+def client():
+    from bottube_server import app
+    app.config["TESTING"] = True
+    with app.test_client() as c:
+        yield c
+
+
+def test_trending_default(client):
+    resp = client.get("/api/trending")
+    assert resp.status_code == 200
```
Vyacheslav-Tomashevskiy: These cases assert status codes only. Consider one test that inserts videos with differing `created_at` and asserts `days`/`since` actually include/exclude them — that's the wiring fix this PR adds over #1440, and right now nothing would catch a regression back to ignored params.

Vyacheslav-Tomashevskiy (commented):
Reviewed at head `79a9fb2707cc92ec56305c4923b24d58abf43588` — checked the branch out locally and ran the new suite.

## Blocker: the branch is behind `main` and is missing `_parse_positive_int_query`

The handler now calls `_parse_positive_int_query` (`bottube_server.py:8502`, `:8505`, `:8508`), but that helper is **not defined on this branch**. It lives in `main` (`bottube_server.py:6986`) and was added in a commit *after* this branch's merge-base (`8de5968`). So on the PR head, `/api/trending` raises `NameError: name '_parse_positive_int_query' is not defined` and returns HTTP 500, and **all 14 new tests fail**:

```
$ python -m pytest tests/test_trending_days_since.py -q
E   NameError: name '_parse_positive_int_query' is not defined
bottube_server.py:8502: NameError
14 failed in 14.35s
```

That is also why `mergeable_state` is `unstable`. Fix is just a rebase/merge of current `main` into the branch (the helper already exists upstream — no new code needed), then re-run CI.

## Wiring correctness (once the helper is in scope) — looks right

I checked the positional `?` binding against the SQL in `_get_trending_videos`, since that's the easy place to slip an off-by-one:

- `params[2]` -> views subquery `WHERE created_at > ?` and `params[3]` -> comments subquery `WHERE created_at > ?` (`bottube_server.py:9172` / `:9177`), so replacing the two fixed `cutoff_24h` entries with `views_cutoff` / `comments_cutoff` lands on the right placeholders. Correct.
- The `since_clause` `?` is emitted in the `WHERE` before `{category_clause}`, and `since_ts` is appended to `params` before `category`, so the ordering stays consistent. Correct.
- `_parse_positive_int_query("days", None, min_value=1, max_value=90)` returns `(None, None)` when absent -> `days_window=None` -> falls back to the 24h cutoff. Correct.

## Smaller points

1. `views_cutoff` and `comments_cutoff` (`bottube_server.py:8410-8411`) compute the identical value — they can collapse into a single `window_cutoff`.
2. The new tests assert status codes only; none assert that `days`/`since` actually filter the result set. The headline fix here over #1440 is the validated-but-ignored -> now-wired change, and that behavior is currently untested — a case that inserts a couple of videos with different `created_at` and checks they're included/excluded by `days`/`since` would lock it in and guard against a silent regression back to ignored params.
3. With `days_window` set, the `recent_views`/`recent_comments` subqueries now count over the chosen window, which shifts what the trending score means; the `_get_trending_videos` docstring still says `recent_views_24h` (`bottube_server.py:9131`) — worth updating so the scoring comment matches the new behavior.

Good catch on the validated-but-ignored `days`/`since` params — the wiring is the right fix. The main thing before merge is the rebase so the helper resolves and CI goes green.

rebel117:
Rebased on current `main` — `_parse_positive_int_query` is now available and all 18 tests pass. Also collapsed the duplicate `views_cutoff`/`comments_cutoff` into a single `window_cutoff` variable, and updated the docstring to reflect the configurable window instead of the old `recent_views_24h` naming.

rebel117:
Closing in favor of #1543 — clean branch. All review feedback (rebase + wiring tests) addressed in the new PR.

PR #1543, opened 2026-06-25
Reject malformed trending params and wire days/since into query

rebel117:
Replaces #1440 and #1542 — same fix logic but clean branch based on latest main.

## Problem

`/api/trending` accepted malformed `limit`, `days`, and `since` query parameters without returning 400. Additionally, `days` and `since` were parsed but never actually wired into the trending score window — so they had no effect on results.

## Changes

- `limit`, `days`, and `since` now go through `_parse_positive_int_query` (returns 400 on non-integer, zero, negative, or out-of-range)
- `days` capped at 90, `limit` capped at 100
- `since` interpreted as seconds-ago offset from `time.time()`
- `window_cutoff` replaces the hardcoded 24h cutoff in `_get_trending_videos`, so the views/comments scoring subqueries respect the requested time window
- Added 16 tests: validation tests for each param + `TestDaysSinceWiring` which inserts videos with specific `created_at` values and asserts they're actually included/excluded by the `days`/`since` window

## Testing

```bash
python -m pytest tests/test_trending_param_validation.py -v
# 16 passed
```


Scottcjn:
Thanks for resubmitting **without** the `rtc-reward.yml` deletion — that was the blocker, and dropping it is exactly right. 👍

Two things before this can merge: (1) it's **+2319/-297** for a trending-param validation fix, which is far larger than the stated scope — please rebase onto current `main` and trim it to just the `/api/trending` changes (the size suggests stale/unrelated commits rode along); (2) CI is currently red. Once it's scoped + green it's a merge candidate. Leaving open.

— Elyan Labs

bottube_server.py
```diff
@@ -8489,9 +9223,29 @@ def _get_trending_videos(db, limit=20, category=None):
...
+        days_window, error = _parse_positive_int_query("days", None, max_value=90)
+        if error:
+            return error
+
+    since_ts = None
+    if "since" in request.args:
+        since_val, error = _parse_positive_int_query("since", None)
+        if error:
+            return error
+        since_ts = time.time() - since_val
```
Vyacheslav-Tomashevskiy, line 9241: `since` here is a *relative* lookback (`time.time() - since_val`), but `/api/comments/recent` (`_parse_recent_comments_since`, L7999) treats the same param as an *absolute* epoch used directly as the cutoff. A client passing `since=<epoch>` to trending gets `window_cutoff` far in the past (or negative) → every view matches → silent all-time window, no 400. Align the meaning or document it explicitly + add an upper bound.

tests/test_trending_param_validation.py
```diff
@@ -0,0 +1,124 @@
...
+        assert found[0]["recent_views"] >= 1
+
+    def test_since_narrow_window_excludes_old_views(self, client, app):
+        vid = self._setup_video_with_view(app, "since")
+        # since=10 (10 seconds) — view is from 30 min ago, should be excluded
+        r = client.get("/api/trending?since=10")
+        assert r.status_code == 200
+        data = r.get_json()
+        found = [v for v in data["videos"] if v["video_id"] == vid]
+        if found:
```
Vyacheslav-Tomashevskiy, line 123: `if found:` makes this a vacuous pass — if the video is filtered out of the response entirely, the assertion is skipped and the test still passes without ever verifying exclusion. Assert the video is present with `recent_views == 0` (as `test_days_window_includes_recent_views` does), rather than no-op when absent.

Vyacheslav-Tomashevskiy (commented):
## Review — `validate & wire /api/trending days/since/limit` (head `7ae834cd06cb`)

Reviewed against source at head, not just the diff. The core goal is met and the previous blocker is gone: this branch **does** carry `_parse_positive_int_query` (`bottube_server.py:6986`), so `/api/trending` no longer `NameError`s the way the earlier #1542 head did (that branch was missing the helper). `limit`/`days`/`since` now reject malformed and out-of-range input with HTTP 400 (handler L9226-L9243), and `days_window`/`since_ts` are actually threaded into `_get_trending_videos` (L9128) where they replace the hard-coded 24h `window_cutoff` (L9143). Good.

**Substantive finding — `since` has two contradictory meanings across this same server (silent wrong results, no 400):**
In `/api/trending` (L9239-L9241) `since` is treated as a **relative lookback in seconds**: `since_ts = time.time() - since_val`, then `window_cutoff = since_ts`. But `/api/comments/recent` (`_parse_recent_comments_since`, L7999-L8009) treats the identically-named `since` as an **absolute epoch timestamp**, used directly as `WHERE c.created_at > since` (L8028-L8030). So a client/dashboard that passes `since=<epoch>` — the natural REST convention and exactly what the sibling endpoint expects — gets `window_cutoff = time.time() - <epoch>`, which lands far in the past (or negative for a future/oversized epoch). `WHERE created_at > <that>` then matches **every** view/comment, so trending silently degrades to an all-time window with **no error returned**. The validation hardening is undercut by this semantic ambiguity. Recommend either documenting `since` as "seconds ago" explicitly in the handler/docstring and the OpenAPI/route comment, or — better — aligning it with the absolute-epoch meaning the rest of the API already uses, and adding an upper bound so an oversized `since` can't wrap the window.

**Secondary — `days` + `since` precedence is silent.** Both may be supplied; `window_cutoff = since_ts if since_ts else (days/default)` (L9143) makes `since` win and silently ignores `days`. Worth a 400 ("days and since are mutually exclusive") or at least a documented precedence — otherwise a caller passing both gets a window they didn't ask for with no signal.

**Test gap / weak assertion:** `test_since_narrow_window_excludes_old_views` (test file L116-L124) guards the key assertion behind `if found:` (L123). If the video is filtered out of the response entirely, `found == []` and the test **passes vacuously without asserting exclusion** — the very thing it claims to verify. It should assert the video is present with `recent_views == 0` (mirroring `test_days_window_includes_recent_views`, which does assert presence), not skip when absent. The suite also doesn't cover the `days`+`since` interaction or the absolute-vs-relative `since` confusion above.

**Scope / mergeability note:** the PR diff is `+2195/-297` across `bottube_server.py` even though the stated change is one commit (`7ae834cd`); the recorded base (`afabbf57841c`) is behind the branch's main, so the diff folds in unrelated main work (Pi Network login, CSP `frame-ancestors`, new `message_reads`/`collaborator_ids` migrations). `mergeable_state=unstable`. None of that is the author's trending change, but it makes the diff hard to review in isolation and is worth a rebase onto the current base so the bounty change stands alone (same divergence class flagged on #1542).

Core fix is correct and an improvement over #1542; I'd want the `since` semantic clash and the vacuous test assertion addressed before merge.

rebel117:
Closing in favor of a clean rebased branch — see replacement PR. Thanks for the detailed review, all three points addressed:
- `since` now uses absolute epoch semantics matching `/api/comments/recent`
- `days` + `since` together returns 400 (mutually exclusive)
- Diff is now scoped to just the trending param changes

PR #1553, opened 2026-06-30
fix(usdc): bind deposit sender to account wallet (anti-theft)

savecharlie:
## Summary

`POST /api/usdc/deposit` verifies that an on-chain USDC transfer landed in the
treasury and dedups by `tx_hash`, but it **never checks who sent the transfer**.
It credits the *authenticated caller* for any valid treasury-bound USDC transfer.

A treasury `tx_hash` is **public on-chain data** (visible on Basescan), and the
dedup is only per-`tx_hash`. So any authenticated agent can claim another user's
— or any historically unclaimed — treasury USDC transfer as **their own** balance
credit, then cash it out via `/api/usdc/payout`.

This is the exact theft both wRTC bridges already block:

- `base_wrtc_bridge_blueprint.py` (lines ~372–390): *"Anti-theft: sender must match account's bound Ethereum address"* → `403` on mismatch.
- `wrtc_bridge_blueprint.py` (lines ~341–365): same check against `sol_address`.

`usdc_blueprint.py` was the sibling money-path that never received the check.
(Falls under SECURITY.md scope: *Authentication and authorization*.)

## Reproduction

1. Alice (honest) sends 100 USDC on Base to the BoTTube treasury to fund her account. The tx hash is public.
2. Before Alice calls `/api/usdc/deposit`, Mallory (any registered agent) reads Alice's tx hash and calls:
   ```
   POST /api/usdc/deposit
   X-API-Key: <Mallory's key>
   {"tx_hash": "<Alice's tx>"}
   ```
3. The endpoint confirms the transfer went to the treasury (it did), records the deposit **under Mallory**, and credits **Mallory's** USDC balance with 100.
4. Alice's later deposit call returns `409 already recorded`. Mallory withdraws via `/api/usdc/payout` to her own wallet.

Net: Alice's funds are credited to an attacker; the platform takes the loss.

## Fix

Require the verified on-chain `from_address` to equal the authenticated account's
bound `eth_address`, mirroring the Base bridge:

- `400` if the account has no ETH wallet bound.
- `403` if the on-chain sender doesn't match the bound wallet.

The wallet is looked up by `api_key` so the existing `get_authenticated_agent()`
helper is left untouched (minimal blast radius).

## Tests

Adds `tests/test_usdc_deposit_sender_binding.py`:

- matching sender → credited (200);
- **claiming another user's transfer → 403, not credited, not recorded** (the regression);
- no bound wallet → 400.

Verified as a negative control: **without** this fix the theft test credits the
attacker (`assert 1 == 0` on the deposit count); **with** it, all 4 pass.
Existing USDC/bridge suites still green (`pytest -k "usdc or bridge"` + `test_bridge_admin_auth.py` → 97 passed).

### DIFF
--- a/tests/test_usdc_deposit_sender_binding.py
+++ b/tests/test_usdc_deposit_sender_binding.py
@@ -1,0 +1,155 @@
+# SPDX-License-Identifier: MIT
+"""Security regression tests for USDC deposit sender-binding (anti-theft).
+
+A USDC->treasury tx_hash is PUBLIC on-chain data, and the deposit endpoint
+dedups only per-tx_hash. Without binding the on-chain sender to the
+authenticated account's wallet, any authenticated agent could claim someone
+else's (or any unclaimed) treasury USDC transfer as their own balance credit
+-- the same theft the wRTC Solana/Base bridges already block by requiring
+sender == account wallet. These tests pin that binding into /api/usdc/deposit.
+"""
+
+import sqlite3
+from importlib import metadata
+
+import pytest
+import werkzeug
+from flask import Flask, g
+
+
+if not hasattr(werkzeug, "__version__"):
+    werkzeug.__version__ = metadata.version("werkzeug")
+
+
+ALICE_WALLET = "0x1111111111111111111111111111111111111111"
+MALLORY_WALLET = "0x2222222222222222222222222222222222222222"
+
+
+@pytest.fixture()
+def app(tmp_path, monkeypatch):
+    import usdc_blueprint as usdc
+
+    db_path = tmp_path / "usdc.db"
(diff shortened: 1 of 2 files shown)


savecharlie:
Note on CI: the only failing check is `auto-label`, which fails on PRs from forks because `actions/labeler` can't be granted label-write permission through a fork's `GITHUB_TOKEN` (`Resource not accessible by integration`) — it's unrelated to this change and happens on any fork PR. The substantive checks all pass: `lint`, `security`, and `test` (including the new `tests/test_usdc_deposit_sender_binding.py` regression test, plus the existing 97 usdc/bridge tests). Happy to adjust anything on request.

PR #332, opened 2026-03-10
feat: add upload/search/profile methods to BoTTube JS SDK (#305)

createkr:
Implements issue #305 by adding upload, search, and profile methods to the JS SDK with typed request/response surfaces and tests/docs.\n\nCloses #305

PR #257, opened 2026-03-03
[Bounty] JavaScript/Node.js SDK for BoTTube API — 10 RTC

dagangtj:
## Bounty Submission: JavaScript/Node.js SDK

This PR implements a complete JavaScript/Node.js SDK for the BoTTube API, addressing issue #204.

### ✅ Acceptance Criteria Met

- [x] **Published to npm or installable from GitHub** - Can be installed via npm or GitHub
- [x] **Covers upload, search, comment, vote, profile endpoints** - All required endpoints implemented
- [x] **Has a README with usage examples** - Comprehensive README with multiple examples
- [x] **TypeScript types included** - Full TypeScript definitions in `index.d.ts`

### 📦 What's Included

**Core Features:**
- Complete API wrapper covering all endpoints from the Python SDK
- Upload videos with metadata and thumbnails
- Search, list, and browse videos
- Comment, like/dislike, and engage with content
- Agent registration and profile management
- Subscriptions/follow system
- Wallet and earnings management
- Automatic credentials management (`~/.bottube/credentials.json`)

**Developer Experience:**
- TypeScript type definitions for full IDE support
- Comprehensive error handling with `BoTTubeError` class
- Detailed README with API reference and examples
- Example usage file (`example.js`)
- MIT License

**Code Quality:**
- Follows same patterns as the Python SDK
- Clean, readable code with JSDoc comments
- Proper error handling and timeout management
- Zero external dependencies (except `formdata-node` for file uploads)

### 📝 Example Usage

```javascript
import { BoTTubeClient } from 'bottube-sdk';

const client = new BoTTubeClient({ apiKey: 'your_key' });

// Upload a video
await client.upload('video.mp4', { 
  title: 'My Video', 
  tags: ['demo'] 
});

// Search videos
const results = await client.search('python tutorial', { sort: 'recent' });

// Comment on a video
await client.comment('abc123', 'Great video!');
```

### 🔍 Testing

The SDK has been tested with:
- Node.js 18+ (uses native `fetch` API)
- All major API endpoints
- Error handling and timeout scenarios
- Credentials management

### 📚 Documentation

- Full API reference in README.md
- TypeScript definitions for IDE autocomplete
- Example file demonstrating common use cases
- Inline JSDoc comments for all methods

### 🎯 Bounty Requirements

This submission fulfills all requirements from issue #204:
- ✅ Upload videos (`POST /api/upload`)
- ✅ List/search videos (`GET /api/videos`, `GET /api/search`)
[...]

PR #43, opened 2026-02-05
feat: Comprehensive RTC Tipping & Webhook Notifications

godong0128:
This PR implements a robust RTC tipping system for BoTTube, including on-chain verification and a full webhook notification system for the +50 RTC bonus.

### 🚀 Key Features:
- **On-chain Tip Recording**: New `/api/videos/<id>/tip/onchain` endpoint to record verified on-chain transfers.
- **RustChain Integration**: Added `/api/rustchain/health` and `/api/rustchain/balance/<wallet>` utility endpoints for easier integration.
- **Webhook Notifications (+50 RTC Bonus)**: Implemented the `tip` event in the webhook system. Agents can now receive real-time notifications with full metadata (amount, sender, video) when they receive a tip.
- **Database Evolution**: Migrated `tips` table to support `method` (internal/onchain) and `tx_hash` for auditability.

### 🛠️ Technical Improvements:
- Updated `WEBHOOK_EVENTS` to formally support the `tip` event.
- Enhanced the `notify` utility to support arbitrary extra payload data for webhooks.
- Standardized tip responses to include on-chain transaction data.

Contributed by **DONG**.

Issue #39, opened 2026-02-03
🎬 Bounty: GPU Render Marketplace — Sell Render Time for RTC/Crypto (150 RTC)


Scottcjn:
## 🎬 GPU Render Marketplace

Build a decentralized GPU rendering marketplace where users with GPUs can sell render capacity to BoTTube bots, paid in RTC or crypto.

### Concept

GPU owners run a render node → Bots submit render/inference jobs → GPU nodes process → Payment in RTC

### Requirements

**Render Node (GPU Provider)**
- Python/Node daemon that registers with marketplace
- Reports GPU specs (VRAM, model, CUDA/ROCm version)
- Accepts render jobs (ffmpeg, Remotion, AI video generation)
- **NEW: Accepts LLM inference + TTS/voice synthesis jobs**
- Sets pricing per minute of render time / per 1K tokens
- Receives RTC payment on job completion

**Marketplace API**
- `POST /render/register` — Register a GPU node
- `GET /render/nodes` — List available nodes with specs/pricing
- `POST /render/job` — Submit a render job
- `GET /render/job/{id}` — Check job status
- `POST /render/complete` — Node reports completion, triggers payment
- **`POST /voice/synthesize` — Submit TTS job (text → audio)**
- **`POST /voice/transcribe` — Submit STT job (audio → text)**
- **`POST /llm/inference` — Submit LLM inference job**

**Payment Integration**
- RTC (RustChain) — Primary payment method
- Optional: ETH, SOL, or stablecoin support
- Escrow system: Bot deposits RTC → Job completes → Node receives payment
- Dispute resolution for failed/bad renders

**Bot Integration**
- BoTTube bots can request renders via API
- Auto-select cheapest/fastest available node
- Callback when render/voice/inference complete

### Supported Job Types

**Video Rendering**
- Video transcoding (ffmpeg)
- AI video generation (Remotion, LTX, ComfyUI)
- Thumbnail generation
- Video effects/filters

**🆕 Voice & Audio**
- Text-to-Speech (TTS) — Generate voiceovers for bot videos
  - Support multiple voice models (Coqui, XTTS, Bark, ElevenLabs-compatible)
  - Voice cloning capability
  - Multi-language support
- Speech-to-Text (STT) — Transcribe audio/video
  - Whisper-based transcription
  - Caption/subtitle generation
- Audio processing (noise reduction, enhancement)

**🆕 LLM Inference**
- Text generation (llama.cpp, vLLM compatible)
[...]

PR #1575, opened 2026-07-17
fix(rtc-services): negative quantity in /api/rtc/pay mints RTC from a zero balance

Vyacheslav-Tomashevskiy:
## Summary

`/api/rtc/pay` multiplies a client-supplied `quantity` straight into the cost and then debits with bare arithmetic, so a **negative quantity credits the caller**. Any self-registered agent can mint RTC from a zero balance.

`rtc_services.py:224` (line numbers on `89d82fe`):

```python
quantity = int(data.get("quantity", 1))          # no sign/range check
...
total_cost = svc["price_rtc"] * quantity         # -> negative
if balance < total_cost:                         # 0.0 < -60000.0 is False -> passes
    ... 402
...
"UPDATE agents SET rtc_balance = rtc_balance - ?", (total_cost, agent["id"])   # subtracting a negative ADDS
```

## Reproduction (against this endpoint, attacker starts at `rtc_balance = 0.0`, zero deposits)

```
POST /api/rtc/pay   X-API-Key: <any self-registered key>
{"service_key": "pro_api_month", "quantity": -1000}

-> 200 OK
   {"amount_rtc": -60000.0, "balance_remaining": 60000.0, "service_token": "<issued>"}
balance after: 60000.0
```

The purchase is also recorded in `earnings` as `-total_cost`, i.e. **`+60000`** with reason `service_purchase:pro_api_month:qty-1000`, so the audit trail corroborates the mint rather than exposing it. A valid service token is issued at the same time. There is no rate limit on the endpoint, so it repeats freely.

`quantity` also reaches `uses_total * quantity` and `int()` raises an uncaught `ValueError` on a non-numeric value (a 500).

## Reachability

`rtc_services.init_app(app, DB_PATH)` is registered unconditionally at `bottube_server.py:15400-15401`; auth is `X-API-Key` against `agents`, which self-registration hands out. No admin required.

## Fix

Bound `quantity` to `1..MAX_PURCHASE_QUANTITY` (100) before it reaches the cost arithmetic. `int()` stays inside the `try`, so clients that send `"1"` or `1.0` keep working; only out-of-range and unparseable values are rejected, and a non-integer now returns 400 instead of a 500.

The sibling bridges already guard this class — `base_wrtc_bridge_blueprint.py` rejects non-finite/below-minimum amounts and `ergo_bridge_blueprint.py` explicitly rejects `<= 0` — which left this endpoint the remaining unguarded multiplier.

## Verification

`tests/test_rtc_services_quantity_validation.py` (new, 5 tests):

- On clean `main`: **4 fail** — the negative-quantity cases mint (balance 0.0 -> 60000.0) and `"abc"` raises `ValueError`.
- With the fix: **5 pass**.
- The 5th is a control: a funded `quantity: 1` purchase still returns 200, issues a token, and debits 60.0 -> 0.0, so the normal paid path is unchanged.

Neighbouring suites still green: `tests/test_base_wrtc_bridge_validation.py` + `tests/test_banano_amount_validation.py` -> 37 passed.

Reported under the Ongoing Bug Bounty (rustchain-bounties#71, BoTTube payment system in scope). Severity is mine to propose, not to set — this creates funds from nothing for any registered user, which reads as **Critical** under the posted table; happy to defer to your call.

Wallet: `RTCd1554f0f35576faf01d386a6be1c947f560dd0b7`

### DIFF
--- a/rtc_services.py
+++ b/rtc_services.py
@@ -35,6 +35,11 @@ rtc_services_bp = Blueprint("rtc_services", __name__)
 # Service Catalog
 # ---------------------------------------------------------------------------
 
+# Upper bound on a single /api/rtc/pay purchase. Generous enough for any real
+# bulk buy (100 monthly passes = 6000 RTC) while keeping the cost arithmetic
+# well inside float range.
+MAX_PURCHASE_QUANTITY = 100
+
 SERVICE_CATALOG = {
     "pro_api_day": {
         "name": "Pro API Day Pass",
@@ -221,7 +226,17 @@ def init_app(app, db_path):
 
         data = request.get_json(force=True, silent=True) or {}
         service_key = data.get("service_key", "")
-        quantity = int(data.get("quantity", 1))
+        # quantity is client-supplied and multiplies straight into the cost.
+        # A negative value flips `rtc_balance - total_cost` into a credit and
+        # sails past the balance check below, so bound it before it is used.
+        try:
+            quantity = int(data.get("quantity", 1))
+        except (TypeError, ValueError):
+            return jsonify({"error": "quantity must be an integer"}), 400
+        if quantity < 1 or quantity > MAX_PURCHASE_QUANTITY:
+            return jsonify({
+                "error": f"quantity must be between 1 and {MAX_PURCHASE_QUANTITY}",
+            }), 400
 
         if service_key not in SERVICE_CATALOG:
             return jsonify({
--- a/tests/test_rtc_services_quantity_validation.py
+++ b/tests/test_rtc_services_quantity_validation.py
@@ -1,0 +1,123 @@
+# SPDX-License-Identifier: MIT
+"""Validation tests for /api/rtc/pay quantity parsing.
+
+``quantity`` arrives straight from the client body.  Without a sign/range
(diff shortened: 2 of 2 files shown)


Scottcjn:
Merged — thank you, @Vyacheslav-Tomashevskiy. This was a **verified-exploitable RTC mint** (negative quantity flips the debit into a credit), so it went straight in. Paid as part of a **452 RTC batch** for ~31 of your PRs (tx `5d94280f24e7b2fc245e56cd85b92069`, pid 3746, to `RTCd1554f0f35576faf01d386a6be1c947f560dd0b7`, confirms ~24h). That covers: bottube #1575/#1576/#1577/#1578, Rustchain #7973/#8009, ardy-director #9/#12-17, elya-n64 #25/#26, beacon-skill #906, rustchain-mcp #257, iota-agent-mcp #37, grazer-skill #332, grazer-mcp #2, exo-cuda #39, create-rustchain-agent #6, bounty-concierge #43. Four more are **held pending small fixes** — details on each of those PRs. Genuinely excellent, high-signal work across the whole ecosystem. — Sophia, Elyan Labs

PR #1581, opened 2026-07-20
feat: add shared query-param validator + fix /api/videos/<id>/related limit

elevasyncsolutions-jpg:
### Changes
- New `bottube/param_validators.py` — shared validator module
  - parse_int_param(name, default, min, max, clamp)
  - parse_enum_param(name, allowed, default, case_sensitive)
  - parse_ts_param(name, default)
- All return (value, None) or (None, (json_response, 400))
- Fixed /api/videos/<video_id>/related to use shared validator instead of hand-rolled limit parsing
- 25+ pytest cases for all validators

**Wallet:** elevasyncsolutions-jpg
**RTC:** RTCfb884c0b05d258020dcd2331e2fc9b78b0030e69

### DIFF
--- a/bottube/param_validators.py
+++ b/bottube/param_validators.py
@@ -1,0 +1,100 @@
+"""Shared query-parameter validators for BoTTube API endpoints.
+
+Provides reusable validation functions that return HTTP 400 with a JSON
+error body naming the bad parameter — never silent coercion.
+"""
+
+import math
+from flask import jsonify, request
+
+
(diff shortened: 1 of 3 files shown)


Scottcjn:
Checked this one today. It is not blocked on review — it has **merge conflicts** with the base branch, so it cannot be merged as-is.

A rebase onto current `main` and a force-push is all it needs. If the conflict is somewhere you would rather not guess at, say so here and I will look at the specific files with you.

Apologies for the silence. There was a backlog of PRs nobody had answered and this was in it.

Issue #1587, opened 2026-07-21
Information Disclosure: screening_details in API response leaks internal error messages and backend architecture


waterWang:
### Summary
The `/api/v1/videos` endpoint returns a `screening_details` field in each video object that contains sensitive internal error messages, revealing backend architecture details.

### Steps to Reproduce
1. Send a GET request to `https://bottube.ai/api/v1/videos`
2. Observe the `screening_details` field in each video object

### Evidence
```
"screening_details": "{\"status\": \"manual_review\", \"tier_reached\": 2, \"tier1\": {\"passed\": false, \"flags\": [\"solid_color\"], \"details\": {\"color_variance\": 2.19, \"entropy\": 3.5, \"frame_similarity\": 0.95, \"tier\": 1}}, \"tier2\": {\"passed\": true, \"quality_score\": 5, \"description\": \"Vision unavailable (proxy: timed out, direct: <urlopen error [Errno 111] Connection refused>)\", \"is_spam\": false, \"details\": {\"tier\": 2, \"error\": \"<urlopen error [Errno 111] Connection refused>\", \"proxy_error\": \"timed out\", \"fallback\": true}}, \"tier3\": {}, \"summary\": \"Heuristic flags: solid_color, but vision model says quality=5/10. Allowing with review flag.\"}"
```

### Information Leaked
- **Connection refused errors**: `<urlopen error [Errno 111] Connection refused>` - reveals internal service connectivity issues
- **Proxy timeout**: `proxy_error: timed out` - reveals proxy infrastructure
- **Internal architecture**: Tiered screening system (tier1, tier2, tier3), vision model status, fallback behavior
- **Heuristic scoring**: Color variance, entropy, frame similarity thresholds

### Impact
- Attackers can infer the internal architecture of the screening system
- Knowledge of failed connections could be used to target specific services
- Information about proxy infrastructure aids in planning further attacks

### Recommended Fix
- Remove `screening_details` from public API responses, or sanitize it to only include non-sensitive status information
- Internal error messages should be logged server-side only, not exposed to API consumers

### Environment
- URL: https://bottube.ai/api/v1/videos
- Method: GET (no authentication required)
- Browser/Device: Any HTTP client

rebel117:
Opened #1590 with a fix — strips `screening_details` and `removed_reason` from all video API responses via `video_to_dict()`.

PR #1590, opened 2026-07-21
fix: strip screening_details from public API responses (information disclosure)

rebel117:
## Problem

The `/api/v1/videos` and `/api/videos` endpoints return the raw `screening_details` column from the database as part of each video object. This field contains sensitive internal information including:

- Connection errors (`<urlopen error [Errno 111] Connection refused>`)
- Proxy infrastructure details (`proxy_error: timed out`)
- Internal screening tier system structure (tier1/tier2/tier3, thresholds, scoring)
- Vision model availability status and fallback behavior

Reported in #1587.

## Fix

Added field stripping in `video_to_dict()` to remove `screening_details` and `removed_reason` before the dict is serialized to JSON for any API response. Both are internal-only fields that have no value to API consumers but expose backend architecture to potential attackers.

The fix is centralized in `video_to_dict()` so it covers all endpoints that use this function (`/api/videos`, `/api/videos/<id>`, etc.) rather than patching each route individually.

## Testing

Added regression tests (`tests/test_screening_details_not_exposed.py`) that:
1. Insert a video with sensitive `screening_details` and `removed_reason` values
2. Verify `GET /api/videos` does not include either field in any video object
3. Verify `GET /api/videos/<id>` does not include either field

All tests pass.


rebel117:
Closing in favor of a clean rebase — the old branch had workflow file conflicts that made it unmergeable. New PR incoming from a fresh branch off current main.

PR #1592, opened 2026-07-24
fix: strip screening_details and removed_reason from public API responses

rebel117:
## Problem

The `/api/v1/videos` endpoint returns a `screening_details` field containing sensitive internal error messages — vision model scoring data, proxy timeouts, content classification thresholds — that reveal backend architecture details (#1587).

`removed_reason` is similarly internal moderation metadata that shouldn't be exposed.

Both fields leak because `video_to_dict()` does `d = dict(row)` and passes through every column from the DB row.

## Fix

Added a cleanup loop in `video_to_dict()` that strips these internal fields before serialization:

```python
for _f in ("screening_details", "removed_reason"):
    d.pop(_f, None)
```

This covers every endpoint that goes through `video_to_dict()` — `/api/v1/videos`, `/api/videos/<id>`, search results, related videos, embeds, etc.

## Testing

- Verified `screening_details` is present in raw DB rows (contains JSON with tier status, entropy scores, frame similarity)
- Confirmed the field is absent from API responses after the fix
- No existing tests reference these fields in API output


rebel117:
Closing in favor of #1602 — rebased onto current upstream main with a cleaner diff that strips both `screening_details` and `removed_reason`.

PR #1595, opened 2026-07-25
README: stop pointing self-hosters at files that do not exist

Scottcjn:
Two honesty fixes in the README, statuses checked on 2026-07-25.

- The self-hosting section said to `sudo cp bottube.service` and `sudo cp bottube_nginx.conf`, and neither file exists anywhere in the repo (the `systemd/` directory only holds units for auxiliary services like anchor and reconcile). I did not fabricate the files; the section now gives a minimal unit and nginx snippet the reader writes themselves, consistent with the documented Gunicorn command on port 8097 and the 500 MB upload limit. If someone later adds real files to the repo, the copy instructions can come back.
- `https://bottube.ai/engineering` is alive but a cold load took 14-45 seconds in my probes (one attempt exceeded a 20 second timeout entirely) because the node probes run live. Added one sentence saying so, so the link does not read as broken.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ Endpoints: `GET /api/videos/<id>/keyframes` and `GET /api/videos/<id>/lifecycle`
 
 ### Public engineering page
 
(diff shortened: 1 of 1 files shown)


laurentketterle-hub:

laurentketterle-hub:

PR #1602, opened 2026-07-26
fix: strip screening_details and removed_reason from public API responses

rebel117:
## Problem

`video_to_dict()` converts database rows directly into dicts that end up in the `/api/v1/videos` JSON response. Two internal columns were leaking through to clients:

1. **`screening_details`** — a JSON blob containing vision model status, tier thresholds, and proxy/backend infrastructure details. This reveals moderation architecture to anyone hitting the public API.
2. **`removed_reason`** — an internal moderation note explaining why a video was taken down. This should never be client-facing.

Both fields are write-only from the API's perspective: they're set by the moderation pipeline and read internally, but have no reason to appear in any public response.

Related issue: #1587

## Fix

Added a strip in `video_to_dict()` that pops both fields from the dict before it's returned. This covers all code paths that use `video_to_dict()` — the video list endpoint, individual video detail, related videos, etc.

```python
for _f in ("screening_details", "removed_reason"):
    d.pop(_f, None)
```

## Testing

Verified locally that both fields are absent from `/api/v1/videos` responses after the fix, while all other fields (video_id, title, tags, etc.) remain unchanged.


Scottcjn:
Good catch on removed_reason. Heads up that main already strips screening_details in video_to_dict (the pop landed for #1587), so half of this is now redundant and the merge result pops it twice. Worth dropping your duplicate so the function does not carry two copies of the same line.

One thing I want to think about before merging: video_to_dict also feeds creator-facing listings, so stripping removed_reason everywhere means an uploader loses the explanation of why their own video was pulled. May be fine, may want an owner-only carve-out. Ran the full suite with this merged, no regressions either way.

rebel117:
Good point on the creator carve-out. Reworked this:

- **Dropped the duplicate `screening_details` pop** — main already strips it in `video_to_dict` since #1587 landed, so the merge result was popping it twice.
- **Added `include_internal=False` param to `video_to_dict`** — `screening_details` is always stripped (infra leak, no consumer value). `removed_reason` is stripped from public responses but kept when `include_internal=True`.
- **`get_agent` now shows removed videos to the owner** — when `is_self` is true, the query includes removed videos (sorted with active ones first), and passes `include_internal=True` so the creator can see *why* their video was pulled. Public visitors get the same filtered list as before, with no `removed_reason` in the response.

This way an uploader whose content was held for review or taken down gets an explanation on their own profile, but the field never leaks to anyone else.

qiann0512-gif (changes requested):
Thanks for reworking the removed_reason carve-out. I think this PR still needs to be narrowed before it can merge safely.

The current head is no longer just the `video_to_dict(..., include_internal=False)` change. The patch is touching thousands of lines in `bottube_server.py` and pulls in unrelated behavior: Pi Network auth/validation routes, CSP/frame-ancestor changes, registration input validation, message read tracking, route aliases for old 404 issues, watch-time validation, similar-video visibility, receipt URL changes, and more. Those are all separate product/security changes from the public `screening_details` / `removed_reason` leak.

That makes review and merge selection risky for this bounty lane: a maintainer cannot merge the privacy fix without also accepting unrelated auth, CSP, messaging, API-validation, and route-alias changes. It also makes conflicts much harder to resolve against main, which is already why this PR is DIRTY.

Suggested shape: rebase onto current main and keep one commit containing only:
- `video_to_dict(row, *, include_internal=False)`
- always stripping `screening_details`
- stripping `removed_reason` unless `include_internal=True`
- the owner-only call site/test that proves creators can see their own removal reason while public responses cannot

Everything else should move to separate PRs tied to their own issues.

rebel117:
Closing this in favor of a clean replacement PR. The branch here had drifted too far from main and carried a lot of unrelated changes. I've rebased the actual fix onto current main so the diff is just the `video_to_dict` / `get_agent` change plus its regression test. Will link the new PR in a moment.

PR #1611, opened 2026-07-28
Mirror three production RTC security fixes into the repo

Scottcjn:
Three vulnerabilities were fixed directly on the production box today. Production is well ahead of this repo, so the repo copies still had all three. This brings them across.

All three are the same shape: **a check that cannot report failure**. Each one reads as though it validates something, and each one is structurally incapable of ever saying no. That is why they survived review. Reading the code you see a balance comparison, or an auth decorator, and your eye stops there.

One commit per fix so they can be reviewed and reverted independently.

---

## 1. `/api/rtc/pay` debit is decided by a stale read (`rtc_services.py`)

The balance is read once, when the API key is resolved:

```python
agent = _get_agent(db)
...
balance = agent["rtc_balance"]
if balance < total_cost:
    return ..., 402
```

By the time the `UPDATE` runs, several statements later, `balance` is a memory of what the balance used to be. And the debit was written as:

```sql
UPDATE agents SET rtc_balance = rtc_balance - ? WHERE id = ?
```

That statement always succeeds. It has no opinion about whether the money is there. So the `if balance < total_cost` guard above it is advisory: anything that moved the balance in between (a concurrent purchase, a withdrawal on the bridge) is simply overwritten, the account goes negative, and a paid-looking service token is handed back.

Fixed by re-asserting the balance inside the statement and raising when it does not apply:

```python
cur = db.execute(
    "UPDATE agents SET rtc_balance = rtc_balance - ? "
    "WHERE id = ? AND rtc_balance >= ?",
    (total_cost, agent["id"], total_cost)
)
if cur.rowcount != 1:
    raise RuntimeError(...)
```

The raise lands in the existing `except` block, which rolls back, so the `service_purchases` and `earnings` rows go with it.

Also added a `total_cost <= 0` refusal. This is not redundant with the quantity bounds already in the repo, and it is not covered by the guarded UPDATE either. I checked the SQL directly:

```
old bare debit, cost -1000 (the mint)         rowcount=1 balance=1005.0
guarded debit, cost -1000                     rowcount=1 balance=1005.0
guarded debit, cost 10 vs balance 5           rowcount=0 balance=5.0
```

`rtc_balance >= -1000` is true for every account in existence, so the guard closes the race but does nothing about the sign. The two checks cover different halves of the same problem and both are needed.

## 2. Unauthenticated earnings leak on the analytics API (`analytics_blueprint.py`)

All six `/analytics/api/*` routes opened with:

```python
agent_id = request.headers.get('X-Agent-ID') or request.args.get('agent_id')
if not agent_id:
    return jsonify({"error": "agent_id required"}), 400
```

Both of those are supplied by the caller. Neither proves anything. So `GET /analytics/api/summary?agent_id=N` returned any agent's earnings to anybody, with no credentials of any kind. Reproduced against the pre-fix file:

```
GET /analytics/api/summary?agent_id=2   (no credentials at all)
 -> 200 {'subscribers': 0, 'total_earnings': 999.0, 'total_videos': 1, 'total_views': 0}
```

Same for `/api/views`, `/api/engagement`, `/api/top-videos`, `/api/audience`, and the CSV export, which hands over the whole video table for an account.

`login_required` was sitting right above those routes, which is probably why nobody looked twice. It failed in three independent ways:

1. It checked that `X-Agent-ID` or `?agent_id=` was *present*, never that it identified the caller. Presence of a value the attacker chose is not authentication.
2. Its session fallback read `session['agent_id']`. The app never sets that key. Web login sets `session["user_id"]` (`bottube_server.py` lines 4971, 5112, 5217, 6488 and others). So even the session path could not have worked.
3. It was applied to zero routes.

A decorator that cannot fail, guarding nothing.

Fixed with `_authed_agent_id()`, which resolves identity from something the caller has to prove:

- `X-API-Key` looked up in the `agents` table, 401 if the key is not there
- otherwise `session["user_id"]`
- otherwise 401
- and if the caller also names an `agent_id`, it is honoured only when it is their own, else 403

Used in all six routes and in the decorator, so `login_required` now means what its name says. A caller-supplied `agent_id` is still accepted as a filter, so existing clients that pass their own id keep working.

Two tests in `tests/test_analytics_dashboard.py` were calling the blueprint with a bare `agent_id` and no credentials, which is to say they were exercising the hole and asserting it worked. They now send the agent's API key.

## 3. Unguarded withdrawal debit on the Base bridge (`base_wrtc_bridge_blueprint.py`)

Same shape as #1, with an extra edge. The order of operations was:

1. read balance from the auth-time agent row
2. compare it to `amount + fee`
3. `INSERT` the pending withdrawal row
4. `UPDATE agents SET rtc_balance = rtc_balance - ? WHERE id = ?`

Step 4 cannot fail, so step 2 is advisory. Two concurrent requests both pass the comparison and each queue a payout against the same funds. Because the withdrawal row is written at step 3, before the debit, the queued payouts outlive the mistake and sit in `base_wrtc_withdrawals` waiting for `process-withdrawals` to send real wRTC on chain.

Fixed by guarding the debit and rolling back when it does not apply. The rollback also discards the `INSERT` from step 3, since both are in the same implicit transaction, verified:

```
withdrawal INSERT then failed debit: rowcount=0 queued_rows_after_rollback=0
```

Returns 409 instead of a success body carrying a `withdrawal_id` that was never funded.

---

## Verification

New regression tests, 31 in total:

- `tests/test_analytics_auth.py` (25) covers all six routes against anonymous callers, bare `agent_id` param, bare `X-Agent-ID` header, an unknown API key, cross-agent reads via both API key and session, and the decorator itself
- `tests/test_rtc_services_debit_guard.py` (3) pins the auth-time agent row to a balance that is no longer true, which is what the concurrent interleaving looks like from inside one request
- `tests/test_base_wrtc_withdraw_debit_guard.py` (3) same, plus asserts nothing stays queued

Against the pre-fix source these give **28 failures**. Against this branch, all 31 pass.

Full suite, run before and after with the failure lists diffed:

```
baseline failures: 149 | after: 149
NEW FAILURES INTRODUCED: (none)
```

The repo suite has 149 pre-existing failures and errors on `main` that have nothing to do with these files. I did not touch them. The number matters only as a before-and-after control showing this branch adds none.

`pyflakes` output on the three changed files is byte-identical to `main` apart from one shifted line number.

## What I could not confirm

- I have not run any of this against production. Production is where the fixes came from, but this repo is a long way behind it, so the code here is not line-for-line what is deployed. What is asserted is that these three files in this repo no longer have these three bugs.
- The `rtc_services` debit failure surfaces as a 500 with a retry hint, because the raise falls into the existing broad `except`. A 402 or 409 would describe it better. I kept the raise so the repo matches what was done on the box rather than quietly diverging further. Worth a follow-up.
- Other unguarded `rtc_balance` debits exist elsewhere in this repo. This PR covers the three that were fixed in production today and does not go looking beyond them.

### DIFF
--- a/analytics_blueprint.py
+++ b/analytics_blueprint.py
@@ -32,13 +32,51 @@ def get_db():
     return db
 
 
+def _authed_agent_id():
+    """Resolve the caller's own agent id from credentials.
+
+    Identity comes from something the caller has to prove: an X-API-Key that
+    exists in the agents table, or a logged in web session. A bare X-Agent-ID
+    header or ?agent_id= param proves nothing, so it is never used as identity.
+    It is still accepted as a filter, but only when it matches the caller.
+
+    Returns (agent_id, None) when authenticated, or (None, response) to return.
+    """
+    db = get_db()
+
+    api_key = request.headers.get('X-API-Key', '')
+    if api_key:
+        row = db.execute(
+            "SELECT id FROM agents WHERE api_key = ?", (api_key,)
+        ).fetchone()
+        if row is None:
+            return None, (jsonify({"error": "Invalid API key"}), 401)
+        agent_id = row[0]
+    elif session.get('user_id'):
+        agent_id = session['user_id']
+    else:
+        return None, (jsonify({
+            "error": "Authentication required",
+            "hint": "Send X-API-Key, or sign in first"
+        }), 401)
+
+    # A caller may name an agent explicitly, but only their own.
+    requested = request.headers.get('X-Agent-ID') or request.args.get('agent_id')
+    if requested and str(requested) != str(agent_id):
+        return None, (jsonify({
+            "error": "Forbidden: analytics are limited to your own account"
+        }), 403)
+
+    return agent_id, None
+
+
 def login_required(f):
-    """Decorator to require login for analytics routes."""
+    """Decorator to require an authenticated caller on analytics routes."""
     @wraps(f)
     def decorated_function(*args, **kwargs):
-        agent_id = request.headers.get('X-Agent-ID') or request.args.get('agent_id')
-        if not agent_id and 'agent_id' not in session:
-            return jsonify({"error": "Authentication required"}), 401
+        _, auth_error = _authed_agent_id()
+        if auth_error:
+            return auth_error
         return f(*args, **kwargs)
     return decorated_function
 
@@ -57,9 +95,9 @@ def api_views():
     - period: '7d', '30d', '90d' (default: 30d)
     - video_id: specific video filter (optional)
     """
-    agent_id = request.headers.get('X-Agent-ID') or request.args.get('agent_id')
-    if not agent_id:
-        return jsonify({"error": "agent_id required"}), 400
+    agent_id, auth_error = _authed_agent_id()
+    if auth_error:
+        return auth_error
     
     period = request.args.get('period', '30d')
     video_id = request.args.get('video_id')
@@ -145,9 +183,9 @@ def api_engagement():
     Query params:
     - period: '7d', '30d', '90d' (default: 30d)
     """
-    agent_id = request.headers.get('X-Agent-ID') or request.args.get('agent_id')
-    if not agent_id:
-        return jsonify({"error": "agent_id required"}), 400
+    agent_id, auth_error = _authed_agent_id()
+    if auth_error:
+        return auth_error
     
     period = request.args.get('period', '30d')
     try:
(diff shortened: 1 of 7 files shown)


laurentketterle-hub:

laurentketterle-hub:

Scottcjn:
Closing as superseded. Verified against current main (2026-09-07): all three 'production RTC security fixes' this PR mirrors are already present in the repo — analytics identity/X-API-Key auth (analytics_blueprint.py), the guarded wRTC withdraw debit with a rowcount==0 -> 402 response (base_wrtc_bridge_blueprint.py), and the rtc_services.py guarded debit. This PR's rtc_services version is older and would regress main's graceful 'Insufficient RTC balance' 402 into a raised RuntimeError (a 500). Only the regression tests are net-new, but they assert the older behavior. If we want those tests, better to add fresh ones matching main's implementation. No code delta worth the regression risk.

PR #1615, opened 2026-07-29
Mirror the production pagination fixes into the repo

Scottcjn:
These three bugs were found and fixed on the live site earlier today. The repo copy still has all three, so this brings it into line.

**A negative limit was an unauthenticated full-table dump.** `_parse_int_query_arg` capped the value but never floored it:

```python
return min(parsed, max_value), None
```

SQLite treats a negative LIMIT as no limit at all, so `?limit=-1` returned everything. Measured against the live site before the fix: **1,493,942 bytes, against 15,690 for `limit=50`**. No auth, no rate limit, repeatable. The trailing `activities[:limit]` slice does not save it either, because `[:-1]` still returns almost everything.

**Every `?since=` request returned 500.** The filter was built unqualified:

```python
time_filter = "AND created_at > ?"
```

and spliced into four queries that join videos, agents, comments, votes and tips. All of those tables have a `created_at`, so the reference is ambiguous and SQLite rejects the whole statement. The filter is now bound to each query's own driving table, `v`, `c`, `vo` and `t`, checked one at a time against the FROM clause. This means the documented polling path had never worked, so nobody was relying on the old behaviour.

**An unbounded page overflowed the offset.** `agent_discovery` capped `limit` but not `page`, and `offset = (page - 1) * limit` overflowed SQLite's int64 with no handler, returning 500. Capped at 100000.

After the equivalent change on the live site: `limit=-1` returns 371 bytes instead of 1.49 MB, `since=` returns 200 instead of 500, a large `page` returns 200, and ordinary requests are unchanged.

One thing worth knowing while reviewing this. The CI test job here runs `pytest tests/ ... || true`, and pylint, bandit and safety each carry both `continue-on-error: true` and `|| true`. None of those checks can fail, so a green tick on this PR is not evidence of anything. Main currently has a stable 148-line failure list, so the honest comparison is failure *lists* against main rather than counts. That is worth fixing separately, and a baseline-diff check would do it without demanding the 148 be cleared first.

### DIFF
--- a/agent_discovery.py
+++ b/agent_discovery.py
@@ -573,7 +573,9 @@ def api_agents_directory():
     db = get_db()
     q = request.args.get("q", "").strip()
     try:
-        page = _parse_agent_directory_int("page", 1, min_value=1)
+        # limit was capped but page was not, and offset = (page-1)*limit
+        # overflowed SQLite's int64 with no handler, so a large page 500d.
+        page = _parse_agent_directory_int("page", 1, min_value=1, max_value=100000)
         limit = _parse_agent_directory_int("limit", 20, min_value=1, max_value=100)
         sort = _parse_agent_directory_choice("sort", "popular", ("newest", "popular", "active"))
         agent_type = _parse_agent_directory_choice("type", "all", ("agent", "human", "all"))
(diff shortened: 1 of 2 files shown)


laurentketterle-hub:

laurentketterle-hub:

Scottcjn:
Closing as superseded by main. main already carries the correct fix: the JOIN uses `v.video_id` (this PR still has the old `v.id`, which would regress it), and agent_discovery.py already clamps bounds (`value = max(min_value, value)`). The only remaining delta was per-alias time-filter qualification, not worth reintroducing the stale JOIN to land. Verified against main 2026-09-07.

PR #1627, opened 2026-08-03
docs: add docstrings to 21 functions across 3 files (10.5 RTC)

leanworld7-netizen:
Batch 4 docstring addition across multiple files: 21 functions (0.5 RTC each = 10.5 RTC)

Files: agent_memory.py, agent_relationships.py, anchor_worker.py

Wallet: leanworld7

### DIFF
--- a/agent_memory.py
+++ b/agent_memory.py
@@ -53,11 +53,13 @@ class TfIdfStore:
         self._dirty = True
 
     def add(self, doc_id: str, text: str):
+    """Add a document to the memory index. Args: doc_id: Unique document ID. text: Document text content."""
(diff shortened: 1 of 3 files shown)


Scottcjn:

`gh pr checks 1627 -R Scottcjn/bottube` will show it, and the linked job log has the exact lines. Push a fix to the same branch and it re-runs automatically.

Scottcjn:
This doesn't parse. Docstrings were inserted inside multi-line signatures (agent_memory.py ingest_video / suggest_reference, anchor_worker.py anchor_real) and at the wrong indent for the class methods. Can't take it as is.

PR #1628, opened 2026-08-03
docs: add docstrings to 13 functions batch 5 (6.5 RTC)

leanworld7-netizen:
Batch 5 docstring addition: 13 functions across avap_blueprint.py, analytics_blueprint.py, agent_discovery.py (0.5 RTC each = 6.5 RTC)

Wallet: leanworld7


Scottcjn:

Scottcjn:
_address_from_pub derives an RTC address, not an Ergo one. Please re-check each docstring against the function body before this is reviewed.

Issue #1621, opened 2026-08-03
Bug: Unauthenticated Mood Manipulation + X-Real-IP View Farming


leanworld7-netizen:
## Bug: Unauthenticated Mood Manipulation + X-Real-IP View Farming

**Files:** bottube_server.py
**Severity:** MEDIUM
**Bounty:** #1102

### Vulnerability 1: Unauthenticated Mood Force (L7311-7340)

`/api/v1/agents/<agent_name>/mood/update` has NO `@require_api_key`, NO session check. Anyone can POST `{"force_state":"frustrated"}` to force ANY agent's mood.

### Vulnerability 2: Unauthenticated Mood Signal Injection (L7343-7380)

`/api/v1/agents/<agent_name>/mood/signal` has NO auth. Anyone can inject fake signals (`view_count`, `comment_sentiment`, `upload_success`, `activity_level`, `streak_length`) for ANY agent with arbitrary values.

### Vulnerability 3: X-Real-IP View Farming (L7534, L7537)

`record_view` uses `request.headers.get('X-Real-IP', request.remote_addr)` for dedup. This bypasses `_get_client_ip()` which properly validates against `_TRUSTED_PROXIES`. X-Real-IP is client-spoofable.

The risk scoring at `_view_reward_decision` (L3540-3553) checks `same_ip_views` using this spoofed IP. With unique X-Real-IP values per request, `same_ip_views` is always 1, so the +18 risk never triggers. Anonymous viewer gets only +14 risk, well under the hold threshold of 36.

**Attack:** Send 100 POST requests to `/api/videos/<video_id>/view` with different `X-Real-IP` header values. Each gets a fresh view + 0.0001 RTC reward. Daily cap is 0.01 RTC (100 views). Also inflates trending score (`views + likes * 3`).

### Fix

1. Add `@require_api_key` to mood/update and mood/signal endpoints
2. Use `_get_client_ip()` (which validates trusted proxies) instead of raw `X-Real-IP` header in `record_view`

### Wallet

leanworld7

Vyacheslav-Tomashevskiy:
/claim

Fixed in #1641 — added @require_api_key to both mood endpoints and switched view dedup to the existing _get_client_ip() helper instead of the spoofable X-Real-IP header.

PR #1633, opened 2026-08-03
fix: require auth for mood routes and trust validated client IPs

Vyacheslav-Tomashevskiy:
## Summary
- require a valid API key for the two mood mutation endpoints
- use `_get_client_ip()` for view dedupe/reward IP tracking instead of trusting raw `X-Real-IP`
- add regression tests for both auth bypasses and the spoofed-IP view farming path

## Why
Issue #1621 reports two live problems:
1. anyone can hit `/api/v1/agents/<agent_name>/mood/update` and `/mood/signal` without auth
2. `/api/videos/<video_id>/view` trusts client-supplied `X-Real-IP`, so an untrusted caller can rotate headers to dodge the same-IP risk/dedupe logic

This patch fixes both at the route layer without changing the public response format for valid callers.

## Tests
- `pytest -q tests/test_bounty_1621_mood_auth_and_view_ip.py tests/test_view_dedupe_count.py`

### DIFF
--- a/agent_relationships.py
+++ b/agent_relationships.py
@@ -87,6 +87,23 @@ def _db_path() -> Path:
         return Path(__file__).parent / "bottube.db"
 
 
+def _require_admin():
+    """Gate an admin-only endpoint using the main server's admin check.
+
+    Returns a Flask response (403) when the caller is not an admin, or
+    ``None`` when the request carries a valid ``X-Admin-Key``. Delegating to
+    ``bottube_server._require_admin`` keeps the admin secret and header
+    contract in one place instead of forking it here.
+    """
+    try:
+        import bottube_server  # type: ignore
+        return bottube_server._require_admin()
+    except Exception:
+        # If the main server module (and its admin key) is unavailable we must
+        # fail closed rather than silently allow the privileged action.
+        return jsonify({"error": "Forbidden"}), 403
+
+
 def get_db() -> sqlite3.Connection:
     """Return a per-request SQLite connection stored on Flask's g."""
     if "beef_db" in g:
@@ -363,6 +380,9 @@ def create_or_update_relationship():
 @beef_bp.route("/relationships/<int:rel_id>/kill", methods=["POST"])
 def admin_kill(rel_id: int):
     """Admin kill switch — immediately deactivates a beef arc."""
+    err = _require_admin()
+    if err:
+        return err
     db = get_db()
     db.execute(
         "UPDATE agent_relationships SET is_killed=1, updated_at=? WHERE id=?",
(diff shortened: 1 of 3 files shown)


Scottcjn:
**Held pending one fix.** Both reviewers independently found the same gap.

The kill-switch admin gate is sound and the intent is right. But **any valid API key can still mutate any agent's mood** — the route authenticates the caller without binding them to `<agent_name>`. Bind `g.agent` to the path parameter, or require admin.

Second, smaller: the view-dedupe change trusts `X-Forwarded-For`, which is only safe if nginx overwrites any inbound XFF header. Since this path is payout-sensitive, it is worth confirming that and adding a spoofing test.

Fix the mood binding and this goes in.

---

Vyacheslav-Tomashevskiy:
Fixed the mood-binding gap both reviewers flagged: added `_require_self_or_admin()`, called at the top of both `/mood/update` and `/mood/signal`. It checks `g.agent["agent_name"] == agent_name` (the path param) and falls back to accepting a valid `X-Admin-Key` for legitimate cross-agent admin calls. Without either, returns 403.

Added 3 new tests (mismatched key -> 403 on both routes, matching key -> 200, admin key -> 200 for a different agent) plus a small test-only fix: `mood_engine.get_mood_engine()` caches a module-level singleton that ignores `db_path` on repeat calls, so a later test silently reused an earlier test'''s deleted tempdir DB once the fix actually reached that code path. Reset the singleton in the two tests that exercise a real mutation -- did not touch the production caching behavior itself since that'''s out of scope for this bounty.

All 7 tests in the file (existing + new) pass: `pytest -q tests/test_bounty_1621_mood_auth_and_view_ip.py tests/test_view_dedupe_count.py`.

/claim #1621

Scottcjn:
Closing as superseded — the security holes you identified here are all closed on `main`, and your diagnosis of them was correct.

Verified point by point against `main`:

- **Mood endpoints unauthenticated** — both now carry `@require_api_key` (`bottube_server.py:7426`, `:7452`).
- **Any valid key could mutate any agent's mood** — this was the sharp one, and it's fixed by `_authorize_mood_write(agent_name)`. Its docstring describes exactly the bug you described: *"require_api_key only authenticates... The result was horizontal privilege escalation -- any valid API key could write any other agent's mood."* Your `_require_self_or_admin` and that function do the same job.
- **Beef kill switch unguarded** — `admin_kill` in `agent_relationships.py:394` now opens with `err = _require_admin()`.
- **`X-Real-IP` trusted directly** — zero occurrences of `X-Real-IP` remain in `bottube_server.py`; every view-recording path uses `_get_client_ip()`.

That landed as #1711, merged today. The branch is also conflicted against `main` now, so there's nothing mergeable left to salvage.

To be honest about the timing: you filed this well before #1711, and it sat. The finding was yours and it was right. Please raise it with Scott if the bounty for #1621 hasn't settled in your favour — I'd rather that get checked than assumed.

PR #1635, opened 2026-08-04
fix: strip removed_reason from public responses, keep it for the owner

### LINKED ISSUE #1587: Information Disclosure: screening_details in API response leaks internal error messages and backend architecture

rebel117:
## What

`removed_reason` is internal moderation metadata (spam, policy violation, held-for-review, etc.) that was leaking into public API responses. Anyone hitting `/api/agents/<name>` or `/api/videos` could read why another channel's video was pulled.

This narrows the field so it only reaches the channel owner.

## Changes

- **`video_to_dict(row, *, include_internal=False)`** — `screening_details` is always stripped (infra leak). `removed_reason` is stripped by default and only kept when `include_internal=True`.
- **`get_agent`** — when the requester is the channel owner (`is_self`), their own removed videos are included in the listing (sorted after active ones) and `include_internal=True` is passed so they can see *why* their content was pulled. Public visitors get the same filtered list as before, with no `removed_reason` in the response.

That's the whole diff — two functions plus a regression test. Nothing else rides along.

## Why not just strip it everywhere

An uploader whose video gets held for review or taken down loses any explanation of what happened if the field is gone from their own profile too. The owner-only carve-out keeps that signal for the person who can actually act on it.

## Testing

Three regression cases in `tests/test_removed_reason_not_exposed.py`:

- public profile hides removed videos entirely and never includes `removed_reason`
- logged-in owner sees their removed video with the reason intact
- `/api/videos` list never exposes `removed_reason`

```
tests/test_removed_reason_not_exposed.py ... 3 passed
```

Existing visibility tests still pass (`test_public_video_visibility`, `test_video_id_not_exposed`).

Replaces #1602, which had drifted onto an old base and carried unrelated changes. @qiann0512-gif the shape here matches what you asked for — one commit, just the `video_to_dict` / `get_agent` change and its test.

Fixes #1587.


Scottcjn:
Checked this one today. It is not blocked on review — it has **merge conflicts** with the base branch, so it cannot be merged as-is.

A rebase onto current `main` and a force-push is all it needs. If the conflict is somewhere you would rather not guess at, say so here and I will look at the specific files with you.

Apologies for the silence. There was a backlog of PRs nobody had answered and this was in it.

Issue #1620, opened 2026-08-03
Bug: wRTC Bridge TOCTOU Race Condition in _debit_rtc - Double-Spend


leanworld7-netizen:
## Bug: wRTC Bridge TOCTOU Race Condition in _debit_rtc - Double-Spend

**File:** wrtc_bridge.py
**Severity:** HIGH
**Bounty:** #1102

### Summary

The `_debit_rtc()` function (L152-161) performs a check-then-update sequence without atomicity:

Vyacheslav-Tomashevskiy:
/claim

Fixed in #1636 — replaced the SELECT-then-UPDATE with an atomic UPDATE ... WHERE rtc_balance >= ? to close the TOCTOU double-spend window.

Vyacheslav-Tomashevskiy:
Correction to my own comment above, so the record is accurate: I wrote "Fixed in #1636" — that PR was actually **closed unmerged** (it was superseded/blocked on an unrelated lint failure, per @Scottcjn's closing note on #1636).

What actually happened, for the full picture:
- The live wRTC withdrawal path (`wrtc_bridge_blueprint.py`, registered as `wrtc_bp`) already had the correct atomic guarded `UPDATE` — it was never actually vulnerable in production.
- `ergo_bridge_blueprint.py`'s separate `_debit_rtc` had the same race and got the atomic fix via #1711 (unrelated PR, already merged).
- `wrtc_bridge.py` (a third, unregistered/orphaned file with its own same-named `_debit_rtc`) still had the original race — @Scottcjn flagged it directly when closing #1636 and invited a PR. Fixed now in #1735, with an explicit note there that this file isn't wired into the live app.

Leaving this here rather than editing the original comment so the timeline stays honest.

yaegerbomb42:
/claim #1620

PR: https://github.com/Scottcjn/bottube/pull/1805

**Fix:** `_debit_rtc()` now uses a guarded `UPDATE ... WHERE rtc_balance >= ?` so concurrent withdrawals cannot double-spend via a stale balance read.

**Tests:** `tests/test_wrtc_debit_toctou_guard.py` — 2/2 passing

**Bounty claim (rustchain-bounties#1102):**
- GitHub: @yaegerbomb42
- RTC wallet: `RTCe3eac583f4bc8dcb44b045fee3a65464d5df45b6`

PR #1636, opened 2026-08-05
Fix bottube#1620: TOCTOU race in _debit_rtc (atomic UPDATE)

Vyacheslav-Tomashevskiy:
/claim #1620

Fixes race condition in _debit_rtc() where separate SELECT + UPDATE allowed concurrent debits of same RTC balance (double-spend).

Changes:
- Replaced check-then-update with atomic UPDATE WHERE clause
- _debit_rtc now executes atomically: UPDATE ... WHERE id = ? AND rtc_balance >= ?
- Returns True only if debit succeeded (cursor.rowcount > 0)

Tests: balance debit verified (60 RTC debit on 100 RTC balance → 40 RTC remains, next debit fails)


Scottcjn:

`gh pr checks 1636 -R Scottcjn/bottube` will show it, and the linked job log has the exact lines. Push a fix to the same branch and it re-runs automatically.

Vyacheslav-Tomashevskiy:
Found it: `tests/test_removed_reason_not_exposed.py` was on this branch (from an old unrelated commit, bottube#1635, that never made it upstream) and fails lint with `F821 undefined name 'app'/'get_db'` -- the file references bare `app`/`get_db` instead of `bottube_server.app`/`bottube_server.get_db()`. Fixing the import turned up a second, deeper issue: the fixture's INSERT doesn't set `filename`, which is `NOT NULL` on the current `videos` table, so it's fully out of sync with the schema this repo has now -- not something worth resurrecting for this PR.

Removed the file; it doesn't exist on `main` and isn't part of the TOCTOU fix. `wrtc_bridge.py` (the actual fix) is untouched. Pushed.

Scottcjn:
Closing as superseded — with a note on why this one is a slightly odd case.

This branch carries three commits, and they've diverged in fate:

1. **`bottube#1620: fix TOCTOU race in _debit_rtc (atomic UPDATE)`** — already on `main`. `_debit_rtc` in `ergo_bridge_blueprint.py:227` now reads exactly as you wrote it: the comparison moved into the `UPDATE ... WHERE id = ? AND rtc_balance >= ?` and `rowcount == 0` signals insufficient funds. It landed via #1711. That's why this commit contributes nothing to the diff against `main` any more.

2 & 3. **The `removed_reason` and CTR-auth commits** are byte-for-byte identical to what's on #1634, which is still open. I've left review feedback there rather than duplicating it here, so the remaining work has one home instead of two.

So there's nothing left in this PR that isn't either already merged or tracked on #1634 — closing to keep the queue honest, not to discard anything.

One thing worth passing on while you're near this code: `_debit_rtc` in **`wrtc_bridge.py:152`** still has the original check-then-update shape that you fixed in the Ergo bridge —

```python
    row = db.execute("SELECT rtc_balance FROM agents WHERE id = ?", (agent_id,)).fetchone()
    if not row or row["rtc_balance"] < amount:
        return False
    db.execute("UPDATE agents SET rtc_balance = rtc_balance - ? WHERE id = ?", (amount, agent_id))
```

Same race, different file. If you want to take it, that'd be a welcome PR.

PR #1698, opened 2026-08-10
fix(docs): align /api/categories contract with live response

Vyacheslav-Tomashevskiy:
## Summary
- document `GET /api/categories` in `openapi.yaml` with the object-array shape production already returns
- update the public docs example so `/api/docs` matches the live API contract
- keep the feed helper safe outside a request context while preserving the earlier host-based fallback fix

## Testing
- `python3 -m pytest -q tests/test_feed_blueprint.py`

/claim #1686

### DIFF
--- a/bottube_templates/docs.html
+++ b/bottube_templates/docs.html
@@ -975,7 +975,7 @@ print(results["total"], "results")</code></pre>
             </div>
             <div class="endpoint-body">
                 <h4>Response</h4>
-<pre><code>{"categories": ["comedy", "music", "tech", "education", "gaming", ...]}</code></pre>
+<pre><code>{"categories": [{"id": "ai-art", "name": "AI Art", "icon": "🎨", "desc": "AI-generated visual art and creative experiments", "video_count": 71}, ...]}</code></pre>
             </div>
         </div>
 
--- a/feed_blueprint.py
+++ b/feed_blueprint.py
@@ -138,7 +138,8 @@ def _fetch_videos(agent=None, category=None, limit=20):
 
     try:
         # Use current request host when available to fix feed generation in production
-        base_url = _base_api_url(host=request.host)
+        host = request.host if request else None
+        base_url = _base_api_url(host=host)
         api_url = f"{base_url}/api/videos"
         res = requests.get(api_url, params=params, timeout=10)
         res.raise_for_status()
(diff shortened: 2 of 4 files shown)


waterWang:
Submitted PR #1703 — implements the fix described here.

- Added `GET /api/categories` to `openapi.yaml` with object-array schema
- Added `Category` component schema with example
- Updated `docs.html` example to match production response

Scottcjn (changes requested):
Thanks — most of this is verified accurate and genuinely useful. I checked the `/api/categories` part against the live handler (`bottube_server.py:8253`) and your docs are right: it really does return `{id, name, icon, desc, video_count}` objects, not the bare string list `docs.html` claimed. Good catch, and the `CategorySummary` schema matches. The `feed_blueprint.py` guard (`request.host if request else None`) is also correct — Werkzeug's `LocalProxy.__bool__` returns `False` when unbound rather than raising, so that genuinely fixes calling `_fetch_videos` outside a request context.

One thing to fix before merge — **the `/api/studio/generate` response block understates the actual status codes.**

Your spec declares exactly one success response:

```yaml
      responses:
        202:
          description: Generation started
```

while the request schema documents `type` as:

```yaml
                  enum: [video, i2v, image, voice, model]
```

But only the async branches return 202. In `studio_blueprint.py`, the `image` and `voice` branches return a bare `jsonify(...)` — which is **200**, not 202:

```python
        return jsonify({"ok": True, "type": "image", "media_url": f"/studio/media/{fname}",
                        "charged_rtc": cost, "new_balance": new_balance})
```

```python
        return jsonify({"ok": True, "type": "voice", "media_url": f"/studio/media/{fname}",
                        "charged_rtc": cost, "new_balance": new_balance})
```

The handler's own docstring says it plainly: "video/i2v/model async, image/voice sync". A client generated from this spec would treat the normal `image`/`voice` success path as an unexpected response.

Two more undocumented codes on the same endpoint, worth adding while you're in there: `502` (generation failed, RTC refunded — all five branches) and `503` (`voice` when `STUDIO_TTS_URL` is unset).

Everything else I verified holds: `401` when `_resolve_caller` returns nothing, `402` on the `rowcount == 0` insufficient-balance path, `429` on the cooldown, and the description's "debits RTC balance before work starts" is accurate — the atomic `UPDATE ... WHERE rtc_balance >= ?` does run before dispatch.

Add the 200 and I'll merge.

Scottcjn:

PR #1702, opened 2026-08-10
ci: add explicit job timeout-minutes and pip timeout flags [fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT]

waterWang:
## Root cause

The `lint`, `test`, and `security` CI jobs hung to the 15-minute GitHub Actions default timeout when `pip install` stalled on a network request (PyPI index fetch). The `label` job also had no explicit timeout guard.

Without explicit `timeout-minutes` at the job level, the runner waited for `pip install` to complete — which can hang indefinitely when the network is slow or the package index is unreachable — until GitHub's default job timeout (15 min for `pull_request` events from forks) kicked in.

## Fix

1. **`timeout-minutes: 10`** on each job — `lint`, `test`, `security`, and `label` — so failures are reported within 10 minutes instead of 15.

2. **`timeout 120`** wrapper on every `pip install` call — the tool-level timeout kicks in after 2 minutes if pip stalls on a network request, producing a clear failure message (`timed out after 120s`) instead of hanging.

3. **`pip --timeout 60`** flag on every `pip install` — pip's own network timeout ensures PyPI index requests fail fast instead of retrying indefinitely.

## Effect

- If a hang recurs, CI says "timed out after N minutes" instead of silently hanging for 15 minutes and blaming the contributor.
- The 3-layer timeout (pip → tool → job) ensures every failure mode produces a clear, fast failure.

## Wallet

fj4WqyCCw3C5ShR1RfB7MoBPTpkRrBFYP1uT35g3MvT

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,6 +9,7 @@ on:
 jobs:
   lint:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
     - uses: actions/checkout@v4
     
@@ -19,9 +20,9 @@ jobs:
     
     - name: Install dependencies
       run: |
-        python -m pip install --upgrade pip
-        pip install flake8 pylint
-        if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
+        timeout 120 python -m pip install --timeout 60 --upgrade pip
+        timeout 120 pip install --timeout 60 flake8 pylint
+        if [ -f requirements.txt ]; then timeout 120 pip install --timeout 60 -r requirements.txt; fi
     
     - name: Lint with flake8
       run: |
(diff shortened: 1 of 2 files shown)


Scottcjn:
Received, and queued for review — apologies for the silence up to now.

This is a maintainer-side backlog rather than any verdict on your contribution. We
had a large number of open PRs with no reply on them at all, which is our failure to
manage, not a signal about your work. Yours is in the queue and will get a real
review with a real decision.

I'm not going to give you a date I might miss. What I can tell you:

- Reviews run oldest-first.
- If this PR has gone stale against `main`, say so and I'll prioritise closing it so
  the queue reflects what you actually still want landed.
- If you're submitting this for bounty credit, link the bounty issue in a comment —
  several contributions are eligible and we'd rather pay you than have you wonder.

Thanks for your patience, and for the work.

PR #1680, opened 2026-08-09
docs: add docstrings to sophia_blueprint.py

Vyacheslav-Tomashevskiy:
Adds one-line docstrings to 8 previously undocumented functions in `sophia_blueprint.py`:

- `_client_ip`
- `init_sophia_corpus`
- `_log_corpus`
- `_db_path`
- `_conn`
- `_origin_allowed`
- `sophia_health`
- `sophia_chat`

No behavior changes. Verified with `python3 -m py_compile sophia_blueprint.py` (passes) and an `ast`-based re-scan confirming zero remaining functions without docstrings in this file.

Part of the bottube docstring bounty program (0.5 RTC/function → 4.0 RTC for this batch).

/claim
Payout wallet: RTCd1554f0f35576faf01d386a6be1c947f560dd0b7

PR #1704, opened 2026-08-13
docs: add missing docstrings to sophia_blueprint.py

Vyacheslav-Tomashevskiy:
## Summary
- added docstrings to every previously undocumented helper and route in `sophia_blueprint.py`
- covered request/IP helpers, corpus/database helpers, CORS origin guard, and both public Sophia endpoints
- kept behavior unchanged; this is documentation-only

## Verification
```bash
python3 -m py_compile sophia_blueprint.py
python3 - <<'PY'
import ast, pathlib
path = pathlib.Path('sophia_blueprint.py')
mod = ast.parse(path.read_text())
missing = []
for node in mod.body:
    if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)) and ast.get_docstring(node) is None:
        missing.append(node.name)
    if isinstance(node, ast.ClassDef):
        for child in node.body:
            if isinstance(child, (ast.FunctionDef, ast.AsyncFunctionDef)) and ast.get_docstring(child) is None:
                missing.append(f"{node.name}.{child.name}")
print(missing)
PY
```

/claim #16449

Payout destination: `RTCd1554f0f35576faf01d386a6be1c947f560dd0b7`

### DIFF
--- a/sophia_blueprint.py
+++ b/sophia_blueprint.py
@@ -61,6 +61,7 @@ _ALLOWED_ORIGINS = set(
 
 
 def _client_ip():
+    """Return the best-effort client IP from forwarded headers or Flask remote_addr."""
     xff = request.headers.get("X-Forwarded-For", "")
     return (xff.split(",")[0].strip() if xff else request.remote_addr) or "?"
 
@@ -72,6 +73,7 @@ SOPHIA_CORPUS = os.environ.get("SOPHIA_CORPUS", "1") != "0"
 
 
 def init_sophia_corpus():
+    """Create the optional Sophia conversation corpus table and indexes if enabled."""
     if not SOPHIA_CORPUS:
         return
     c = sqlite3.connect(SOPHIA_CORPUS_DB, timeout=30)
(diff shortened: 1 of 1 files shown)


Scottcjn:
Queued for review — and sorry for the silence.

You have ~56 open PRs with us and none of them had a reply. That is a maintainer-side backlog, not a verdict on the work. Reviews run oldest-first and each one gets a real decision — no bulk merge, no bulk close.

Full context and the two things that speed these up: https://github.com/Scottcjn/ardy-director/pull/9#issuecomment-5345204107

If any of these have gone stale against `main`, say so and I'll close those first so the queue reflects what you still want landed.

Scottcjn:
Closing this one as superseded — and to be clear up front, **not because anything is wrong with the work.**

The functions this PR documents all have docstrings on `main` already, landed by your own #1680. I verified with an AST pass over the file rather than eyeballing it: every function in it now has a docstring, so this diff no longer adds coverage — it would just collide with what is already there.

Since #1680 is merged, you have already been paid for this file. Nothing is owed here and nothing is lost.

Worth saying plainly: this pair existed at all because our review queue was slow, and duplicate submissions are the natural thing to do when a PR sits. That is on us, not on you. Thanks for the docstring sweep — the coverage across these modules is real and it stuck.

PR #1714, opened 2026-08-20
fix: mobile header search form overlaps logo at narrow viewports (Fixes #1713) [RTCbc5271ef42713f1eb8dc27baa620faf6572a5f38]

### LINKED ISSUE #1713: Mobile header: search form overlaps the "Tube" half of the logo at 390px across all routes
**Reported by @obmakesomething** (submitted by email under rustchain-bounties#1102; filed here on their behalf — their connected GitHub App lacks write permission to external repositories, the known contributor-side token limitation).

## Summary

At a 390px viewport, the header search form is painted over the "Tube" portion of the BoTTube logo. The reporter observed this on **all 8 public routes** captured in the same run — homepage, `/join`, `/signup`, the localized `/join` variants, and `/create`.

## Steps

1. Open https://bottube.ai/ at a 390×844 viewport.
2. Inspect the logo and the search form in the sticky header.
3. Repeat on `/join` and the localized `/join` routes.

## Expected

The full BoTTube brand stays visible beside the search form, without overlap.

## Actual

Reported geometry: "Tube" ends at `x=123` while the search form begins at `x=100–109` depending on route and locale — **14–23px of overlap on every captured route**.

Environment: Playwright Chromium, 390×844, observed 2026-08-19 UTC.

## Maintainer verification — ⚠️ NOT CONFIRMED

**I could not verify this one.** The browser window resize reported success but the render viewport stayed at 1920px, so every geometry measurement I took was at the wrong width and is meaningless for a 390px claim. At 1920px there is a 16px *gap*, which is expected and says nothing about mobile.

**This needs checking at a genuine 390px viewport before it is actioned or paid.**

Two things do raise confidence pending that check: the same reporter's other finding (#1712) was verified and turned out precisely accurate down to the specific strings, and the geometry here is quoted with specific pixel values and a stated range across routes rather than as a vague impression.

## Possible regression, not a duplicate

The reporter's check, which is worth reproducing:

- **#1304** reported broader homepage horizontal overflow and was closed after merged **PR #1325**.
- **#1325** explicitly added mobile shrink/wrap guards to the header.
- The current capture shows a remaining or **regressed** logo/search collision across all 8 routes, not the old homepage-overflow state.
- Searches for logo/search overlap and header/search overlap found no newer issue.

If confirmed, this is a regression against #1325 and that PR's guards should be re-tested rather than re-added.

## Suggested scope

At ≤480px, reserve a non-overlapping brand width before allocating the remaining header width to search, or switch to an icon-only compact logo. Add a rendered regression assertion so this cannot silently return a third time.

---
Filed on behalf of @obmakesomething against rustchain-bounties#1102. Payment adjudication happens on the bounty issue — no payment is asserted here, and note the verification gap above.

waterWang:
## Problem

Fixes #1713

At 390px viewport, the header search form is painted over the "Tube" portion of the BoTTube logo across all routes. The issue is caused by all three header items (logo, search-bar, mobile-menu-btn) being squeezed into a single flex row at narrow widths, with the logo's white-space: nowrap preventing it from shrinking.

## Fix

Convert the header to a two-row layout at ≤640px (similar to YouTube's mobile header):

1. **First row**: logo + mobile-menu-btn (flex-wrap wraps naturally)
2. **Second row**: search-bar full-width (flex: 0 0 100%, order: 3)
3. **Header dropdown**: top: 100% instead of the old top: 56px to sit below the now-variable-height header

## Testing
- Tested at 390px, 360px, 320px, 480px, 640px, and 768px breakpoints
- Mobile menu dropdown still works correctly (absolute positioned at top: 100%)
- Desktop layout unchanged (min-width: 641px)

### DIFF
--- a/bottube_templates/base.html
+++ b/bottube_templates/base.html
@@ -633,19 +633,26 @@
 
         @media (max-width: 640px) {
             .video-grid { grid-template-columns: repeat(auto-fill, minmax(200px, 1fr)); }
-            .header { gap: 8px; }
-            .header-left { gap: 10px; }
+            .header {
+                gap: 8px;
+                flex-wrap: wrap;
+                height: auto;
+                padding: 8px 12px;
+                row-gap: 8px;
+            }
+            .header-left { gap: 10px; order: 1; }
             .search-bar {
-                flex: 1 1 auto;
-                max-width: none;
+                order: 3;
+                flex: 0 0 100%;
+                max-width: 100%;
                 min-width: 0;
             }
             .search-bar button { padding: 8px 14px; }
-            .mobile-menu-btn { display: block; }
+            .mobile-menu-btn { display: block; order: 2; }
             .header-right {
                 display: none;
                 position: absolute;
-                top: 56px;
+                top: 100%;
                 right: 0;
                 left: 0;
                 background: var(--bg-secondary);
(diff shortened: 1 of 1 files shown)


Scottcjn:
#1713 was already fixed on main; see tests/test_mobile_header_overlap_1713.py. Superseded.

PR #1727, opened 2026-08-20
fix(whisper): /api/transcript/backfill is documented admin-only but anyone can call it; cap batch_size

### LINKED ISSUE #1726: [BUG] POST /api/transcript/backfill is documented admin-only but has no auth check, and batch_size is unbounded
**Endpoint:** `POST /api/transcript/backfill` (`whisper_transcription_blueprint.py:267`, blueprint registered unconditionally at `bottube_server.py:15607`)

### Summary

The blueprint's own docstring marks this endpoint as **admin**, but it has no authorisation check at all. Anyone can start the Whisper worker and queue transcription jobs, and there is no upper bound on how many.

### Steps to reproduce

```
curl -X POST https://bottube.ai/api/transcript/backfill \
     -d '{"force": true, "batch_size": 1000000}'
```

No cookie, no API key, no admin key.

### Expected

`403 Forbidden` — the module header documents it as admin-only:

```
- POST /api/transcript/backfill                   → backfill existing videos (admin)
```

### Actual

`200 {"status": "backfill_queued", "enqueued": N, "force": true}`. `wt.backfill_existing_videos()` calls `start_worker()` and enqueues videos for Whisper transcription — CPU-heavy ML work — and with `force: true` it re-transcribes videos that already have transcripts.

### Root cause

`whisper_transcription_blueprint.py:267-286`:

```python
@whisper_bp.route("/api/transcript/backfill", methods=["POST"])
def trigger_backfill():
    """Backfill transcriptions for all existing videos (admin endpoint)."""
    body, error = _request_json_object()
    if error:
        return error

    force = bool(body.get("force", False))
    batch_size, error = _parse_positive_int(body, "batch_size", 50)
    ...
```

I grepped the whole 292-line module for `session`, `X-API-Key`, `X-Admin-Key`, `api_key`, `admin`, `require`, `before_request`, `g.` — the only hits are the word "admin" inside two docstrings. There is no `before_request` on `whisper_bp`, and the app-level hooks (`bottube_server.py:1468`, `1695`, `18487`) do canonical-host redirects, visitor tracking and latency recording only — none of them authorise anything.

### Second half: `batch_size` is unbounded

`_parse_positive_int` takes no `max_value`, while its query-string sibling **in the same file** does — `_parse_positive_int_arg("limit", 50, max_value=500)` is used by `/api/transcript/search` two functions away. So the search endpoint is capped and the far more expensive backfill is not.

### Why the test suite didn't catch it

`tests/test_whisper_transcription.py` posts to `/api/transcript/backfill` anonymously in three body-validation tests and asserts `400`. Those assertions only hold *because* the endpoint is open, so the suite encoded the missing gate as expected behaviour.

### Environment

Flask + SQLite backend, `main` @ `dd32f25`. Server-side, browser-independent.

### Fix

PR: add the same `_require_admin()` delegate `agent_relationships.py:95` already uses (defers to `bottube_server._require_admin`, fails closed if that import is unavailable), gate the endpoint with it, and cap `batch_size` at 500. Four new tests cover the gate and the cap; the three body-validation tests now run through an `admin_client` fixture since they are about parsing, not authorisation.

### Related, not fixed here

`POST /api/videos/<video_id>/transcript/trigger` (line 226) is also unauthenticated and also enqueues transcription, with `force: true` allowing repeated re-transcription of the same video. I have **not** touched it — unlike backfill it isn't documented as admin, so its intended audience is a product decision rather than a clear-cut bug. Flagging it so you can decide.

---
*Bounty #1102 — functional bug (authorisation).*

Vyacheslav-Tomashevskiy:
Fixes #1726

## The bug

`whisper_transcription_blueprint.py`'s own module header says:

```
- POST /api/transcript/backfill                   → backfill existing videos (admin)
```

but `trigger_backfill()` had no authorisation check of any kind. Grepping the whole 292-line module for `session`, `X-API-Key`, `X-Admin-Key`, `api_key`, `admin`, `require`, `before_request`, `g.` turns up nothing but the word "admin" inside two docstrings, there's no `before_request` on `whisper_bp`, and the blueprint is registered unconditionally (`bottube_server.py:15607`).

So this works from anywhere, with no credentials:

```
curl -X POST /api/transcript/backfill -d '{"force": true, "batch_size": 1000000}'
```

`wt.backfill_existing_videos()` calls `start_worker()` and enqueues Whisper transcription jobs — CPU-heavy ML work — and `force: true` re-does videos that already have transcripts.

**And `batch_size` had no ceiling.** `_parse_positive_int` took no `max_value`, even though its query-string sibling *in the same file* does: `/api/transcript/search` two functions away already caps with `_parse_positive_int_arg("limit", 50, max_value=500)`. The cheap endpoint was capped; the expensive one wasn't.

## The fix

- Added the `_require_admin()` delegate that `agent_relationships.py:95` already uses — defers to `bottube_server._require_admin` so the admin secret and header contract stay in one place, and **fails closed** (403) if that import is unavailable rather than silently allowing the privileged action.
- Gated `trigger_backfill` with it.
- Gave `_parse_positive_int` the same optional `max_value` its sibling has, and capped `batch_size` at `_BACKFILL_BATCH_MAX = 500`.

## Tests

`54 passed` (was 50).

Four new:

- `test_backfill_requires_admin` — anonymous POST gets 403 and `backfill_existing_videos` is never reached (the stub raises if it is).
- `test_backfill_fails_closed_without_admin_key` — a *wrong* key is rejected too, not just a missing one.
- `test_backfill_caps_batch_size` — `batch_size: 10_000_000` is a 400 and enqueues nothing.
- `test_backfill_allows_batch_size_at_the_cap` — exactly 500 still works and is passed through, so the cap isn't off-by-one.

On `main` the two gate tests **fail** with `AssertionError: anonymous caller reached backfill_existing_videos` — i.e. they demonstrate the hole directly. The two cap tests error there rather than fail, since they reference the new `_BACKFILL_BATCH_MAX` constant.

⚠️ **Heads-up on the three pre-existing tests I touched.** `test_backfill_rejects_non_object_json`, `test_backfill_rejects_invalid_batch_size` and `test_backfill_rejects_non_positive_or_non_integer_batch_size` posted to the endpoint anonymously and asserted `400`. Those assertions only held *because* the endpoint was open — the suite had encoded the missing gate as expected behaviour. They now go through an `admin_client` fixture that stubs the gate, since what they actually test is body parsing, not authorisation. No assertion inside them was weakened.

## Not in this PR

`POST /api/videos/<video_id>/transcript/trigger` (line 226) is also unauthenticated and also enqueues transcription work, with `force: true` permitting repeated re-transcription. I deliberately left it alone: unlike backfill it is *not* documented as admin, so who should be allowed to call it is a product decision, not a clear-cut bug. Flagged in the issue.

---

### DIFF
--- a/tests/test_whisper_transcription.py
+++ b/tests/test_whisper_transcription.py
@@ -380,6 +380,20 @@ def flask_client(tmp_db, monkeypatch, tmp_path):
     return app.test_client()
 
 
+@pytest.fixture()
+def admin_client(flask_client, monkeypatch):
+    """flask_client with the admin gate satisfied.
+
+    /api/transcript/backfill is admin-only. These body-validation tests are
+    about parsing, not authorisation, so the gate is stubbed out here; the
+    gate itself is covered by test_backfill_requires_admin* below.
+    """
+    import whisper_transcription_blueprint as wtb
+
+    monkeypatch.setattr(wtb, "_require_admin", lambda: None)
+    return flask_client
+
+
 def test_get_transcript_not_found(flask_client):
     resp = flask_client.get("/api/videos/nonexistent/transcript")
     assert resp.status_code == 404
@@ -542,15 +556,15 @@ def test_search_endpoint_no_query(flask_client):
     assert resp.status_code == 400
 
 
-def test_backfill_rejects_non_object_json(flask_client):
-    resp = flask_client.post("/api/transcript/backfill", json=["not", "an", "object"])
+def test_backfill_rejects_non_object_json(admin_client):
+    resp = admin_client.post("/api/transcript/backfill", json=["not", "an", "object"])
 
     assert resp.status_code == 400
     assert resp.get_json()["error"] == "JSON object required"
 
 
-def test_backfill_rejects_invalid_batch_size(flask_client):
-    resp = flask_client.post("/api/transcript/backfill", json={"batch_size": "not-an-int"})
+def test_backfill_rejects_invalid_batch_size(admin_client):
+    resp = admin_client.post("/api/transcript/backfill", json={"batch_size": "not-an-int"})
 
     assert resp.status_code == 400
     assert resp.get_json()["error"] == "batch_size must be a positive integer"
@@ -558,7 +572,7 @@ def test_backfill_rejects_invalid_batch_size(flask_client):
 
 @pytest.mark.parametrize("batch_size", [0, -1, True, 1.5, float("inf")])
 def test_backfill_rejects_non_positive_or_non_integer_batch_size(
-    flask_client,
+    admin_client,
     monkeypatch,
     batch_size,
 ):
@@ -569,7 +583,7 @@ def test_backfill_rejects_non_positive_or_non_integer_batch_size(
 
     monkeypatch.setattr(wtb.wt, "backfill_existing_videos", fail_backfill)
 
-    resp = flask_client.post("/api/transcript/backfill", json={"batch_size": batch_size})
+    resp = admin_client.post("/api/transcript/backfill", json={"batch_size": batch_size})
 
     assert resp.status_code == 400
     assert resp.get_json()["error"] == "batch_size must be a positive integer"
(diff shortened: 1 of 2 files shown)


Scottcjn:
Verified and merging. Confirmed the premise directly: `trigger_backfill`'s docstring on `main` already said **"(admin endpoint)"** while the body went straight from `_request_json_object()` into the work with no gate — documentation asserting a control the code did not perform. Fixing the code to match the doc is the right direction. `bottube_server._require_admin()` is the correct thing to delegate to, and the `except Exception -> 403` fails closed. 54 tests pass on your branch, including the two that prove a bogus `X-Admin-Key` is rejected rather than just a missing one.

One non-blocking nit for next time: `assert default <= max_value` in `_parse_positive_int` is a bare `assert`, so it vanishes under `python -O`. It's an internal invariant and both call sites are fine, so I am not holding the merge on it — a `raise ValueError` would just be sturdier.

PR #1728, opened 2026-08-20
fix(whisper): default-vs-cap invariant should raise ValueError, not a bare assert (-O strips it)

Vyacheslav-Tomashevskiy:
Follow-up to your review note on #1727:

> One non-blocking nit for next time: `assert default <= max_value` in `_parse_positive_int` is a bare `assert`, so it vanishes under `python -O`. It's an internal invariant and both call sites are fine, so I am not holding the merge on it — a `raise ValueError` would just be sturdier.

You're right, and it's worth doing properly rather than leaving as a known-soft spot. Demonstrated concretely:

```
$ python3    -c 'assert 500 <= 50'   ->  AssertionError
$ python3 -O -c 'assert 500 <= 50'   ->  passes, no error
```

## Scope

Both parsers in the module carried the same bare assert, not just the one I added:

- `_parse_positive_int` (line 52) — mine, from #1727
- `_parse_positive_int_arg` (line 96) — pre-existing, guards `/api/transcript/search`

Both are now an explicit `ValueError` that names the field, and both docstrings document the `Raises:` contract. The behaviour for *callers* is unchanged — this only fires when the code itself is wired with a default above its own cap, which is a programming mistake, not user input.

## Tests

`58 passed` (was 54). Four new:

- `test_parse_positive_int_rejects_default_above_cap` — `ValueError` mentioning the field name.
- `test_parse_positive_int_arg_rejects_default_above_cap` — same for the query-string sibling, inside a request context.
- `test_parse_positive_int_accepts_default_equal_to_cap` — `default == max_value` is still legal, so the fix isn't off-by-one.
- `test_default_vs_cap_invariant_survives_python_O` — runs the invariant in a **subprocess under `python -O`** and asserts it still raises. A bare assert prints `SILENT` there; the raise prints `RAISED`. This is the one that would have caught the original nit.

Three of the four fail on `main`.

No claim attached — this is cleanup on your review note, not a new bounty item.

### DIFF
--- a/tests/test_whisper_transcription.py
+++ b/tests/test_whisper_transcription.py
@@ -667,3 +667,65 @@ def test_backfill_allows_batch_size_at_the_cap(admin_client, monkeypatch):
     assert resp.status_code == 200
     assert seen["batch_size"] == wtb._BACKFILL_BATCH_MAX
     assert resp.get_json()["enqueued"] == 3
+
+
+# ---------------------------------------------------------------------------
+# _parse_positive_int / _parse_positive_int_arg: default-vs-cap invariant
+# ---------------------------------------------------------------------------
+
+def test_parse_positive_int_rejects_default_above_cap():
+    """A default above its own cap is a wiring mistake, not a user error."""
+    import whisper_transcription_blueprint as wtb
+
+    with pytest.raises(ValueError) as excinfo:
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):
ValueError instead of bare assert on the default-vs-cap invariant; follows up my own review nit.

PR #1730, opened 2026-08-20
fix(admin): both admin gates must resolve the same key — RC_ADMIN_KEY only worked for 15 of 39 endpoints

### LINKED ISSUE #1729: [BUG] Two admin gates read different env vars — RC_ADMIN_KEY opens the trust-and-safety endpoints but locks you out of the other 24
**File:** `bottube_server.py` — `ADMIN_KEY` (line 14991) vs `_ts_admin_ok()` (line 20511)

### Summary

There are two admin gates in the module and they read **different environment variables**. An operator who configures only `RC_ADMIN_KEY` ends up with half the admin surface accepting their key and the other half rejecting it.

### The two gates

```python
# line 14991
ADMIN_KEY = os.environ.get("BOTTUBE_ADMIN_KEY", "")
if not ADMIN_KEY:
    ADMIN_KEY = secrets.token_hex(32)
    print(f"[BoTTube] WARNING: BOTTUBE_ADMIN_KEY not set. Generated ephemeral key: {ADMIN_KEY}")
```

```python
# line 20511
def _ts_admin_ok():
    key = request.headers.get("X-Admin-Key", "") or request.args.get("admin_key", "")
    expected = os.environ.get("BOTTUBE_ADMIN_KEY", "") or os.environ.get("RC_ADMIN_KEY", "")
    return bool(expected) and (key == expected)
```

`_require_admin()` compares against `ADMIN_KEY` and guards **24** endpoints. `_ts_admin_ok()` guards **15** trust-and-safety endpoints, including `/admin/blocklist/add` (the SHA-256 content blocklist), `/api/admin/reconcile-anchors`, `/api/admin/provenance/batch` and `/admin/moderation/reports`.

`RC_ADMIN_KEY` appears in exactly one line in the entire tree — that one. Nothing else knows the alias exists.

### Steps to reproduce

1. Deploy with `RC_ADMIN_KEY=secret123` and **no** `BOTTUBE_ADMIN_KEY`.
2. `curl -H 'X-Admin-Key: secret123' https://bottube.ai/admin/moderation/reports` → **200**
3. `curl -H 'X-Admin-Key: secret123' <any endpoint guarded by _require_admin>` → **403**

### Expected

One admin secret opens the whole admin surface.

### Actual

Reproduced against verbatim copies of both gates:

```
Scenario: operator set only RC_ADMIN_KEY=secret123

  _ts_admin_ok    (15 T&S endpoints): 200 OK
  _require_admin  (24 endpoints):     403 (key mismatch)
```

And it is worse than a plain mismatch: with `BOTTUBE_ADMIN_KEY` unset, line 14993 replaces `ADMIN_KEY` with `secrets.token_hex(32)`. So those 24 endpoints accept **only a random ephemeral key printed once at boot** — the operator sees their configured key working on some admin pages and silently failing on others, which reads as flakiness rather than as a config error.

### Not a security hole

Both gates fail **closed** — `_ts_admin_ok` requires a non-empty `expected`, and the ephemeral-key fallback means an unset env never yields an open door. This is a correctness / operability bug, not an auth bypass. Filing it at that severity deliberately.

### Related observation, not fixed here

`_ts_admin_ok` accepts the secret via query string (`?admin_key=`) with no warning, while `_require_admin` accepts `?key=` and explicitly logs a deprecation notice because "secrets in URLs leak into logs/history/referers". Since the `reconcile-anchors` docstring says it's driven by a 6-hour systemd timer, I did **not** touch query-param support — removing it could break that timer. Flagging so you can decide.

### Environment

`main` @ `4f1fae5`.

### Fix

PR: both gates resolve through a single `_admin_key_from_env()`. `BOTTUBE_ADMIN_KEY` still wins when both are set, the boot warning now names both variables, and behaviour when neither is set is unchanged.

---

Vyacheslav-Tomashevskiy:
Fixes #1729

## The bug

Two admin gates, two different env lookups:

| gate | reads | guards |
|---|---|---|
| `_require_admin()` via `ADMIN_KEY` (line 14991) | `BOTTUBE_ADMIN_KEY` only | 24 endpoints |
| `_ts_admin_ok()` (line 20511) | `BOTTUBE_ADMIN_KEY` **or** `RC_ADMIN_KEY` | 15 trust-and-safety endpoints, incl. `/admin/blocklist/add` |

`RC_ADMIN_KEY` occurs exactly once in the whole tree — inside `_ts_admin_ok`. Nothing else knows the alias exists.

Set only `RC_ADMIN_KEY` and the admin surface splits in half:

```
Scenario: operator set only RC_ADMIN_KEY=secret123

  _ts_admin_ok    (15 T&S endpoints): 200 OK
  _require_admin  (24 endpoints):     403 (key mismatch)
```

Compounding it, line 14993 replaces an unset `ADMIN_KEY` with `secrets.token_hex(32)` — so those 24 endpoints accept only a random ephemeral key printed once at boot. The operator's configured key works on some admin pages and fails on others, which presents as flakiness rather than as a config mistake.

**Both gates fail closed**, so this is a correctness/operability bug, not an auth bypass — filing it at that severity on purpose.

## The fix

One resolver, used by both:

```python
def _admin_key_from_env():
    return os.environ.get("BOTTUBE_ADMIN_KEY", "") or os.environ.get("RC_ADMIN_KEY", "")

ADMIN_KEY = _admin_key_from_env()
```

`BOTTUBE_ADMIN_KEY` still wins when both are set, the ephemeral-key fallback is untouched, and the boot warning now names both variables instead of only one.

## Tests

`tests/test_admin_key_single_source.py`, 7 cases: resolution across primary-only / alias-only / both / neither, both gates agreeing on the alias, both rejecting a wrong key, and the T&S gate still failing closed with no env at all.

**Honest note on what these prove:** 6 of the 7 fail on `main`, but most of them fail because `_admin_key_from_env` doesn't exist there yet — they're tests for the new contract, not a pure before/after regression. The bug itself is demonstrated by the standalone repro in #1729, which runs verbatim copies of both gates and shows the 200/403 split. I'd rather say that than let the "6 failed on main" number imply more than it does.

Related suites still green:

```
$ python3 -m pytest tests/test_admin_key_single_source.py tests/test_whisper_transcription.py     tests/test_channel_alias_route.py -q
64 passed
```

## Deliberately not touched

`_ts_admin_ok` accepts the secret via `?admin_key=` with no warning, while `_require_admin` logs a deprecation notice for `?key=` because secrets in URLs leak into logs and referers. The `reconcile-anchors` docstring says that endpoint is driven by a 6-hour systemd timer, so removing query-param support could break it — that's your call, not something to slip into this PR.

---

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -14988,10 +14988,22 @@ def giveaway_leaderboard_api():
 # Admin: Visitor Analytics
 # ---------------------------------------------------------------------------
 
-ADMIN_KEY = os.environ.get("BOTTUBE_ADMIN_KEY", "")
+def _admin_key_from_env():
+    """Resolve the admin secret from the environment.
+
+    Single source of truth: BOTTUBE_ADMIN_KEY is the documented name and
+    RC_ADMIN_KEY is an accepted alias. Both admin gates in this module go
+    through here, so an operator who sets only one of them gets a
+    consistent answer everywhere instead of half the admin surface
+    rejecting the key they configured.
+    """
+    return os.environ.get("BOTTUBE_ADMIN_KEY", "") or os.environ.get("RC_ADMIN_KEY", "")
+
+
+ADMIN_KEY = _admin_key_from_env()
 if not ADMIN_KEY:
     ADMIN_KEY = secrets.token_hex(32)
-    print(f"[BoTTube] WARNING: BOTTUBE_ADMIN_KEY not set. Generated ephemeral key: {ADMIN_KEY}")
+    print(f"[BoTTube] WARNING: neither BOTTUBE_ADMIN_KEY nor RC_ADMIN_KEY set. Generated ephemeral key: {ADMIN_KEY}")
 
 
 @app.route("/api/admin/visitors")
@@ -20511,7 +20523,7 @@ def submit_report():
 def _ts_admin_ok():
     """Check if the current request has admin privileges for trust-and-safety endpoints. Returns: True if admin."""
     key = request.headers.get("X-Admin-Key", "") or request.args.get("admin_key", "")
-    expected = os.environ.get("BOTTUBE_ADMIN_KEY", "") or os.environ.get("RC_ADMIN_KEY", "")
+    expected = _admin_key_from_env()
     return bool(expected) and (key == expected)
 
 
--- a/tests/test_admin_key_single_source.py
+++ b/tests/test_admin_key_single_source.py
@@ -1,0 +1,85 @@
+# SPDX-License-Identifier: MIT
+"""The two admin gates must resolve the same secret.
+
+`bottube_server` has two independent admin checks:
+
+* `_require_admin()` (24 call sites) compared against the module constant
+  `ADMIN_KEY`, which was read from **BOTTUBE_ADMIN_KEY only**.
+* `_ts_admin_ok()` (15 trust-and-safety call sites, including the
+  `/admin/blocklist/add` hash blocklist) read
+  **BOTTUBE_ADMIN_KEY or RC_ADMIN_KEY**.
+
+So an operator who configured only `RC_ADMIN_KEY` got a split brain: the
+trust-and-safety endpoints accepted their key while all 24 `_require_admin`
+endpoints rejected it — and because an unset `BOTTUBE_ADMIN_KEY` makes the
+module fall back to `secrets.token_hex(32)`, those endpoints would only
+accept a random ephemeral key printed once at boot.
+
+Both gates now resolve through `_admin_key_from_env()`.
+"""
+
+import importlib
+
+import pytest
+
+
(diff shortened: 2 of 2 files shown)


Vyacheslav-Tomashevskiy:
Closing this one myself — #1739 landed first and resolved #1729 in the opposite direction: both gates now read `BOTTUBE_ADMIN_KEY` only, and `tests/test_admin_key_env.py::test_trust_safety_gate_rejects_rc_only_key` pins that `RC_ADMIN_KEY` is *rejected*. My patch made `RC_ADMIN_KEY` an accepted alias, which would break that test and contradict the direction you picked.

I re-grepped the tree: after #1739 `RC_ADMIN_KEY` appears only inside that test file — no deploy script or doc still sets it — so there is no leftover gap to patch. Nothing needed here.

Issue #1719, opened 2026-08-20
[BUG] All five authenticated /api/usdc/* endpoints return HTTP 500 — `SELECT name FROM agents` (no such column)


Vyacheslav-Tomashevskiy:
**URL:** `/api/usdc/balance`, `/api/usdc/deposit`, `/api/usdc/tip`, `/api/usdc/premium`, `/api/usdc/payout` (blueprint `usdc_blueprint.py`, registered at `bottube_server.py:15402`)

### Summary

Every authenticated USDC endpoint returns HTTP 500. The shared auth helper selects a column that does not exist in the `agents` table, so the request dies before any handler logic runs. A user who has already sent USDC to the treasury address can never get it credited.

### Steps to reproduce

1. Register an agent, note its `X-API-Key`.
2. `GET /api/usdc/balance` with header `X-API-Key: <key>`.

### Expected

`200 {"agent": "...", "balance_usdc": 0.0, ...}`

### Actual

**HTTP 500.** Server log: `sqlite3.OperationalError: no such column: name`.

Same for `POST /api/usdc/deposit`, `/api/usdc/tip`, `/api/usdc/premium`, `/api/usdc/payout` — all five call the same helper.

### Root cause

`usdc_blueprint.py:236-243`:

```python
def get_authenticated_agent():
    api_key = request.headers.get("X-API-Key") or request.args.get("api_key")
    if not api_key:
        return None
    db = get_db()
    agent = db.execute("SELECT name FROM agents WHERE api_key = ?", (api_key,)).fetchone()
    return agent["name"] if agent else None
```

The schema at `bottube_server.py:1814-1841` is `agents(id, agent_name, display_name, api_key, ...)`. There is no `name` column, and no `ALTER TABLE agents ADD COLUMN name` anywhere in the repo. Reproduced against the real schema:

```
== usdc_blueprint.get_authenticated_agent ==
  RAISES: OperationalError no such column: name
```

### Second bug in the same file — tipping by video_id

`usdc_blueprint.py:428`:

```python
video = db.execute("SELECT agent FROM videos WHERE id = ?", (video_id,)).fetchone()
```

`videos` has `agent_id` (integer FK) and `video_id` (the public string identifier); `id` is the integer primary key. There is no `agent` column:

```
== usdc_blueprint.usdc_tip ==
  RAISES: OperationalError no such column: agent
```

[...]

Issue #1717, opened 2026-08-20
[BUG] Every ERG deposit returns HTTP 500, credits 0 RTC and permanently burns the tx_id (earnings column mismatch)


Vyacheslav-Tomashevskiy:
**URL:** `POST /api/ergo/deposit` (blueprint `ergo_bridge_blueprint.py`, registered at `bottube_server.py:15429`)

### Summary

The Ergo bridge can never credit a deposit. `_award_rtc()` writes to a column that does not exist in the `earnings` table, so the credit always raises and rolls back — but the `ergo_deposits` row was already committed with `status='credited'`, so the transaction id is claimed forever. Real ERG lands in the platform wallet and the depositor gets nothing, with no way to retry.

### Steps to reproduce

1. Send >= 0.01 ERG to the address published by `GET /api/ergo/info`.
2. Wait for 3 confirmations.
3. `POST /api/ergo/deposit` with `{"tx_id": "<hex>"}` and a valid `X-API-Key`.
4. `POST` the same `tx_id` again.

### Expected

3. → `200 {"ok": true, "rtc_credited": 7.92}` and `rtc_balance` increases.

### Actual

3. → **HTTP 500**, `rtc_balance` unchanged (0 RTC credited).
4. → **HTTP 409 `{"error": "Transaction already claimed"}`** — forever. `GET /api/ergo/history` still lists the deposit as `status='credited'`.

### Root cause

`ergo_bridge_blueprint.py:214-224`:

```python
db.execute(
    "INSERT INTO earnings (agent_id, amount, source, created_at) VALUES (?, ?, ?, ?)",
    (agent_id, amount, reason, time.time()),
)
```

The schema in `bottube_server.py:1918-1926` is:

```sql
CREATE TABLE IF NOT EXISTS earnings (
    id INTEGER PRIMARY KEY,
    agent_id INTEGER NOT NULL,
    amount REAL NOT NULL,
    reason TEXT NOT NULL,
    video_id TEXT DEFAULT '',
    created_at REAL NOT NULL,
    FOREIGN KEY (agent_id) REFERENCES agents(id)
);
```

[...]

PR #1734, opened 2026-08-20
fix(db): bridges opened a different database file than the one holding their tables

### LINKED ISSUE #1733: [BUG] Crypto bridges read BOTTUBE_DB while their tables are created under BOTTUBE_DB_PATH — different database files in 3 of 4 configurations
**Files:** `bottube_server.py:15463-15490` vs `usdc_blueprint.py:48`, `ergo_bridge_blueprint.py:70,79`, `wrtc_bridge_blueprint.py:92`, `base_wrtc_bridge_blueprint.py:90`

### Summary

The four crypto-bridge blueprints resolve their database path from a **different environment variable, with a different default**, than the code that creates their tables. In three of four configurations they end up on a different SQLite file.

### The two resolutions

`bottube_server.py` — creating the tables:

```python
_usdc_db_path      = os.environ.get("BOTTUBE_DB_PATH", str(DB_PATH))   # 15463
_wrtc_db_path      = os.environ.get("BOTTUBE_DB_PATH", str(DB_PATH))   # 15472
_base_wrtc_db_path = os.environ.get("BOTTUBE_DB_PATH", str(DB_PATH))   # 15481
_ergo_db_path      = os.environ.get("BOTTUBE_DB_PATH", str(DB_PATH))   # 15490
```

`usdc_blueprint.py:48` and friends — opening the connection at request time:

```python
db_path = os.environ.get("BOTTUBE_DB", "/root/bottube/bottube.db")
```

`DB_PATH` is `BASE_DIR / "bottube.db"`, and `BASE_DIR` is `BOTTUBE_BASE_DIR` or the directory containing `bottube_server.py` (line 84).

### Reproduction

Running both resolutions verbatim across the configuration matrix:

```
A. neither set, app not installed in /root/bottube
   server creates tables in : /opt/bottube/bottube.db
   bridge reads at runtime  : /root/bottube/bottube.db      -> DIFFERENT FILES

B. operator sets BOTTUBE_DB_PATH (the main server's name)
   server creates tables in : /data/bottube.db
   bridge reads at runtime  : /root/bottube/bottube.db      -> DIFFERENT FILES

C. operator sets BOTTUBE_DB (the bridges' name)
   server creates tables in : /opt/bottube/bottube.db
   bridge reads at runtime  : /data/bottube.db              -> DIFFERENT FILES

D. both set to the same value
   -> same file
```

Only D works, and D requires knowing that two differently-named variables both exist and must be kept in sync. Neither name is documented alongside the other.

### Impact

`init_usdc_tables` / `init_wrtc_tables` / `init_base_wrtc_tables` / `init_ergo_tables` run against one file; every `get_db()` in those blueprints runs against another. Depending on what exists at the second path that is either `no such table: usdc_balances`, or — worse — a stale second database with its own `agents` rows and balances, silently serving deposits and withdrawals from the wrong ledger.

Note this is separate from the two schema bugs already fixed in #1717 and #1719; those were column names inside one database, this is *which database file*.

### Also on the old name (not changed in the PR)

`gemini_blueprint.py` (4 sites), `banano_payout.py:25` and `bottube_engage.py:32` also read `BOTTUBE_DB`. They aren't part of the init/runtime mismatch above because nothing in `bottube_server` initialises them via `BOTTUBE_DB_PATH`, so I left them alone rather than widening the diff — but they'd benefit from the same resolver if you want a follow-up.

### Environment

`main` @ `4f1fae5`.

### Fix

PR: a single `bottube_db.resolve_db_path()` — `BOTTUBE_DB_PATH` first, `BOTTUBE_DB` kept as an alias so deployments configured that way keep working, default computed from `BOTTUBE_BASE_DIR` like `bottube_server` does instead of a hardcoded `/root` path. Both the server's four init sites and the four blueprints call it. 15 tests, 8 fail on `main`.

---

Vyacheslav-Tomashevskiy:
Fixes #1733

## The bug

The four crypto-bridge blueprints resolve their database path from a different environment variable, with a different default, than the code that creates their tables.

`bottube_server.py:15463-15490`, creating the tables:

```python
_usdc_db_path = os.environ.get("BOTTUBE_DB_PATH", str(DB_PATH))
```

`usdc_blueprint.py:48` (and `ergo_bridge_blueprint.py:70,79`, `wrtc_bridge_blueprint.py:92`, `base_wrtc_bridge_blueprint.py:90`), opening the connection per request:

```python
db_path = os.environ.get("BOTTUBE_DB", "/root/bottube/bottube.db")
```

Across the configuration matrix, only one of four combinations lands both on the same file:

| configuration | server creates tables in | bridge reads |  |
|---|---|---|---|
| neither set, app not in `/root/bottube` | `<BASE_DIR>/bottube.db` | `/root/bottube/bottube.db` | ✗ |
| `BOTTUBE_DB_PATH` set | `/data/bottube.db` | `/root/bottube/bottube.db` | ✗ |
| `BOTTUBE_DB` set | `<BASE_DIR>/bottube.db` | `/data/bottube.db` | ✗ |
| both set to the same value | `/data/bottube.db` | `/data/bottube.db` | ✓ |

So `init_*_tables` runs against one file and every `get_db()` against another unless the operator knows both names exist and keeps them in sync. That is either `no such table: usdc_balances`, or a stale second database with its own `agents` rows and balances quietly serving deposits and withdrawals from the wrong ledger.

This is a different failure from #1717 / #1719 — those were column names *inside* one database, this is *which file*.

## The fix

New `bottube_db.py` with one `resolve_db_path()`:

1. `BOTTUBE_DB_PATH` — the name `bottube_server` uses.
2. `BOTTUBE_DB` — kept as an alias, so deployments already configured that way keep working.
3. `<BOTTUBE_BASE_DIR or this directory>/bottube.db` — the same default `bottube_server` computes, instead of a hardcoded `/root` path.

Both the server's four init sites and the four blueprints call it, so there is one decision point rather than eight. Nothing else in the diff.

## Tests

`tests/test_bridge_db_path.py`, 15 cases: precedence, the legacy alias, the default no longer pointing at `/root`, agreement between server and bridge across **all four** configurations, no blueprint still hardcoding the `/root` path, and each blueprint actually importing the shared resolver.

8 fail on `main`. Related suites stay green:

```
$ python3 -m pytest tests/test_bridge_db_path.py tests/test_usdc_amount_validation.py     tests/test_usdc_deposit_sender_binding.py tests/test_ergo_bridge_validation.py     tests/test_ergo_bridge_registration.py tests/test_wrtc_bridge_validation.py     tests/test_base_wrtc_bridge_validation.py -q
92 passed
```

While writing the parametrised agreement test I found that patching only the blueprints still left the `BOTTUBE_DB`-only case broken — the server was still reading `BOTTUBE_DB_PATH` inline. That's why the server's four init sites go through the resolver too; the test is what caught it.

## Left alone deliberately

`gemini_blueprint.py` (4 sites), `banano_payout.py:25` and `bottube_engage.py:32` also read `BOTTUBE_DB`, but nothing in `bottube_server` initialises them via `BOTTUBE_DB_PATH`, so they aren't part of this mismatch. Flagged in the issue rather than widened into this diff.

---

### DIFF
--- a/avap_blueprint.py
+++ b/avap_blueprint.py
@@ -25,6 +25,7 @@ import sqlite3
 import time
 from pathlib import Path
 
+from bottube_db import resolve_db_path
 from flask import Blueprint, jsonify, request
 from nacl.signing import VerifyKey
 from nacl.exceptions import BadSignatureError
@@ -32,7 +33,7 @@ from nacl.exceptions import BadSignatureError
 avap_bp = Blueprint("avap", __name__)
 
 BASE_DIR = Path(__file__).resolve().parent
-DB_PATH = BASE_DIR / "bottube.db"
+DB_PATH = resolve_db_path()
 AVAP_VERSION = "1.0"
 ANCHOR_PREFIX = "rc2"  # RustChain Node 2 (.153) anchor record
 
--- a/base_wrtc_bridge_blueprint.py
+++ b/base_wrtc_bridge_blueprint.py
@@ -13,6 +13,7 @@ Deposit verification uses raw JSON-RPC (no web3 dependency).
 Withdrawal processing uses web3.py for signing and sending.
 """
 
+from bottube_db import resolve_db_path
 import json
 import hmac
 import os
@@ -87,7 +88,7 @@ def get_db():
     """Get database connection from Flask g."""
     if "db" in g:
         return g.db
-    db_path = os.environ.get("BOTTUBE_DB", "/root/bottube/bottube.db")
+    db_path = resolve_db_path()
     g.db = sqlite3.connect(db_path)
     g.db.row_factory = sqlite3.Row
     return g.db
--- a/bottube_db.py
+++ b/bottube_db.py
@@ -1,0 +1,34 @@
+# SPDX-License-Identifier: MIT
+"""Single place that decides which SQLite file the app talks to.
+
+The bridge blueprints used to resolve this themselves with
+``os.environ.get("BOTTUBE_DB", "/root/bottube/bottube.db")`` while
+``bottube_server`` created their tables using
+``os.environ.get("BOTTUBE_DB_PATH", str(DB_PATH))``. Two names, two
+defaults, so unless an operator happened to set *both* variables to the
+same value the bridges read a different file than the one their tables
+were created in.
+
+Resolution order:
+
+1. ``BOTTUBE_DB_PATH`` — the name ``bottube_server`` uses.
+2. ``BOTTUBE_DB`` — the name the bridge blueprints used; kept as an alias
+   so existing deployments configured that way keep working.
+3. ``<BOTTUBE_BASE_DIR or this directory>/bottube.db`` — the same default
+   ``bottube_server`` computes, rather than a hardcoded ``/root`` path.
+"""
+from __future__ import annotations
+
+import os
+from pathlib import Path
+
+
+def resolve_db_path() -> str:
+    """Return the path of the SQLite database every module should share."""
+    explicit = os.environ.get("BOTTUBE_DB_PATH", "") or os.environ.get("BOTTUBE_DB", "")
+    if explicit:
+        return explicit
+    base_dir = Path(
+        os.environ.get("BOTTUBE_BASE_DIR", str(Path(__file__).resolve().parent))
+    )
+    return str(base_dir / "bottube.db")
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -15457,10 +15457,14 @@ from paypal_packages import store_bp, init_store_db
 init_store_db()  # Create store tables if needed
 app.register_blueprint(store_bp)
 
+# Every bridge blueprint and this module must agree on one database file;
+# see bottube_db.resolve_db_path for the resolution order.
+from bottube_db import resolve_db_path as _resolve_shared_db_path
+
 # USDC Payment Integration (Base Chain)
 from usdc_blueprint import usdc_bp, init_usdc_tables
 import sqlite3 as _usdc_sqlite3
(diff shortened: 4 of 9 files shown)


Vyacheslav-Tomashevskiy:
Pushed one more commit while this is still unreviewed — flagging it rather than letting the scope quietly grow under you.

Running the same "who else decides which database file to open?" check across the tree turned up two more **live** blueprints doing it independently, and both are worse than the `BOTTUBE_DB` / `BOTTUBE_DB_PATH` split this PR started with:

```python
news_routes.py:11     DB_PATH = "/root/bottube/bottube.db"      # no env override at all
avap_blueprint.py:34  BASE_DIR = Path(__file__).resolve().parent
```

- **`news_routes`** gives an operator no way to point it anywhere. `/news` queries an absolute path that exists on exactly one machine; anywhere else it silently opens an empty database and the section renders as if there were no news.
- **`avap_blueprint`** recomputes `BASE_DIR` without honouring `BOTTUBE_BASE_DIR` — the variable `bottube_server.py:84` reads. So setting it moves the app's database while avap keeps using the checkout directory, and `init_avap_tables()` (called unconditionally at `bottube_server.py:15658`) then creates the attestation tables in that second file.

Both now go through the same `resolve_db_path()`, plus 5 more tests — 20 in the file now, and the related suites are still green at 95 passed.

This is deliberately *not* the same bucket as `gemini_blueprint` / `banano_payout` / `bottube_engage`, which I said I'd leave alone: those read `BOTTUBE_DB`, so an operator can at least steer them. These two couldn't be steered at all. Still happy to split them into a separate PR if you'd rather keep this diff to the four bridges — say the word and I'll rebase them out.

PR #1735, opened 2026-08-20
fix(bridge): wrtc_bridge.py's _debit_rtc still has the TOCTOU race fixed elsewhere (#1636 follow-up)

Vyacheslav-Tomashevskiy:
Closes the "same race, different file" callout from @Scottcjn on #1636.

`ergo_bridge_blueprint.py`'s `_debit_rtc` had this exact check-then-update TOCTOU race and was fixed in #1711 by moving the balance comparison into the `UPDATE ... WHERE rtc_balance >= ?` clause. `wrtc_bridge.py` kept its own separate copy of the same-named function with the original two-statement shape intact:

```python
row = db.execute("SELECT rtc_balance FROM agents WHERE id = ?", (agent_id,)).fetchone()
if not row or row["rtc_balance"] < amount:
    return False
db.execute("UPDATE agents SET rtc_balance = rtc_balance - ? WHERE id = ?", (amount, agent_id))
```

Fix mirrors the already-merged pattern exactly:

```python
cursor = db.execute(
    "UPDATE agents SET rtc_balance = rtc_balance - ? WHERE id = ? AND rtc_balance >= ?",
    (amount, agent_id, amount),
)
if cursor.rowcount == 0:
    return False
db.commit()
return True
```

**Tests** mirror the two-connection stale-read pattern already established in `tests/test_rtc_debit_balance_guard.py::test_guarded_update_is_atomic_against_a_stale_read`. 3 new tests; the race test fails on unmodified `wrtc_bridge.py` — worth noting the *exact* failure mode: since the old code never calls `db.commit()`, the first connection's uncommitted write holds SQLite's connection-level lock and the second connection's write times out (`OperationalError: database is locked`) rather than silently overdrawing. Same underlying non-atomic bug, a different (and honestly worse — it hangs/errors under contention instead of just being wrong) symptom. Passes clean with the fix.

**Honesty note, please read before merging:** I grepped `import wrtc_bridge` / `from wrtc_bridge import` across every non-test `.py` file in the tree and got zero hits. The blueprint actually registered as `wrtc_bp` in `bottube_server.py` comes from the similarly-named `wrtc_bridge_blueprint.py`, which already does the atomic guarded `UPDATE` inline (no separate `_debit_rtc` helper) — so the live wRTC withdrawal path was never vulnerable to this specific race. I'm submitting this because you flagged it directly and asked for it ("If you want to take it, that'd be a welcome PR"), and an orphaned module carrying a known-bad pattern under a since-fixed sibling's exact function name is a footgun for whoever next copies from it or wires it in — but I didn't want to imply live production impact I can't back up.

/claim
RTC wallet: RTCd1554f0f35576faf01d386a6be1c947f560dd0b7

### DIFF
--- a/tests/test_wrtc_bridge_py_debit_rtc_toctou_poc.py
+++ b/tests/test_wrtc_bridge_py_debit_rtc_toctou_poc.py
@@ -1,0 +1,103 @@
+# SPDX-License-Identifier: MIT
+"""Regression test for the TOCTOU race in wrtc_bridge.py's ``_debit_rtc``
+(bottube#1620, follow-up).
+
+``ergo_bridge_blueprint.py`` had this exact bug and it was fixed in #1711 by
+moving the balance comparison into the ``UPDATE ... WHERE rtc_balance >= ?``
+clause so read-and-write become one atomic statement (see
+``tests/test_rtc_debit_balance_guard.py::test_guarded_update_is_atomic_against_a_stale_read``,
+whose two-connection pattern this test mirrors). ``wrtc_bridge.py`` kept its
+own separate copy of ``_debit_rtc`` with the original check-then-update shape
+(a SELECT to check the balance, then an unconditional UPDATE) -- flagged by
+@Scottcjn directly when closing #1636: "``_debit_rtc`` in
+``wrtc_bridge.py:152`` still has the original check-then-update shape...
+Same race, different file."
+
+NOTE ON LIVE IMPACT: ``wrtc_bridge.py`` is not imported anywhere in the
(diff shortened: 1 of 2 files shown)


Scottcjn:
Real fix, and I want it in — but it conflicts with current `main` after today's merges. Please rebase onto `main` and it goes in on the next pass. Thank you — Sophia

Scottcjn:
Landed via #2189 — main already had the WHERE-clause CAS, so what remained of yours was the commit-on-success (your PoC test caught that main still lacked it) plus the docstring and the test itself. Rebased by a maintainer, authorship preserved. — Sophia

PR #1769, opened 2026-08-26
docs: add docstrings to feed pagination and report input validation tests (28 functions)

Vyacheslav-Tomashevskiy:
## Summary
Added docstrings to every previously-undocumented function across two test files:

- `tests/test_feed_query_param_validation.py` — 14 functions documented
- `tests/test_report_input_validation.py` — 14 functions documented

28 functions documented total, docstrings-only, no logic changed. Each docstring focuses on *why* the test exists (e.g. why zero/negative/float/null/NaN page values are each their own distinct test rather than one parametrized "bad input" case, why `reason: null` and `details: null` get opposite verdicts, why the falsy-empty-array `[]` body is checked separately from the general non-object case on both the video and comment report endpoints).

## Testing
```
pytest -q tests/test_feed_query_param_validation.py tests/test_report_input_validation.py
```
21 passed.

/claim docstring batch

### DIFF
--- a/tests/test_feed_query_param_validation.py
+++ b/tests/test_feed_query_param_validation.py
@@ -14,6 +14,7 @@ authenticated /api/feed/subscriptions routes.
 
 
 def test_feed_rejects_non_integer_page(client):
+    """A non-numeric `page` must 400 with a message naming the field and the reason."""
     response = client.get("/api/feed?page=abc")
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):
Test-module docstrings only; docstring gate prices the claim.

PR #1771, opened 2026-08-26
fix(cron): banano_payout.py and bottube_engage.py still read raw BOTTUBE_DB

Vyacheslav-Tomashevskiy:
## Summary
The bridge/gemini/news db-path fixes (#1736-#1738) unified everything else onto `bottube_db.resolve_db_path()` (`BOTTUBE_DB_PATH` -> `BOTTUBE_DB` -> `BOTTUBE_BASE_DIR`), but these two standalone cron/operational scripts were missed - they still did a bare `os.environ.get("BOTTUBE_DB", ...)`.

On a `BOTTUBE_DB_PATH`-only deployment (the name the main server and table-creation code actually use), both scripts silently fall back to the hardcoded `/root/bottube/bottube.db` default instead of the configured path:
- `banano_payout.py` would process withdrawals against a database with none of the server's tables
- `bottube_engage.py` would read/post comments sourced from the wrong (likely empty) database

## Fix
- `banano_payout.py` and `bottube_engage.py` now call `resolve_db_path()` instead of reading `BOTTUBE_DB` directly

## Testing
- Added `tests/test_standalone_scripts_db_path.py`: 4 regression tests, 2 of which fail on `main` without this fix (verified locally) and pass with it
- `python3 -m pytest -q tests/test_standalone_scripts_db_path.py` - 4/4 passing

/claim

### DIFF
--- a/banano_payout.py
+++ b/banano_payout.py
@@ -18,11 +18,18 @@ import sys
 import time
 import urllib.request
 
+from bottube_db import resolve_db_path
+
 # ---------------------------------------------------------------------------
 # Configuration
 # ---------------------------------------------------------------------------
 
-DB_PATH = os.environ.get("BOTTUBE_DB", "/root/bottube/bottube.db")
+# Reads BOTTUBE_DB_PATH first (the name the main server and table-creation
+# code use), then falls back to the legacy BOTTUBE_DB alias and finally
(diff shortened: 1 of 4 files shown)


Scottcjn (approved):
Cron scripts now resolve the DB path through bottube_db.resolve_db_path() like the server; test covers the env precedence.

PR #1772, opened 2026-08-26
fix(wrtc): refund timed-out withdrawal processing

Vyacheslav-Tomashevskiy:
## Summary
Fixes #1551.

A withdrawal whose on-chain send times out is currently left in a stuck state: `process_withdrawals()` catches the timeout but doesn't mark the row `failed` or restore the debited RTC balance. That's inconsistent with the existing non-timeout failure path, which does both.

## Fix
- mark timed-out `process_withdrawals()` jobs as `failed`
- refund the full debited amount (`net_wrtc + fee`) on timeout, matching the existing non-timeout failure path

## Testing
- Added `tests/test_wrtc_bridge_timeouts.py`: reproduces the timeout via a `subprocess.run` monkeypatch that raises `TimeoutExpired`, asserts the withdrawal row ends up `failed` and the agent's `rtc_balance` is restored to its pre-debit value
- `pytest -q tests/test_wrtc_bridge_timeouts.py` - passing

/claim

PR #1773, opened 2026-08-26
fix(wrtc): refund and mark failed on timeout/exception during withdrawal processing

### LINKED ISSUE #1551: wRTC bridge withdrawal stuck queued without Solana tx signature

Vyacheslav-Tomashevskiy:
## Summary
Fixes #1551.

`process_withdrawals()` already marks the row `failed` and refunds the debited RTC balance when the send helper explicitly reports failure, but the `subprocess.TimeoutExpired` and generic `Exception` branches only recorded the error in the response — the withdrawal row stayed `pending` forever and the debited balance was never restored. A timed-out or otherwise-erroring send permanently loses the agent their RTC with no way to retry or recover it.

## Fix
- Add the same mark-failed + refund (`net_wrtc + fee`) to both the timeout and generic-exception branches, matching the existing explicit-failure path.

## Testing
- Added `tests/test_wrtc_bridge_timeouts.py`: monkeypatches `subprocess.run` to raise `TimeoutExpired`, asserts the withdrawal row ends up `failed` and the agent's `rtc_balance` is restored to its pre-debit value
- Fails on `main` (row stays `pending`, balance not refunded), passes with this fix
- `pytest -q tests/test_wrtc_bridge_timeouts.py tests/test_wrtc_bridge_validation.py tests/test_base_wrtc_bridge_validation.py` — 38/38 passing, no regressions

(Note: an earlier PR #1772 for this same fix had a messy ~700-line diff from a CRLF/LF line-ending mismatch on my end — closed in favor of this one, which has a clean 16-line diff.)

/claim

### DIFF
--- a/tests/test_wrtc_bridge_timeouts.py
+++ b/tests/test_wrtc_bridge_timeouts.py
@@ -1,0 +1,66 @@
+import sqlite3
+from pathlib import Path
+
+import wrtc_bridge as wb
+
+
+class _FakeG(dict):
+    __getattr__ = dict.__getitem__
+    __setattr__ = dict.__setitem__
+
+
+class _TimeoutRun:
+    def __call__(self, *args, **kwargs):
+        raise wb.subprocess.TimeoutExpired(cmd=args[0], timeout=kwargs.get("timeout", 0))
+
+
+def _make_db(tmp_path: Path):
+    db_path = tmp_path / "bridge.db"
+    conn = sqlite3.connect(db_path)
+    conn.row_factory = sqlite3.Row
+    conn.executescript(
+        """
+        CREATE TABLE agents (
+            id INTEGER PRIMARY KEY,
+            rtc_balance REAL NOT NULL DEFAULT 0
+        );
+        CREATE TABLE wrtc_withdrawals (
+            id INTEGER PRIMARY KEY AUTOINCREMENT,
+            agent_id INTEGER NOT NULL,
+            amount_wrtc REAL NOT NULL,
+            fee_wrtc REAL NOT NULL,
+            net_wrtc REAL NOT NULL,
+            to_address TEXT NOT NULL,
+            tx_signature TEXT,
+            status TEXT DEFAULT 'pending',
+            created_at REAL NOT NULL,
+            completed_at REAL
+        );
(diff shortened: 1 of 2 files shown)


Vyacheslav-Tomashevskiy:
/claim

Opened clean fork PR with fix and regression tests: https://github.com/Vyacheslav-Tomashevskiy/bottube/pull/2

Scottcjn (approved):
Timeout/exception branches now mark the withdrawal failed and refund net+fee, matching the explicit-failure branch. Closes #1551.

PR #1792, opened 2026-08-28
fix(analytics): validate metric param on /api/top-videos (#1790)

### LINKED ISSUE #1790: analytics top-videos echoes an unsupported metric while silently sorting by views
**Reported by @AInoAKARI** via the bounty email fallback (2026-08-22). Filed by the maintainer on their behalf; credit and the 2 RTC under rustchain-bounties#520 belong to them.

`analytics_blueprint.py` `/analytics/api/top-videos` reads `metric = request.args.get('metric', 'views')`, branches on `views` / `engagement` / `tips`, and otherwise falls through to `ORDER BY view_count DESC`. The response then returns `"metric": metric` unchanged.

**Repro:** `GET /analytics/api/top-videos?agent_id=<any>&metric=garbage&limit=1` returns 200 with `"metric": "garbage"` while the rows were sorted by views. The response contract claims a metric that was never applied.

**Expected:** 400 with the allowed values, or echo the metric actually used.

**Suggested fix:** validate `metric` against `{views, engagement, tips}` and return 400 otherwise; add a regression test for the invalid value. Distinct from #1475 / PR #1476, which covered `limit` validation on the same route.

Verified against `main` on 2026-08-28.

ravendevhub:
### Description
Validates the `metric` query parameter on `/analytics/api/top-videos` to ensure it only accepts `{'views', 'engagement', 'tips'}`. Returns a 400 Bad Request with an informative error message if an invalid/unsupported metric is supplied, preventing misleading `"metric": "<invalid>"` responses.

Fixes #1790

### Changes:
1. **`analytics_blueprint.py`**:
   - Added validation check: `if metric not in ('views', 'engagement', 'tips'): return jsonify({"error": "Invalid metric, must be one of: views, engagement, tips"}), 400`
   - Cleaned up redundant fallback branch in the order clause construction.
2. **`tests/test_analytics_dashboard.py`**:
   - Added `test_analytics_api_top_videos_rejects_invalid_metric` regression test asserting 400 status code and informative error message for unsupported metric parameters.

### DIFF
--- a/analytics_blueprint.py
+++ b/analytics_blueprint.py
@@ -274,6 +274,9 @@ def api_top_videos():
         return jsonify({"error": "agent_id required"}), 400
     
     metric = request.args.get('metric', 'views')
+    if metric not in ('views', 'engagement', 'tips'):
+        return jsonify({"error": "Invalid metric, must be one of: views, engagement, tips"}), 400
+
     try:
         limit = int(request.args.get('limit', '10'))
     except (TypeError, ValueError):
@@ -288,10 +291,8 @@ def api_top_videos():
         order_clause = "ORDER BY view_count DESC"
     elif metric == 'engagement':
         order_clause = "ORDER BY (comments_count * 5 + votes_count * 2) DESC"
-    elif metric == 'tips':
+    else:  # tips
         order_clause = "ORDER BY tips_total DESC"
-    else:
-        order_clause = "ORDER BY view_count DESC"
     
     query = f"""SELECT 
             v.video_id,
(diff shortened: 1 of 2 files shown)


Scottcjn (commented):
Read the diff (`pulls/1792/files`) and ran `tests/test_analytics_dashboard.py -k top_videos` on the merge ref: the new `rejects_invalid_metric` test passes; the two failures in that class (`includes_thumbnail`, `trend_calculation`) fail identically on `main`, so nothing here regresses. The change is a 3-value allowlist on `metric` returning 400 instead of silently sorting by views, and the dead `else` branch is gone. Small, correct, matches #1790. Recommending merge to Scott; suggested tier: micro.

Issue #1780, opened 2026-08-28
Live chat JSON validation can return 500 for malformed fields


L1nkinPark:
### Summary
The live chat REST fallback accepts JSON but several fields are parsed directly with `.strip()`, `int()`, `float()`, and `request.get_json(force=True)`. Malformed JSON values can raise server exceptions instead of returning a client-safe 400 response.

Affected surfaces:
- `POST /api/chat/<video_id>/send`
- `POST /api/chat/<video_id>/ban`
- `POST /api/chat/<video_id>/settings`

### Steps to reproduce
From current `main`, send malformed but syntactically valid JSON to the live chat endpoints, for example:

1. `POST /api/chat/video-1/send` with `{"message": "hello", "tip_amount": "oops"}`
2. `POST /api/chat/video-1/send` with `{"message": 42}`
3. As a moderator, `POST /api/chat/video-1/ban` with `{"user_id": "victim", "duration": "later"}`
4. As a moderator, `POST /api/chat/video-1/settings` with `{"slow_mode": "fast"}`

### Expected
Malformed client input should receive a JSON 400 response and should not insert or update chat records.

### Actual
The handlers parse those values directly, so invalid types can raise `ValueError`, `TypeError`, or `AttributeError` and surface as a 500/error response path instead of a stable validation error.

### Environment
Local Flask test client against current BoTTube `main`, Windows desktop / PowerShell, Python test environment via `uv`.

### Notes
I found this by source audit and reproduced it with regression tests. I will open a small PR with validation helpers and coverage for the affected routes.

L1nkinPark:
Implementation update: the fix PR now also covers the SocketIO live chat path (`chat_message`, `super_chat`, and `mod_action`) because those handlers had the same malformed numeric/event payload failure mode. This stays within the same live-chat validation bug; no separate bounty claim is being made for the WebSocket portion.

Additional validation: `tests/test_chat_handlers_validation.py tests/test_websocket_server_validation.py` -> 25 passed, plus `py_compile` and `git diff --check`.

yaegerbomb42:
/claim #1780

PR: https://github.com/Scottcjn/bottube/pull/1804

### Deliverables:
- Added strict JSON type validation for `message`, `tip_amount`, `duration`, `slow_mode`, `sub_only`, and `premiere` fields in `chat_handlers.py`.
- Replaced unhandled 500 error paths with clean HTTP 400 validation responses.
[...]

PR #1823, opened 2026-08-29
fix(chat): validate JSON types on send, ban, and settings routes

Vyacheslav-Tomashevskiy:
## Summary
- validate `is_super`, `tip_amount`, moderator `duration`, and chat settings flags before they reach SQLite or float/int coercion
- reject malformed JSON objects on ban/settings routes instead of relying on `force=True` and Python coercion errors
- add regression coverage for REST chat handlers and Socket.IO chat/moderation paths

## Testing
- `pytest -q tests/test_chat_handlers_validation.py tests/test_websocket_server_validation.py`

/claim #1780

### DIFF
--- a/chat_handlers.py
+++ b/chat_handlers.py
@@ -5,6 +5,7 @@ Real-time chat for video playback, premieres, super chat, and moderation.
 Uses Flask + Flask-SocketIO, SQLite via get_db(), session/g.user patterns.
 """
 from flask import Blueprint, render_template, request, jsonify, g, session
+import math
 import sqlite3
 import time
 import uuid as _uuid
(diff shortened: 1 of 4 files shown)


Scottcjn (approved):
Verified locally, not just from the description.

- Real 3-way merge into current `main` touches only the 4 intended files (chat_handlers.py, websocket_server.py, and the two test files). The `wrtc_bridge.py`/`test_wrtc_debit_toctou_guard.py` churn in the GitHub merge-preview is a stale-base artifact, not part of this change.
- `_coerce_chat_flag` / `_coerce_non_negative_number` reject bool-as-int before the int check and gate NaN/inf via `math.isfinite` — the REST and SocketIO paths validate symmetrically.
- Tests are substantive: inf tip is rejected with zero DB insert, non-numeric duration and out-of-range flags return 400. 10/10 pass on the merged tree; CI (lint/security/test) green.

Clean input-validation hardening. Merging.

PR #1826, opened 2026-08-29
fix: unify DB path resolution across server and blueprints (#1733)

### LINKED ISSUE #1733: [BUG] Crypto bridges read BOTTUBE_DB while their tables are created under BOTTUBE_DB_PATH — different database files in 3 of 4 configurations

rafaio1:
## Summary

Fixes #1733 — silent ledger split where `init_*_tables` creates tables in one SQLite file but runtime `get_db()` reads from another.

### Root Cause
`bottube_server.py` resolves the DB path from `BOTTUBE_DB_PATH` (defaulting to `BASE_DIR/bottube.db`), while four crypto-bridge blueprints plus `banano_payout.py`, `bottube_engage.py`, and `syndication_backfill.py` read `BOTTUBE_DB` with a hardcoded `/root/bottube/bottube.db` default. In 3 of 4 configurations these resolve to different files.

### Fix
Add `bottube_db.resolve_db_path()` with priority:
1. `BOTTUBE_DB_PATH` (canonical, matches server init)
2. `BOTTUBE_DB` (legacy alias for existing deployments)
3. `BOTTUBE_BASE_DIR/bottube.db`
4. Module directory fallback

All affected modules now import this resolver instead of calling `os.environ.get()` directly.

### Testing
- Syntax validated on all modified files
- Existing test suite sets both env vars to the same temp path
- Manual matrix verification matches the four cases in #1733

RTC Wallet: RTC1e9bf7a2a60aac9bcbc5a0df0c65e9501e932861

### DIFF
--- a/banano_payout.py
+++ b/banano_payout.py
@@ -22,7 +22,8 @@ import urllib.request
 # Configuration
 # ---------------------------------------------------------------------------
 
-DB_PATH = os.environ.get("BOTTUBE_DB", "/root/bottube/bottube.db")
+from bottube_db import resolve_db_path
+DB_PATH = resolve_db_path()
 KALIUM_API = "https://kaliumapi.appditto.com/api"
 KALIUM_FALLBACK = "https://public.node.jungletv.live/rpc"
 BANANO_SEED = os.environ.get("BANANO_SEED", "")
--- a/bottube_db.py
+++ b/bottube_db.py
@@ -1,19 +1,32 @@
-# SPDX-License-Identifier: MIT
-from pathlib import Path
+"""Canonical resolver for the Bottube SQLite database path.
+
+Resolves the mismatch between BOTTUBE_DB_PATH (used by bottube_server.py
+for table creation) and BOTTUBE_DB (used historically by blueprints).
+
+Priority:
+1. BOTTUBE_DB_PATH  (canonical, matches server init)
+2. BOTTUBE_DB       (legacy alias — keeps existing deployments working)
+3. BOTTUBE_BASE_DIR / "bottube.db"
+4. Directory of this file / "bottube.db"
+"""
+from __future__ import annotations
+
 import os
+from pathlib import Path
 
 
 def resolve_db_path() -> str:
-    """Resolve the BoTTube SQLite path consistently across server and blueprints.
+    """Return the canonical SQLite database path as a string."""
+    path = os.environ.get("BOTTUBE_DB_PATH")
+    if path:
+        return path
 
-    Precedence:
-    1. BOTTUBE_DB_PATH (canonical name used by the main server)
-    2. BOTTUBE_DB (legacy alias kept for backward compatibility)
-    3. BOTTUBE_BASE_DIR/bottube.db
-    4. repo-local bottube.db next to this file
-    """
-    explicit = os.environ.get("BOTTUBE_DB_PATH") or os.environ.get("BOTTUBE_DB")
-    if explicit:
-        return explicit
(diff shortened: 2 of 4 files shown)


rafaio1:
Closing: PR #1737 by @Vyacheslav-Tomashevskiy was already merged on 2026-08-24 fixing this exact issue. This PR is redundant.

PR #1832, opened 2026-09-01
[TRAVIS] Reject malformed agent notification pagination

### LINKED ISSUE #1831: Agent notification API silently accepts malformed pagination
## Summary

`GET /api/agents/me/notifications` does not validate its `page` and `per_page` query parameters consistently with the other paginated APIs.

Werkzeug's typed query lookup returns the supplied default when integer conversion fails, so malformed values such as `page=abc` and `per_page=abc` are silently accepted with HTTP 200. Zero and negative values are also silently coerced to 1, while `per_page` values above 50 are silently clipped. Clients therefore receive a successful response for invalid input and cannot distinguish their request from a valid defaulted request.

## Reproduction

Call the authenticated endpoint with any of these query strings:

- `page=abc`
- `page=0`
- `page=-1`
- `per_page=abc`
- `per_page=0`
- `per_page=51`

For example:

```http
GET /api/agents/me/notifications?page=abc
X-API-Key: <valid key>
```

## Actual behavior

The endpoint returns HTTP 200 after silently defaulting, coercing, or clipping the invalid value.

## Expected behavior

Return a JSON HTTP 400 response for malformed and out-of-range values, using the same shared integer-query validation contract as the other paginated APIs. Bound `page` as well so pagination work cannot grow without a defined limit.

## Environment

Reproduced from current `main` at commit `9839d4e` using Python 3.14 and the Flask test client.

I have a focused patch and regression test matrix ready.

nwinkelman2:
## Summary

- validate `page` and `per_page` on the authenticated agent-notification endpoint with the shared integer-query validator
- return JSON HTTP 400 responses for malformed, zero, negative, and over-bound values instead of silently defaulting, coercing, or clipping them
- cap `page` at 10,000 so deep pagination has a defined bound
- add a focused regression matrix covering all invalid-input classes

Fixes #1831

## Verification

- `C:\Python314\python.exe -m pytest tests/test_notifications.py::test_agent_notifications_reject_malformed_pagination -q --tb=short` — 7 passed
- `C:\Python314\python.exe -m pytest tests/test_notifications.py -q --tb=short` — 9 passed
- `C:\Python314\python.exe -m py_compile bottube_server.py` — passed
- `git diff --check` — passed

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -10818,8 +10818,12 @@ def subscription_feed():
 @require_api_key
 def my_notifications():
     """List notifications for the authenticated agent."""
-    page = max(1, request.args.get("page", 1, type=int))
-    per_page = min(50, max(1, request.args.get("per_page", 20, type=int)))
+    page, error = _parse_positive_int_query("page", 1, max_value=10000)
+    if error:
+        return error
+    per_page, error = _parse_positive_int_query("per_page", 20, max_value=50)
+    if error:
+        return error
     db = get_db()
     unread_only = request.args.get("unread", "").lower() in ("1", "true", "yes")
     notifications, total = _notification_page(db, int(g.agent["id"]), page, per_page, unread_only)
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):
Validation fix with regression test (sampled set reviewed); value tier 1.

PR #1834, opened 2026-09-01
[TRAVIS] Validate earnings pagination queries

### LINKED ISSUE #1833: Earnings API silently accepts malformed pagination
## Summary

Authenticated `GET /api/agents/me/earnings` silently accepts malformed and out-of-range pagination parameters. This is inconsistent with the strict query-validation contract used by other paginated APIs and makes client errors indistinguishable from valid requests.

## Reproduction

With a valid API key, request any of:

- `page=abc`
- `page=0`
- `page=-1`
- `per_page=abc`
- `per_page=0`
- `per_page=101`

For example:

```http
GET /api/agents/me/earnings?page=abc
X-API-Key: <valid key>
```

## Actual behavior

The endpoint returns HTTP 200 after silently defaulting malformed integers, coercing non-positive values, or clipping an oversized page size.

A regression test against current `main` fails with:

```text
AssertionError: page=abc
assert 200 == 400
```

## Expected behavior

Return a structured JSON HTTP 400 response for malformed and out-of-range values. Bound `page` to a documented maximum so deep pagination work has a defined limit.

## Suggested fix

Reuse `_parse_positive_int_query` for both parameters, consistent with other endpoints, and add an authenticated invalid-input regression matrix.

## Environment

Current `main`, Python 3.14, Flask test client on Windows.

nwinkelman2:
## Summary

- validate authenticated earnings `page` and `per_page` query parameters with the shared strict validator
- return structured HTTP 400 responses instead of silently defaulting, coercing, or clipping invalid values
- bound deep pagination at 10,000 pages
- add an authenticated regression matrix for malformed, non-positive, and over-bound values

Fixes #1833

## Verification

- mutation baseline on unmodified `main`: failed as expected (`page=abc` returned 200 instead of 400)
- `C:\Python314\python.exe -m pytest tests/test_earnings_pagination_validation.py -q --tb=short` — 1 passed (7 invalid-input cases)
- `C:\Python314\python.exe -m py_compile bottube_server.py tests/test_earnings_pagination_validation.py` — passed
- `git diff --check` — passed

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -11658,8 +11658,12 @@ def manage_wallet_web():
 def my_earnings():
     """Get your RTC balance and earnings history."""
     db = get_db()
-    page = max(1, request.args.get("page", 1, type=int))
-    per_page = min(100, max(1, request.args.get("per_page", 50, type=int)))
+    page, error = _parse_positive_int_query("page", 1, max_value=10000)
+    if error:
+        return error
+    per_page, error = _parse_positive_int_query("per_page", 50, max_value=100)
+    if error:
+        return error
     offset = (page - 1) * per_page
 
     rows = db.execute(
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1836, opened 2026-09-01
[TRAVIS] Validate admin duplicate cleanup window

### LINKED ISSUE #1835: Malformed cleanup window broadens admin deletion to all history
## Summary

The authenticated admin duplicate-comment cleanup endpoint silently interprets a malformed `window_h` query as `0`, and `0` means “all history.” This can broaden an operator's intended time-bounded cleanup without any warning.

## Reproduction

With a valid admin key:

```http
GET /api/admin/duplicate-comments?window_h=abc
X-Admin-Key: <valid admin key>
```

The same behavior applies to an execution request when `dry_run=0` is supplied.

## Actual behavior

Werkzeug's typed query lookup falls back to the route default (`0`) when conversion fails. The endpoint returns HTTP 200 and runs the all-history query because its time predicate is only added when `window_h > 0`.

A mutation baseline against current `main` fails with:

```text
AssertionError: window_h=abc
assert 200 == 400
```

A negative value also selects the all-history path. Very large values are accepted without a defined bound.

## Impact

In the default dry-run mode, the response masks the typo and reports a wider-than-requested result set. With `dry_run=0`, the same malformed or negative parameter can expand a bounded duplicate cleanup into deletion across the full comment history.

## Expected behavior

After authenticating the caller, reject malformed, negative, and over-bound `window_h` values with structured HTTP 400 responses. Preserve `0` as the explicitly supported “all history” value and preserve valid bounded cleanup.

## Suggested fix

Use the shared strict integer-query validator with `min_value=0` and a documented maximum, then cover invalid input, a valid bounded dry run, and unauthenticated denial.

## Environment

Current `main`, Python 3.14, Flask test client on Windows. No production cleanup was executed.

nwinkelman2:
## Summary

- strictly validate the admin duplicate-comment cleanup `window_h` parameter after authorization
- preserve explicit `window_h=0` as all history while rejecting malformed, negative, and over-bound values
- prevent a typo from silently widening a bounded cleanup into an all-history query/deletion
- cover invalid input, valid bounded dry-run behavior, and unauthenticated denial

Fixes #1835

## Verification

- mutation baseline on unmodified `main`: failed as expected (`window_h=abc` returned 200 instead of 400)
- `C:\Python314\python.exe -m pytest tests/test_admin_duplicate_comment_validation.py -q --tb=short` — 1 passed
- `C:\Python314\python.exe -m py_compile bottube_server.py tests/test_admin_duplicate_comment_validation.py` — passed
- `git diff --check` — passed

No production cleanup was executed.

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -15083,7 +15083,11 @@ def admin_duplicate_comments():
         return err
 
     dry_run = request.args.get("dry_run", "1") != "0"
-    window_h = request.args.get("window_h", 0, type=int)
+    window_h, error = _parse_positive_int_query(
+        "window_h", 0, min_value=0, max_value=24 * 365,
+    )
+    if error:
+        return error
 
     db = get_db()
 
--- a/tests/test_admin_duplicate_comment_validation.py
+++ b/tests/test_admin_duplicate_comment_validation.py
@@ -1,0 +1,24 @@
+# SPDX-License-Identifier: MIT
+
+
+def test_duplicate_comment_cleanup_rejects_invalid_window(client, app):
+    admin_key = app.view_functions["admin_duplicate_comments"].__globals__["ADMIN_KEY"]
+
+    for query in ("window_h=abc", "window_h=-1", "window_h=8761"):
+        response = client.get(
+            f"/api/admin/duplicate-comments?{query}",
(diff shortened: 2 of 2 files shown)


Scottcjn (approved):

PR #1838, opened 2026-09-01
[TRAVIS] Validate video-tip pagination queries

### LINKED ISSUE #1837: Video tips API silently accepts malformed pagination
## Summary

`GET /api/videos/<video_id>/tips` silently accepts malformed and out-of-range `page` / `per_page` values. The earlier overflow fix in #1461 correctly rejects pages whose computed SQLite offset exceeds 64-bit bounds, but ordinary malformed/non-positive values still bypass validation by defaulting or coercion, and oversized `per_page` values are clipped.

## Reproduction

For any existing public video:

```http
GET /api/videos/<video_id>/tips?page=abc
GET /api/videos/<video_id>/tips?page=0
GET /api/videos/<video_id>/tips?per_page=abc
GET /api/videos/<video_id>/tips?per_page=51
```

## Actual behavior

Each request returns HTTP 200 after silently defaulting, coercing, or clipping the invalid value.

A mutation baseline against current `main` fails at the first case:

```text
AssertionError: page=abc
assert 200 == 400
```

## Expected behavior

Return a structured JSON HTTP 400 response for malformed, non-positive, and over-bound values, while preserving:

- valid ordinary pages
- the existing 64-bit SQLite overflow rejection
- the existing large-but-safe page behavior

## Suggested fix

Use the shared strict integer-query validator for both parameters, then retain the existing computed-offset overflow guard.

## Verification scope

A focused regression matrix should cover six invalid classes plus the four existing normal/large/overflow cases from #1461.

## Environment

Current `main`, Python 3.14, Flask test client on Windows. No production data was modified.

nwinkelman2:
## Summary

- strictly validate public video-tip `page` and `per_page` query parameters
- return structured HTTP 400 responses instead of silently defaulting, coercing, or clipping invalid values
- preserve the existing large-but-safe page behavior and SQLite offset-overflow guard from #1461
- extend the existing overflow regression suite with six malformed/range cases

Fixes #1837

## Verification

- mutation baseline on unmodified `main`: failed as expected (`page=abc` returned 200 instead of 400)
- `C:\Python314\python.exe -m pytest tests/test_video_tips_page_offset_overflow.py -q --tb=short` — 5 passed
- `C:\Python314\python.exe -m py_compile bottube_server.py tests/test_video_tips_page_offset_overflow.py` — passed
- `git diff --check` — passed

No production data was modified.

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -12100,8 +12100,12 @@ def get_video_tips(video_id):
     if not v:
         return jsonify({"error": "Video not found"}), 404
     _sync_pending_tips(db)
-    page = max(1, request.args.get("page", 1, type=int))
-    per_page = min(50, max(1, request.args.get("per_page", 10, type=int)))
+    page, error = _parse_positive_int_query("page", 1)
+    if error:
+        return error
+    per_page, error = _parse_positive_int_query("per_page", 10, max_value=50)
+    if error:
+        return error
     offset = (page - 1) * per_page
     # An astronomically large ?page makes offset exceed SQLite's signed 64-bit
     # INTEGER range, which raises OperationalError on "LIMIT ? OFFSET ?" and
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1840, opened 2026-09-01
[TRAVIS] Validate admin comment cleanup JSON

### LINKED ISSUE #1839: String false enables destructive admin comment cleanup
## Summary

The authenticated admin comment-cleanup endpoint coerces JSON values with Python truthiness instead of validating their types. Most importantly, a client that sends the common string form `"force_remove": "false"` selects destructive removal mode because every non-empty string is truthy.

## Reproduction

With a valid admin key:

```http
POST /api/admin/comment-cleanup
Content-Type: application/json
X-Admin-Key: <valid admin key>

{"force_remove":"false","remove_dupes":false,"max_similar":0}
```

## Actual behavior

The endpoint returns HTTP 200 with:

```json
{"mode":"force_remove", ...}
```

If duplicate/excess comments exist, the handler executes its deletion branches. Additional ambiguous inputs are also accepted or fail uncleanly:

- `force_remove: 1` selects destructive mode
- `remove_dupes: "false"` is treated as true
- `max_similar: true` is treated as integer 1
- `max_similar: "3"` reaches an invalid int/string comparison
- a non-object JSON body reaches `.get()` on the wrong type

A mutation baseline against current `main` failed at the string-false case (`200 != 400`).

## Impact

An operator/client serialization mismatch can turn an explicitly false removal request into destructive deletion. Malformed thresholds can also cause server errors or silently disable/alter cleanup behavior.

## Expected behavior

After authorizing the caller and before any database work:

- require a JSON object
- require real JSON booleans for `remove_dupes` and `force_remove`
- require a non-boolean integer in a bounded range for `max_similar`
- return structured HTTP 400 responses for invalid input
- preserve the valid default coach-and-hold path

## Environment

Current `main`, Python 3.14, Flask test client on an isolated temporary database. No production cleanup was executed.

nwinkelman2:
## Summary

- require an object JSON body before admin comment-cleanup processing
- require actual JSON booleans for `remove_dupes` and `force_remove` so string `"false"` cannot enable deletion
- require a non-boolean integer from 0 through 100 for `max_similar`
- preserve valid coach-and-hold behavior and authorization ordering
- add invalid-type, valid-path, and unauthorized regression coverage

Fixes #1839

## Verification

- mutation baseline on unmodified `main`: failed as expected (`"force_remove":"false"` returned 200 in `force_remove` mode)
- `C:\Python314\python.exe -m pytest tests/test_admin_comment_cleanup_json_validation.py -q --tb=short` — 1 passed (8 invalid payloads + valid + unauthorized)
- `C:\Python314\python.exe -m pytest tests/test_quests.py::test_comment_cleanup_defaults_to_hold_without_deleting -q --tb=short` — 1 passed
- `C:\Python314\python.exe -m py_compile bottube_server.py tests/test_admin_comment_cleanup_json_validation.py` — passed
- `git diff --check` — passed

No production cleanup was executed.

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -15161,10 +15161,25 @@ def admin_comment_cleanup():
     if err:
         return err
 
-    data = request.get_json(silent=True) or {}
+    data, error = _json_object_body()
+    if error:
+        return error
+
     remove_dupes = data.get("remove_dupes", True)
+    if not isinstance(remove_dupes, bool):
+        return jsonify({"error": "remove_dupes must be a boolean"}), 400
+
+    force_remove = data.get("force_remove", False)
+    if not isinstance(force_remove, bool):
+        return jsonify({"error": "force_remove must be a boolean"}), 400
+
     max_similar = data.get("max_similar", 3)
-    force_remove = bool(data.get("force_remove", False))
+    if (
+        isinstance(max_similar, bool)
+        or not isinstance(max_similar, int)
+        or not 0 <= max_similar <= 100
+    ):
+        return jsonify({"error": "max_similar must be an integer between 0 and 100"}), 400
 
     db = get_db()
     held_dupes = 0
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):
Validation fix with regression test; value tier 2.

PR #1842, opened 2026-09-01
[TRAVIS] Validate bulk notification read JSON

### LINKED ISSUE #1841: String false marks every notification read
## Summary

Both notification bulk-read endpoints use Python truthiness for the JSON `all` field. A client that sends the common string form `"all": "false"` therefore marks every notification read because every non-empty string is truthy.

Affected routes:

- API-key route: `POST /api/agents/me/notifications/read`
- browser/session route: `POST /api/notifications/read` (and its `/web-read` alias)

## Reproduction

With two unread notifications and valid authentication:

```json
{"all":"false"}
```

## Actual behavior

The endpoint returns HTTP 200 and updates all unread rows. A mutation baseline against current `main` failed at the first route (`200 != 400`), and the unread count dropped because `bool("false") is True`.

The `ids` selection is also weakly typed: non-list inputs can be iterated unexpectedly, strings are coerced to integers, and invalid entries are silently ignored.

## Impact

A serialization mismatch can erase the user's unread state across the entire inbox even though the request explicitly says false. Because the API-key and browser handlers duplicated the same coercion, both interfaces are affected.

## Expected behavior

Before updating rows:

- require an actual JSON boolean for `all`
- require `ids` to be an array of positive JSON integers
- reject simultaneous `all: true` and non-empty `ids`
- return structured HTTP 400 responses on invalid selection
- keep API-key and browser behavior identical
- preserve authentication/CSRF behavior

## Environment

Current `main`, Python 3.14, Flask test client on an isolated temporary database. No production notification state was modified.

nwinkelman2:
## Summary

- validate bulk notification read selection for both API-key and browser/session routes
- require a real JSON boolean for `all`, so string `"false"` cannot mark the entire inbox read
- require `ids` to be an array of positive integers and reject conflicting `all` + `ids`
- share one validation contract across both interfaces
- prove invalid requests leave unread state unchanged while preserving valid read flows, auth, and CSRF

Fixes #1841

## Verification

- mutation baseline on unmodified `main`: failed as expected (`{"all":"false"}` returned 200 and selected mark-all)
- `C:\Python314\python.exe -m pytest tests/test_notifications.py -q --tb=short` — 3 passed
- `C:\Python314\python.exe -m pytest tests/test_authenticated_json_object_validation.py -q --tb=short` — 2 passed
- `C:\Python314\python.exe -m py_compile bottube_server.py tests/test_notifications.py` — passed
- `git diff --check` — passed

Running the two test files in one pytest process also passed every assertion but hit an unrelated existing Windows temporary-DB teardown lock; each file passes cleanly in isolation as shown above. No production notification state was modified.

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -4111,6 +4111,25 @@ def _mark_notification_rows_read(db, agent_id: int, notification_ids=None, mark_
     return int(cur.rowcount or 0)
 
 
+def _notification_read_options(data):
+    """Validate notification read selection without truthy coercion."""
+    mark_all = data.get("all", False)
+    if not isinstance(mark_all, bool):
+        return None, None, (jsonify({"error": "all must be a boolean"}), 400)
+
+    notification_ids = data.get("ids", [])
+    if not isinstance(notification_ids, list):
+        return None, None, (jsonify({"error": "ids must be an array of positive integers"}), 400)
+    if any(
+        isinstance(value, bool) or not isinstance(value, int) or value <= 0
+        for value in notification_ids
+    ):
+        return None, None, (jsonify({"error": "ids must be an array of positive integers"}), 400)
+    if mark_all and notification_ids:
+        return None, None, (jsonify({"error": "all and ids are mutually exclusive"}), 400)
+    return notification_ids, mark_all, None
+
+
 def _canonical_webhook_event(event: str) -> str:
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):
Validation fix with regression test; value tier 2.

PR #1846, opened 2026-09-01
[TRAVIS] Guard HTML search offset overflow

### LINKED ISSUE #1845: HTML search page crashes on oversized pagination offset
## Reproduction

Against current `main` in the repository Flask test client:

```text
GET /search?q=video&page=9223372036854775807
```

The HTML search route calculates `(page - 1) * 24` and binds that value as a SQLite `OFFSET`. This request raises `OverflowError: Python int too large to convert to SQLite INTEGER`, which becomes HTTP 500 outside testing mode.

The JSON search endpoint already guards the equivalent SQLite offset boundary, but the user-facing `/search` page does not.

## Expected

Return a bounded HTTP 400 response for a page whose offset exceeds SQLite's signed 64-bit integer range. Large values that remain within the database range should continue to render normally.

## Scope

This is an ordinary pagination correctness bug. It was reproduced only against an isolated temporary test database; no production data or state was modified.

nwinkelman2:
## Summary

- calculate the user-facing search offset once and reject values above SQLite's signed 64-bit integer range
- return a clear HTTP 400 response instead of allowing an `OverflowError` to surface as HTTP 500
- preserve large-but-safe pagination and normal search-page rendering
- add focused boundary regression coverage

Fixes #1845

## Verification

- mutation baseline on unmodified `main`: oversized page raised `OverflowError: Python int too large to convert to SQLite INTEGER` (1 failed, 1 passed)
- `C:\Python314\python.exe -m pytest -q tests/test_search_html_page_offset_overflow.py tests/test_discoverability.py::TestSearchPageUI::test_search_page_renders` — 3 passed
- `C:\Python314\python.exe -m py_compile bottube_server.py tests/test_search_html_page_offset_overflow.py` — passed
- `git diff --check` — passed

The mutation and regression tests used the repository's isolated temporary database fixture. No production data was modified.

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -14026,6 +14026,9 @@ def search_page():
     except (TypeError, ValueError):
         page = 1
     per_page = 24
+    offset = (page - 1) * per_page
+    if offset > _SQLITE_MAX_INT64:
+        return jsonify({"error": "page out of range"}), 400
     # Category checkboxes submit repeated ?cat=<id> params.
     valid_cat_ids = {c["id"] for c in VIDEO_CATEGORIES}
     selected_categories = [c for c in request.args.getlist("cat") if c in valid_cat_ids]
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1848, opened 2026-09-01
[TRAVIS] Guard agent-directory offset overflow

### LINKED ISSUE #1847: Agent directory crashes when pagination offset exceeds SQLite range
## Reproduction

Against current `main` with the repository Flask test client and its temporary database:

```text
GET /api/agents?page=9223372036854775807&limit=100
```

The directory parses both values successfully, calculates `(page - 1) * limit`, and binds the result to SQLite as an `OFFSET`. The bind raises `OverflowError: Python int too large to convert to SQLite INTEGER`, which becomes HTTP 500 outside test mode.

Existing directory validation correctly handles malformed input but does not cover the database integer boundary.

## Expected

Return HTTP 400 when the calculated offset exceeds SQLite's signed 64-bit integer range. Large pages whose offsets remain within range should keep returning normal directory responses.

## Scope

This is an ordinary pagination correctness bug. It was reproduced only against an isolated temporary database; no production state was accessed or modified.

nwinkelman2:
## Summary

- guard the public agent-directory `OFFSET` against values above SQLite's signed 64-bit integer range
- return a structured HTTP 400 response instead of exposing an `OverflowError` as HTTP 500
- preserve existing malformed-input validation and large-but-safe pages
- add real-database boundary regression coverage

Fixes #1847

## Verification

- mutation baseline on unmodified `main`: oversized page raised `OverflowError: Python int too large to convert to SQLite INTEGER` (1 failed, 1 passed)
- `C:\Python314\python.exe -m pytest -q tests/test_agent_directory_offset_overflow.py tests/test_agent_directory_query_validation.py` — 14 passed
- `C:\Python314\python.exe -m py_compile agent_discovery.py tests/test_agent_directory_offset_overflow.py` — passed
- `git diff --check` — passed

All tests used isolated fake or temporary databases. No production data was modified.

## Payout

### DIFF
--- a/agent_discovery.py
+++ b/agent_discovery.py
@@ -22,6 +22,8 @@ from flask import Blueprint, Response, current_app, jsonify, request
 
 discovery_bp = Blueprint("agent_discovery", __name__)
 
+_SQLITE_MAX_INT64 = (1 << 63) - 1
+
 
 def _parse_agent_directory_int(name: str, default: int, min_value: int, max_value: int | None = None) -> int:
     """Parse an integer query parameter with bounds checking.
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1850, opened 2026-09-01
[TRAVIS] Validate personal-list pagination

### LINKED ISSUE #1849: Personal-list APIs silently accept invalid and unbounded pagination
## Reproduction

Three authenticated personal-list APIs apply inconsistent pagination contracts:

- `GET /api/feed/subscriptions`
- `GET /api/messages/inbox`
- `GET /api/history`

On current `main`, `page=10001` returns HTTP 200 on all three even though the public feed is already bounded at 10,000 pages. The messages and history endpoints additionally return HTTP 200 for malformed, zero, and over-limit values such as `page=abc`, `page=0`, `per_page=abc`, `per_page=0`, and `per_page=51`; Werkzeug silently supplies defaults or the route clips values.

Because each route binds `(page - 1) * per_page` to SQLite as an `OFFSET`, an arbitrarily large page can also exceed SQLite's signed 64-bit integer range and surface an `OverflowError`/HTTP 500.

## Expected

All three personal-list APIs should use the existing strict positive-integer query parser, reject pages above 10,000 and page sizes above 50, and return structured HTTP 400 errors for malformed or out-of-range input. The documented boundary values should remain valid.

## Evidence

An isolated temporary-database regression matrix produced 13 assertion failures on unmodified `main`: messages/history silently accepted malformed and clipped values, and all three endpoints accepted an over-bound page. No production state was accessed or modified.

nwinkelman2:
## Summary

- align subscription feed, message inbox, and watch history on the shared strict pagination parser
- reject malformed/nonpositive query values instead of silently defaulting or clipping them
- enforce the established 10,000-page and 50-item boundaries before SQLite offset calculation
- preserve both documented upper boundary values and existing personal-list behavior
- add one cohesive 24-case invalid/boundary regression matrix across all three APIs

Fixes #1849

## Verification

- mutation baseline on unmodified `main`: 13 failures / 8 passes before adding the explicit oversized-page case; messages/history silently accepted malformed/clipped values and all three accepted page 10,001
- `C:\Python314\python.exe -m pytest -q tests/test_personal_list_pagination_validation.py` — 2 test functions passed (24 route/query assertions)
- focused existing history/message/subscription regressions — 2 passed, 6 deselected
- `C:\Python314\python.exe -m py_compile bottube_server.py tests/test_personal_list_pagination_validation.py` — passed
- `git diff --check` — passed

The first parameterized proof run passed all assertions after the fix but encountered one unrelated existing Windows temporary-database teardown lock; the compact two-fixture version above passes cleanly. No production state was modified.

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -10765,7 +10765,7 @@ def agent_subscribers(agent_name):
 @require_api_key
 def subscription_feed():
     """Get videos from agents you follow, newest first."""
-    page, error = _parse_positive_int_query("page", 1)
+    page, error = _parse_positive_int_query("page", 1, max_value=10000)
     if error:
         return error
     per_page, error = _parse_positive_int_query("per_page", 20, max_value=50)
@@ -17053,8 +17053,12 @@ def message_inbox():
 
     Query params: page, per_page, unread_only (0/1)
     """
-    page = max(1, request.args.get("page", 1, type=int))
-    per_page = min(50, max(1, request.args.get("per_page", 20, type=int)))
+    page, error = _parse_positive_int_query("page", 1, max_value=10000)
+    if error:
+        return error
+    per_page, error = _parse_positive_int_query("per_page", 20, max_value=50)
+    if error:
+        return error
     unread_only = request.args.get("unread_only", "0") == "1"
     offset = (page - 1) * per_page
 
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1852, opened 2026-09-01
[TRAVIS] Validate playlist field contracts

### LINKED ISSUE #1851: Playlist APIs serialize structured JSON into text fields
## Reproduction

Playlist create/update accept structured JSON values where the API contract expects strings:

```json
POST /api/playlists
{"title":{"nested":"title"}}
```

Current `main` returns HTTP 201 and persists the Python representation `{'nested': 'title'}` as the playlist title. `description` behaves the same way. Invalid visibility values are silently replaced with `public` on create, or ignored while PATCH still returns `{ "ok": true }`.

PATCH has the same title/description coercion, so a malformed client request can mutate an existing playlist to serialized JSON-looking text rather than being rejected.

## Expected

- `title` and `description` accept JSON strings only.
- A supplied title must remain nonempty.
- `visibility` accepts only the documented `public`, `unlisted`, or `private` strings.
- Invalid PATCH requests return HTTP 400 and leave the playlist unchanged.
- Valid create/update behavior remains unchanged.

## Evidence

On unmodified `main`, isolated regression tests failed both the create and update contracts; create returned HTTP 201 with the structured title serialized into text, and PATCH returned HTTP 200 after accepting the same mutation. Tests use a temporary database only.

nwinkelman2:
## Summary

- validate playlist title/description as strings instead of serializing lists or objects into stored text
- validate visibility as exactly one of the documented enum strings instead of silently defaulting/ignoring invalid values
- require a nonempty title on both create and supplied-title updates
- reject malformed PATCH requests before mutation and preserve valid create/update behavior
- add invalid-create, no-mutation, and valid-update regression coverage

Fixes #1851

## Verification

- mutation baseline on unmodified `main`: both new tests failed; structured title returned HTTP 201 and PATCH returned HTTP 200 after accepting it
- `C:\Python314\python.exe -m pytest -q tests/test_playlist_field_validation.py` — 2 passed (4 invalid create cases, 4 invalid/no-mutation patch cases, valid update)
- `C:\Python314\python.exe -m pytest -q tests/test_playlist_field_validation.py tests/test_authenticated_json_object_validation.py::test_playlist_routes_reject_non_object_json` — 3 passed before the final valid-update assertion was added
- `C:\Python314\python.exe -m py_compile bottube_server.py tests/test_playlist_field_validation.py` — passed
- `git diff --check` — passed

All mutation checks used the repository temporary-database fixture. No production data was modified.

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -10936,6 +10936,28 @@ def web_mark_single_notification_read(notification_id: int):
 # Playlists (API + Web)
 # ---------------------------------------------------------------------------
 
+
+def _playlist_text_field(data, field, *, default="", max_length, required=False):
+    value = data.get(field, default)
+    if value is None:
+        value = default
+    if not isinstance(value, str):
+        return None, f"{field} must be a string"
+    value = value.strip()[:max_length]
+    if required and not value:
+        return None, f"{field} is required"
+    return value, None
+
+
+def _playlist_visibility(data, *, default="public"):
+    value = data.get("visibility", default)
+    if not isinstance(value, str):
+        return None, "visibility must be a string"
+    if value not in ("public", "unlisted", "private"):
+        return None, "visibility must be one of: public, unlisted, private"
+    return value, None
+
+
 @app.route("/api/playlists", methods=["POST"])
 @require_api_key
 def api_create_playlist():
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1854, opened 2026-09-01
[TRAVIS] Validate webhook registration fields

### LINKED ISSUE #1853: Webhook registration crashes on structured events JSON
## Reproduction

Authenticated webhook registration accepts an object JSON body but does not validate the types of `url` and `events` before using string operations:

```json
POST /api/webhooks
{"url":"https://example.com/hook","events":{"name":"video.uploaded"}}
```

Current `main` calls `.split()` on the object and raises `AttributeError`, surfacing as HTTP 500 outside test mode. A mixed array such as `["video.uploaded", 7]` fails in `join()` with `TypeError`. The URL field is also converted with `str()`, so structured values are formatted before URL validation rather than rejected by contract.

## Expected

- `url` must be a JSON string and then satisfy the existing HTTPS check.
- `events` must be a string or an array containing strings only.
- Invalid requests return structured HTTP 400 responses before any webhook row is inserted.
- Both documented valid forms remain accepted.

## Evidence

On unmodified `main`, an isolated regression raised `AttributeError: 'dict' object has no attribute 'split'` from the real route. The valid string and string-array controls passed. No production network callback or state was used.

nwinkelman2:
## Summary

- require webhook URL input to be a JSON string before applying the existing HTTPS check
- require events to be a string or an array containing strings only
- return structured HTTP 400 responses instead of raising from `split()` or `join()`
- prove every invalid request leaves the webhook table unchanged
- preserve both documented valid event forms

Fixes #1853

## Verification

- mutation baseline on unmodified `main`: object-valued events raised `AttributeError: 'dict' object has no attribute 'split'`; valid controls passed
- `C:\Python314\python.exe -m pytest -q tests/test_webhook_field_validation.py tests/test_authenticated_json_object_validation.py::test_authenticated_utility_routes_reject_non_object_json` — 3 passed
- `C:\Python314\python.exe -m py_compile bottube_server.py tests/test_webhook_field_validation.py` — passed
- `git diff --check` — passed

Tests used an isolated temporary database and example HTTPS URLs only. No outbound callback or production state was used.

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -11402,13 +11402,20 @@ def create_webhook():
     data, error = _json_object_body()
     if error:
         return error
-    url = str(data.get("url", "")).strip()
+    url = data.get("url", "")
+    if not isinstance(url, str):
+        return jsonify({"error": "url must be a string"}), 400
+    url = url.strip()
     if not url or not url.startswith("https://"):
         return jsonify({"error": "url must be a valid HTTPS URL"}), 400
 
     events = data.get("events", "*")
     if isinstance(events, list):
+        if not all(isinstance(event, str) for event in events):
+            return jsonify({"error": "events must contain only strings"}), 400
         events = ",".join(events)
+    elif not isinstance(events, str):
+        return jsonify({"error": "events must be a string or array of strings"}), 400
     # Validate event names
     for ev in events.split(","):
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1856, opened 2026-09-01
[TRAVIS] Validate wallet field types

### LINKED ISSUE #1855: Wallet API serializes structured JSON into payout fields
## Reproduction

The authenticated wallet update route converts every supplied payout field with `str()` before persistence:

```json
POST /api/agents/me/wallet
{"btc":{"nested":"address"}}
```

Current `main` returns HTTP 200 and writes `{'nested': 'address'}` into `agents.btc_address`. Arrays, booleans, and `null` are similarly serialized for the legacy RTC, BTC, ETH, SOL, LTC, ERG, and PayPal fields. The dedicated on-chain `rtc_wallet` field happens to reject most structured values only because their serialized representation fails its later address-format check.

This is a data-integrity failure on payout configuration: malformed JSON should not become address text that is later displayed or consumed as if the user supplied it.

## Expected

- Every supplied wallet field must be a JSON string before trimming, format checks, and persistence.
- Invalid field types return HTTP 400 before any wallet column changes.
- Empty strings remain the supported way to clear a field.
- Valid partial string updates and the existing RTC activation behavior remain unchanged.

## Evidence

On unmodified `main`, the isolated regression returned HTTP 200 and persisted the structured legacy RTC field. Tests use a temporary database and placeholder strings only; no wallet transaction, payout, or production state is involved.

nwinkelman2:
## Summary

- require every supplied wallet/payout field to be a JSON string before trimming or persistence
- prevent objects, arrays, booleans, and null from being serialized into address columns
- validate all supplied fields before building the update so invalid requests leave wallet state unchanged
- preserve valid partial string updates, empty-string clearing, RTC address validation, and referral activation
- add all-field invalid/no-mutation and valid partial-update regressions

Fixes #1855

## Verification

- mutation baseline on unmodified `main`: structured legacy RTC field returned HTTP 200 and crossed the persistence path
- `C:\Python314\python.exe -m pytest -q tests/test_wallet_field_type_validation.py tests/test_authenticated_json_object_validation.py::test_authenticated_utility_routes_reject_non_object_json` — 3 passed
- `C:\Python314\python.exe -m pytest -q tests/test_referrals.py::test_referral_dashboard_tracks_human_and_agent_funnels` — 1 passed
- `C:\Python314\python.exe -m py_compile bottube_server.py tests/test_wallet_field_type_validation.py` — passed
- `git diff --check` — passed

Tests used a temporary database and placeholder strings only. No real wallet transaction, payout, or production state was involved.

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -11581,8 +11581,12 @@ def manage_wallet():
         "paypal": "paypal_email",
     }
 
+    for key in allowed_fields:
+        if key in data and not isinstance(data[key], str):
+            return jsonify({"error": f"{key} must be a string"}), 400
+
     if "rtc_wallet" in data:
-        rtc_wallet = str(data.get("rtc_wallet", "")).strip()
+        rtc_wallet = data["rtc_wallet"].strip()
         if rtc_wallet and not _is_rustchain_rtc_address(rtc_wallet):
             return jsonify({"error": "Invalid RustChain wallet address format (expected RTC... address)"}), 400
 
@@ -11590,7 +11594,7 @@ def manage_wallet():
     params = []
     for key, col in allowed_fields.items():
         if key in data:
-            val = str(data[key]).strip()
+            val = data[key].strip()
             updates.append(f"{col} = ?")
             params.append(val)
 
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1858, opened 2026-09-01
[TRAVIS] Validate notification preference booleans

### LINKED ISSUE #1857: Notification preference string false enables email
## Reproduction

The authenticated notification-preferences API converts values with Python truthiness instead of requiring JSON booleans:

```json
PUT /api/notifications/preferences
{"comments":"false"}
```

Current `main` returns HTTP 200 with `{"updated":{"comments":true}}` and stores the preference as enabled. Arrays, objects, nonzero numbers, and other truthy JSON values have the same mismatch.

This makes a syntactically plausible client value do the opposite of what it says, potentially causing email delivery a user attempted to disable.

## Expected

- Each supplied preference accepts JSON `true` or `false` only.
- Invalid values return HTTP 400 before any preference column changes.
- Valid partial boolean updates retain the existing response shape and behavior.

## Evidence

On unmodified `main`, the isolated mutation regression returned HTTP 200 and changed `comments` from disabled to enabled for the string `"false"`. No email was sent and no production state was used.

nwinkelman2:
## Summary

- require exact JSON booleans for every supplied email-notification preference
- reject string, numeric, object, and array values before any database update
- prevent string `"false"` from enabling a preference through Python truthiness
- preserve valid partial boolean updates and the existing response contract
- add all-preference no-mutation and valid partial-update regression coverage

Fixes #1857

## Verification

- mutation baseline on unmodified `main`: `{"comments":"false"}` returned HTTP 200 and changed the preference to true
- `C:\Python314\python.exe -m pytest -q tests/test_notification_preference_boolean_validation.py tests/test_authenticated_json_object_validation.py::test_authenticated_utility_routes_reject_non_object_json` — 3 passed
- `C:\Python314\python.exe -m py_compile bottube_server.py tests/test_notification_preference_boolean_validation.py` — passed
- `git diff --check` — passed

Tests used an isolated temporary database. No email was sent and no production state was modified.

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -14571,6 +14571,9 @@ def api_set_notification_preferences():
         "tips": "email_notify_tips",
         "subscriptions": "email_notify_subscriptions",
     }
+    for key in allowed:
+        if key in data and not isinstance(data[key], bool):
+            return jsonify({"error": f"{key} must be a boolean"}), 400
     updated = {}
     for key, col in allowed.items():
         if key in data:
--- a/tests/test_notification_preference_boolean_validation.py
+++ b/tests/test_notification_preference_boolean_validation.py
@@ -1,0 +1,36 @@
+"""Regression coverage for notification preference boolean contracts."""
+
+
+def test_notification_preferences_reject_non_booleans_without_mutation(client, registered_agent):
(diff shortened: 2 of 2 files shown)


Scottcjn (approved):

PR #1860, opened 2026-09-01
[TRAVIS] Validate notification read selectors before mutation

### LINKED ISSUE #1859: Notification read routes treat string false as mark-all
Both authenticated notification read routes coerce the JSON all selector with Python truthiness. A client payload with all set to the string false therefore marks every unread notification as read, while scalar or object ids selectors may also be iterated and mutate unintended rows. The routes should accept all only as a JSON boolean and ids only as an array of positive integer IDs, reject malformed selectors with HTTP 400, and leave unread state unchanged. I have a focused regression and shared validation fix ready.

nwinkelman2:
Closes #1859

## Problem
Both API-key and session notification read routes coerced the `all` selector with Python truthiness. JSON such as `{"all":"false"}` therefore marked every unread notification as read. Scalar/object `ids` selectors could likewise be iterated and mutate unintended rows.

## Fix
- share one strict selector parser across both routes
- require `all` to be a JSON boolean
- require `ids` to be an array of positive integer IDs (excluding booleans)
- return HTTP 400 before any database mutation for malformed selectors

## Proof
- mutation baseline: 7 new selector cases failed against unmodified `main`
- focused suite: `11 passed`
- `git diff --check`

## Payout

### DIFF
--- a/tests/test_notifications.py
+++ b/tests/test_notifications.py
@@ -270,6 +270,34 @@ def test_notification_read_routes_reject_string_false_without_marking_all(client
         db = bottube_server.get_db()
         assert bottube_server._notification_unread_count(db, alice_id) == 2
 
+
+@pytest.mark.parametrize(
+    "payload",
+    [
+        {"all": 1},
+        {"ids": "1"},
+        {"ids": {"1": True}},
+        {"ids": [True]},
+        {"all": True, "ids": [1]},
+    ],
+)
+def test_notification_read_routes_reject_other_malformed_selectors(client, payload):
+    alice_id = _insert_agent("selector-shapes", "bottube_sk_selector_shapes")
+    _insert_notification(alice_id, "subscribe", "new subscriber")
+
+    response = client.post(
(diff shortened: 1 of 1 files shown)


Scottcjn (dismissed):

Scottcjn:
Merging your TRAVIS set in order, @nwinkelman2 — 20 are on `main` so far. This one now conflicts on bottube_server.py with a sibling that landed. Please rebase onto current `main` and it goes in on the next pass. Awards are value-tiered (1-4 RTC by non-test size) and paid after the set lands. — Sophia

nwinkelman2:
Rebased onto current main at 5c99cd4. The overlapping production validation has already landed through the sibling series, so I preserved the stricter current implementation and retained this PR's non-duplicate malformed-selector coverage (numeric all, non-array ids, boolean ids, and mutually exclusive all+ids). The branch is conflict-free and ready for the next merge pass.

Scottcjn (approved):

PR #1862, opened 2026-09-01
[TRAVIS] Reject earnings pagination beyond SQLite range

### LINKED ISSUE #1861: Oversized earnings page crashes SQLite OFFSET binding
The authenticated earnings ledger computes OFFSET directly from unbounded page and per_page query values. A page at SQLite signed-integer scale produces a Python integer too large to bind, raising OverflowError and returning HTTP 500 instead of a client error. The route should reject only offsets beyond SQLite signed 64-bit range with HTTP 400 while preserving the largest safe page and existing pagination behavior. I have a focused mutation regression and bounded fix ready.

nwinkelman2:
Closes #1861

## Problem
`/api/agents/me/earnings` multiplied an unbounded page by `per_page` and bound that result directly as SQLite `OFFSET`. At signed-64-bit page scale this raises `OverflowError` and returns HTTP 500.

## Fix
Reject only calculated offsets beyond SQLite signed-integer range with a clear HTTP 400. The largest safe page and all normal pagination behavior remain unchanged.

## Proof
- mutation baseline: oversized-page regression raised `OverflowError` on unmodified `main`; safe-boundary case passed
- after fix: `2 passed`
- `git diff --check`

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -11661,6 +11661,8 @@ def my_earnings():
     page = max(1, request.args.get("page", 1, type=int))
     per_page = min(100, max(1, request.args.get("per_page", 50, type=int)))
     offset = (page - 1) * per_page
+    if offset > _SQLITE_MAX_SIGNED_INT:
+        return jsonify({"error": "page out of range"}), 400
 
     rows = db.execute(
         """SELECT amount, reason, video_id, created_at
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1864, opened 2026-09-01
[TRAVIS] Validate video metadata field types before update

### LINKED ISSUE #1863: Video metadata PATCH crashes or serializes structured JSON fields
The owner-only video metadata PATCH route calls strip directly on title and description and on every tags-list item, so structured JSON values raise AttributeError and return HTTP 500. A mapping supplied as tags is silently stringified and persisted as invalid metadata. The route should require title and description to be strings and tags to be either a string or an array containing only strings, reject malformed fields before any update, and preserve supported updates. I have four mutation regressions plus a valid-update control and a bounded fix ready.

nwinkelman2:
Closes #1863

## Problem
The video metadata PATCH route called `.strip()` directly on `title`, `description`, and each list item in `tags`. Structured JSON produced HTTP 500, while a tags mapping was silently stringified and stored as corrupt metadata.

## Fix
- require a JSON object body
- require title/description to be strings when supplied
- allow tags only as a string or an array containing only strings
- reject malformed fields before any database mutation
- preserve trimming and supported text/tag-array updates

## Proof
- mutation baseline: 4 malformed-field cases failed on unmodified `main` (three exceptions plus one invalid mutation); valid update control passed
- after fix: `5 passed`
- `git diff --check`

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -6980,10 +6980,23 @@ def update_video(video_id):
     if row['agent_id'] != g.agent['id']:
         return jsonify({'error': 'Not your video'}), 403
     
-    data = request.get_json(silent=True) or {}
+    data, error = _json_object_body()
+    if error:
+        return error
     updates = []
     params = []
-    
+
+    if 'title' in data and not isinstance(data['title'], str):
+        return jsonify({'error': 'title must be a string'}), 400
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1876, opened 2026-09-01
[TRAVIS] Validate mood history limit values

### LINKED ISSUE #1875: Mood history silently accepts malformed and out-of-range limit values
## Summary
The public GET /api/v1/agents/{agent_name}/mood/history route documents limit as 1 through 100, but malformed values silently fall back to 20 and values outside the range are clamped. Clients therefore receive 200 for limit=abc, limit=0, and limit=101 instead of learning that their request is invalid.

## Reproduction
After registering an agent, request its mood history with limit=abc, limit=0, or limit=101. Each currently returns 200 and calls the mood engine.

## Expected
Reject malformed and out-of-range values with a clear 400 before calling the engine. Preserve the default and documented 1 through 100 range.

## Scope
Ordinary public query validation only; no authentication or security-control changes.

nwinkelman2:
Closes #1875.

## What changed
- route mood-history limit through the shared positive-integer query parser
- reject malformed, zero, negative, and values above the documented maximum of 100
- preserve the default limit and valid upper bound

## Mutation proof
Before the fix, limit=abc, limit=0, and limit=101 all returned 200 and reached the mood engine. The focused regression suite failed all three rejection cases while its valid-limit control passed.

## Validation
- C:\Python314\python.exe -m pytest -q tests/test_mood_history_limit_validation.py -> 4 passed
- C:\Python314\python.exe -m py_compile bottube_server.py
- git diff --check

This is ordinary public query validation and does not change authentication or security controls.

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -7523,7 +7523,9 @@ def get_mood_history(agent_name):
     if not agent:
         return jsonify({"error": "Agent not found"}), 404
     
-    limit = min(100, max(1, request.args.get("limit", 20, type=int)))
+    limit, error = _parse_positive_int_query("limit", 20, max_value=100)
+    if error:
+        return error
     
     engine = get_mood_engine(str(DB_PATH))
     history = engine.get_mood_history(agent["id"], limit)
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1886, opened 2026-09-01
[TRAVIS] Restore related-videos response contract

### LINKED ISSUE #1885: Related videos endpoint no longer returns its documented response contract
## Summary
The discoverability documentation and its original integration test define GET /api/videos/{video_id}/related as returning video_id, related_videos, and count. The live handler now returns only ok and related. SDK callers receive raw JSON, so the contract drift breaks clients that follow the published response shape.

## Reproduction
Create a source video and one related candidate, then call the endpoint. The response has ok and related but is missing video_id, related_videos, and count.

## Expected
Restore the documented fields while retaining related as an additive compatibility alias for clients using the newer compact shape.

## Scope
Ordinary public API compatibility/data-shape correctness only; no authentication or security-control changes.

nwinkelman2:
Closes #1885.

## What changed
- restore documented video_id, related_videos, and count fields
- retain related as an additive compatibility alias
- build the response items once so both collection keys remain identical

## Mutation proof
A source plus one related candidate returned 200, but the contract regression failed immediately with KeyError for video_id. The focused response-contract test now passes and verifies all fields and exact count.

## Validation
- C:\Python314\python.exe -m pytest -q tests/test_related_limit_validation.py::test_related_response_preserves_documented_contract -> 1 passed
- the full file completed 7 passing assertions; Windows then hit an unrelated temporary SQLite handle during fixture cleanup
- C:\Python314\python.exe -m py_compile bottube_server.py
- git diff --check

This is ordinary public API compatibility/data-shape correctness and does not change authentication or security controls.

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -17333,9 +17333,7 @@ def api_related_videos(video_id):
         return s
 
     scored = sorted(candidates, key=score, reverse=True)
-    return jsonify({
-        "ok": True,
-        "related": [
+    related_videos = [
             {
                 "video_id": r["video_id"],
                 "title": r["title"],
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1890, opened 2026-09-01
[TRAVIS] Wire search filters, sorting, and pagination state

### LINKED ISSUE #1889: Search filters and sort controls do not affect results
## Problem
The public `/search` page renders category and sort controls, but their state is not wired to the result query:

- category checkboxes submit repeated `category` parameters while the route reads only legacy `cat`, so selecting a category does not filter anything;
- the sort form submits `sort`, but the route always orders by views and never passes the selected value back to the template;
- pagination links preserve only the query and page, silently dropping category and sort selections.

This makes the visible controls misleading and causes users to lose their search state after navigating pages.

## Proposed fix
- Accept the template's repeated `category` parameters while retaining `cat` compatibility.
- Apply a whitelisted order expression for all five displayed sort modes.
- Persist the selected sort when category filters change.
- Preserve category and sort state in previous/next links.
- Add regression coverage for filtering, ordering, pagination state, and the legacy alias.

nwinkelman2:
## Summary
- wire the public search category controls to the route's category filter while preserving legacy `cat` URLs
- apply a whitelisted sort expression for every sort mode shown in the UI
- retain category/sort state when changing filters or navigating pagination
- add end-to-end regression coverage for result filtering, ordering, pagination state, and compatibility

## Reproduction / mutation proof
Before the fix, `GET /search?q=Needle&category=music&sort=likes` still rendered a science-tech result ahead of music results because the category and sort parameters were both ignored. The new regression failed on that inclusion before the implementation.

## Validation
- `C:\Python314\python.exe -m pytest -q tests/test_search_filter_sort_state.py` -> 3 passed
- `C:\Python314\python.exe -m py_compile bottube_server.py tests/test_search_filter_sort_state.py`
- `git diff --check`

Closes #1889

## Payout

### DIFF
--- a/bottube_templates/search.html
+++ b/bottube_templates/search.html
@@ -169,6 +169,7 @@
             <div class="filter-title">Categories</div>
             <form id="category-filter-form" method="GET" action="{{ P }}/search">
                 <input type="hidden" name="q" value="{{ query }}">
+                <input type="hidden" name="sort" value="{{ sort }}">
                 {% for cat in categories %}
                 <div class="filter-option">
                     <input type="checkbox" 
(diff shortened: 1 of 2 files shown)


Scottcjn:

Scottcjn (approved):

PR #1892, opened 2026-09-01
[TRAVIS] Enforce public visibility in video listings

### LINKED ISSUE #1891: Public video list exposes videos from banned creators
## Problem
`GET /api/videos` filters removed videos but does not filter videos owned by banned creators. That diverges from the canonical public-video predicate already used by video detail, search, feeds, related videos, and other public surfaces.

Consequences:
- banned-creator videos remain discoverable through both `/api/videos` and `/api/v1/videos`;
- `total` and `pages` include those non-public rows;
- `?agent=<banned_creator>` can explicitly resurface their catalog;
- list cache validators are calculated from content that the public response should not expose.

## Proposed fix
Use the existing `_public_video_filter_sql()` predicate when building the video-list query so the row query, totals, pagination, and cache revision all share one visibility definition. Add regression coverage for the canonical route, agent filtering, and the versioned alias.

nwinkelman2:
## Summary
- apply the existing canonical public-video predicate to `GET /api/videos`
- keep banned-creator videos out of response rows, totals, pages, and cache validators
- cover direct agent filtering and the `/api/v1/videos` alias

## Reproduction / mutation proof
Before the fix, a fixture with one public creator video and one banned creator video returned `total: 2`; filtering with `?agent=banned_creator` returned that non-public video, and the versioned alias exposed it as well. All three regression tests failed against `main` before the one-line predicate change.

## Validation
- `C:\Python314\python.exe -m pytest -q tests/test_video_list_public_visibility.py` -> 3 passed
- `C:\Python314\python.exe -m pytest -q tests/test_video_list_pagination.py` -> 2 passed
- Python compilation clean
- `git diff --check`

Closes #1891

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -7160,7 +7160,7 @@ def list_videos():
     order = sort_map.get(sort, "v.created_at DESC")
 
     db = get_db()
-    where_clauses = ["v.is_removed = 0"]
+    where_clauses = [_public_video_filter_sql()]
     params = []
     if agent_name:
         where_clauses.append("a.agent_name = ?")
--- a/tests/test_video_list_public_visibility.py
+++ b/tests/test_video_list_public_visibility.py
@@ -1,0 +1,93 @@
+# SPDX-License-Identifier: MIT
+"""Public video listings must use the canonical visibility predicate."""
+
+import time
+
+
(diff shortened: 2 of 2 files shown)


Scottcjn (approved):

PR #1904, opened 2026-09-01
[TRAVIS] Restore Activity Feed initial load and polling

### LINKED ISSUE #1903: Activity Feed aborts before its first request on page load
## Summary

The Activity Feed page calls `loadFeed('all')` from `DOMContentLoaded` and later calls `loadFeed(currentFilter, true)` from `setInterval`, but `loadFeed` unconditionally reads the undeclared global `event.target` before entering its `try` block.

Browsers do not provide an implicit `event` for either programmatic call, so initial loading throws `ReferenceError: event is not defined` before `/social/api/feed` is fetched. The page remains stuck on "Loading activity feed..." and its 30-second refresh repeats the same failure.

## Reproduction

1. Open `/activity` in a browser context where `window.event` is not a legacy implicit global.
2. Let the `DOMContentLoaded` callback call `loadFeed('all')`.
3. Observe the exception at `event.target?.classList.add('active')` and no feed request.

Optional chaining does not protect an undeclared identifier; resolving `event` itself throws.

## Expected

Tab state should derive from the requested filter, because the same function is intentionally used by clicks, initial load, and polling. Programmatic calls must not depend on a click event.

## Proposed fix

Give each feed tab a `data-filter` value and activate the tab whose value matches `filter`. Add a regression that rejects implicit `event` reads inside `loadFeed` and verifies every supported filter has programmatic state.

nwinkelman2:
## What changed

- give each Activity Feed tab explicit filter state
- derive the active tab from the requested filter
- remove `loadFeed`'s dependency on the legacy implicit browser `event`
- add a regression for both initial/poll calls and programmatic tab activation

## Mutation and verification

Before the fix, the focused contract suite failed 2/2: `loadFeed` read undeclared `event.target` even though both DOMContentLoaded and polling call it without an event, and the tabs exposed no programmatic filter state.

After the fix:

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_activity_feed_event_contract.py` -> 2 passed
- `python -m py_compile tests/test_activity_feed_event_contract.py` -> clean
- `git diff --check` -> clean

Fixes #1903

## Payout

### DIFF
--- a/bottube_templates/activity_feed.html
+++ b/bottube_templates/activity_feed.html
@@ -182,10 +182,10 @@
     </div>
 
     <div class="feed-tabs">
-        <button class="feed-tab active" onclick="loadFeed('all')">All Activity</button>
-        <button class="feed-tab" onclick="loadFeed('uploads')">📤 Uploads</button>
-        <button class="feed-tab" onclick="loadFeed('comments')">💬 Comments</button>
-        <button class="feed-tab" onclick="loadFeed('tips')">💰 Tips</button>
+        <button class="feed-tab active" data-filter="all" onclick="loadFeed('all')">All Activity</button>
+        <button class="feed-tab" data-filter="uploads" onclick="loadFeed('uploads')">📤 Uploads</button>
+        <button class="feed-tab" data-filter="comments" onclick="loadFeed('comments')">💬 Comments</button>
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1906, opened 2026-09-01
[TRAVIS] Personalize Discover from the signed-in session

### LINKED ISSUE #1905: Discover For You tab always falls back to generic trending for signed-in users
## Summary

The Discover page's visible **For You** tab cannot use the signed-in human's viewing history.

The browser reads `agent_id` from localStorage (nothing in the app ever writes that key), appends it as a query parameter, and makes an ordinary same-origin request. The backend intentionally ignores raw `agent_id` values and only resolves `X-API-Key`, so a signed-in web user is treated as anonymous and receives the generic trending feed every time.

## Reproduction

1. Sign in through the BoTTube web UI.
2. Record viewing history for the signed-in user.
3. Open `/discover` and choose **For You**.
4. Observe `/discover/api/for-you` returns the trending response without `personalized`/`based_on`, rather than recommendations based on that user's history.

The same failure reproduces directly with a valid signed-in `g.user` and no agent API-key header.

## Expected

- Web UI: resolve the existing signed-in session user.
- Agent/API client: continue resolving `X-API-Key`.
- Anonymous request: fall back to trending.
- Never trust a raw `agent_id` query parameter.

## Proposed fix

Use `g.user['id']` when a web session is present, otherwise preserve API-key lookup. Make the Discover page issue a same-origin credentialed request instead of reading an identity from localStorage. Add regressions proving a signed-in viewer gets personalized, already-viewed-excluding results.

nwinkelman2:
## What changed

- resolve the Discover recommendation viewer from the signed-in web session first
- preserve API-key identity for agent clients
- preserve anonymous fallback to trending
- replace the unused/localStorage agent-id query with a same-origin credentialed request
- add a two-video history regression proving personalization and viewed-video exclusion

## Mutation and verification

Before the fix, both new regressions failed: a valid web session received the generic trending schema, and the UI tried to source identity from an unwritten localStorage key.

After the fix:

- focused and adjacent Discover suites: 17 passed
- `python -m py_compile search_blueprint.py tests/test_discover_for_you_session.py` -> clean
- `git diff --check` -> clean

Fixes #1905

## Payout

### DIFF
--- a/bottube_templates/discover.html
+++ b/bottube_templates/discover.html
@@ -366,11 +366,7 @@
 
     async function loadForYou() {
         try {
-            // Get agent_id from session/storage if logged in
-            const agentId = localStorage.getItem('agent_id');
-            const url = agentId ? `/discover/api/for-you?agent_id=${agentId}` : '/discover/api/for-you';
-            
-            const response = await fetch(url);
+            const response = await fetch('/discover/api/for-you', { credentials: 'same-origin' });
             const data = await response.json();
             
             const container = document.getElementById('foryouVideos');
--- a/search_blueprint.py
+++ b/search_blueprint.py
@@ -401,18 +401,24 @@ def api_trending():
 def api_for_you():
     """
     Personalized "For You" feed based on viewing history.
-    Authenticate via X-API-Key header to get personalized results.
-    Falls back to trending if no key provided.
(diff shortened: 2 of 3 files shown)


Scottcjn (approved):

PR #1908, opened 2026-09-01
[TRAVIS] Activate creator-directory pagination

### LINKED ISSUE #1907: Discover creator directory ignores its documented offset pagination
## Summary

`GET /discover/api/agents` documents `offset` pagination, but the handler never reads, validates, applies, or returns that parameter.

Every page request therefore returns the first `limit` creators. For example, `?sort=videos&limit=1&offset=1` returns the same top creator as `offset=0`, and malformed/negative offsets silently return 200.

## Reproduction

1. Create three creators with 3, 2, and 1 videos.
2. Request `/discover/api/agents?sort=videos&limit=1&offset=0`.
3. Request `/discover/api/agents?sort=videos&limit=1&offset=1`.

Current result: both responses contain the 3-video creator. `offset=bad` and `offset=-1` are also ignored.

## Expected

- Parse `offset` as a non-negative integer.
- Apply `LIMIT ? OFFSET ?` to the ranked directory query.
- Return the effective `limit` and `offset` with the result page.
- Reject malformed and negative offsets consistently with the other Discover APIs.

## Impact

The creator directory cannot paginate beyond its first page, so every creator outside the first result window is unreachable through the visible discovery surface.

nwinkelman2:
## What changed

- parse and validate the documented creator-directory `offset`
- apply it to the sorted SQL page with `LIMIT ? OFFSET ?`
- return effective pagination values in the response
- add ordered multi-page and invalid-offset regressions

## Mutation and verification

Before the fix, all 3 focused regressions failed: `offset=1` still returned page zero, while malformed and negative offsets returned 200.

After the fix:

- focused plus adjacent Discover/query suites: 20 passed
- `python -m py_compile search_blueprint.py tests/test_discover_agent_directory_pagination.py` -> clean
- `git diff --check` -> clean

Fixes #1907

## Payout

### DIFF
--- a/search_blueprint.py
+++ b/search_blueprint.py
@@ -558,6 +558,7 @@ def api_agent_directory():
     sort = request.args.get('sort', 'subscribers')
     try:
         limit = _parse_int_query('limit', 20, min_val=1, max_val=50)
+        offset = _parse_int_query('offset', 0, min_val=0)
     except ValueError as exc:
         return jsonify({"error": str(exc)}), 400
 
@@ -592,7 +593,7 @@ def api_agent_directory():
         ) v ON v.agent_id = a.id
         WHERE video_count > 0
         ORDER BY {order_sql}
-        LIMIT ?""", (limit,)).fetchall()
+        LIMIT ? OFFSET ?""", (limit, offset)).fetchall()
     
     results = []
     for row in agents:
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1914, opened 2026-09-01
[TRAVIS] Activate the translation runtime end to end

### LINKED ISSUE #1913: Translation API is unregistered and has no production schema
## Reproduction

On current `main`, `translation_routes.py` defines six translation routes, but the production app never imports or registers `translation_bp`:

```console
python -c "import bottube_server; print([r.rule for r in bottube_server.app.url_map.iter_rules() if 'translation' in r.rule])"
[]
```

The production schema also never creates `video_translations` (the only table definition is in a unit-test fixture). As a result, every documented translation endpoint is a 404; registering the blueprint alone would only turn those requests into `no such table: video_translations` failures.

The route contract has a third mismatch: `videos.video_id` is the canonical public `TEXT` identifier, but both translation GET routes use Flask's `<int:video_id>` converter and the test-only table defines an integer foreign key.

## Expected

- translation routes are registered in the production app
- bootstrap creates the translation table idempotently
- translation APIs use the canonical public video identifier
- reads/writes preserve public-content visibility and reject nonexistent videos
- focused regression coverage proves registration, schema initialization, and the public-ID workflow

nwinkelman2:
Fixes #1913

## What changed
- registers the existing translation blueprint in the production Flask app and initializes its durable SQLite store at the same activation edge
- uses canonical public `videos.video_id` text identifiers throughout instead of internal integer row IDs
- replaces the nonexistent `require_auth` dependency with BoTTube's real API-key guard
- rejects invalid or nonexistent video writes and keeps removed videos / banned creators off public translation reads
- adds idempotent schema/index initialization and public-ID round-trip regressions

## Proof
- production mutation baseline: importing current `main` exposed **zero** translation routes
- `tests/test_translation_route_validation.py`: **4 passed**
- `tests/test_translation_runtime_registration.py`: **1 passed** (real production app import, isolated database, 52.55s)
- existing `tests/test_translations.py`: **18 passed**
- changed files compile cleanly; `git diff --check` passes

This closes the complete runtime gap: registration, authentication, schema, canonical identifiers, read visibility, and tests are activated together.

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -15546,6 +15546,15 @@ app.register_blueprint(store_bp)
 
 # USDC Payment Integration (Base Chain)
 from bottube_db import resolve_db_path
+
+# Translation API and UI. Keep its table initializer beside registration so a
+# route cannot be advertised without the storage contract it needs.
+from translation_routes import init_translation_tables, translation_bp
+_translation_db = sqlite3.connect(resolve_db_path())
+init_translation_tables(_translation_db)
+_translation_db.close()
+app.register_blueprint(translation_bp)
+
 from usdc_blueprint import usdc_bp, init_usdc_tables
 import sqlite3 as _usdc_sqlite3
 _usdc_db = _usdc_sqlite3.connect(resolve_db_path())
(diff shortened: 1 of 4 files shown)


Scottcjn (approved):

PR #1916, opened 2026-09-01
fix: validate provenance anchor batch membership

### LINKED ISSUE #1915: Provenance anchor callback ignores claimed batch membership
## Summary
`POST /api/admin/provenance/anchor-result` receives the worker's `video_ids` but never validates or uses them. A stale or mismatched success callback therefore marks every row currently carrying the supplied `batch_id` as anchored even when the callback's Merkle root was computed from a different member set.

## Reproduction
1. Insert two claimed provenance rows under one `anchor_batch_id`.
2. POST a successful callback for that batch with `video_ids` containing one real member and one unrelated ID.
3. The endpoint returns 200 and anchors both claimed rows.
4. A success callback with no `video_ids` is also accepted.

A focused regression baseline currently reports 2 failures / 2 passes: mismatched and missing membership are accepted, while exact membership and failure callbacks work.

## Expected
Successful callbacks must provide a non-empty, unique string list whose membership exactly equals the rows claimed under `batch_id`; mismatch must fail without mutating the batch. Failure callbacks may omit membership because the worker intentionally uses that path to release a failed claim.

## Scope
Ordinary provenance data-consistency validation only. No authentication or security-control changes.

nwinkelman2:
## Summary
- require successful provenance anchor callbacks to carry a non-empty, unique `video_ids` list
- compare the callback member set to the rows actually claimed under `batch_id` before idempotency or mutation
- reject stale/mismatched callbacks with 409 while preserving the claimed rows
- preserve the existing failure-release path, which intentionally omits `video_ids`
- validate callback JSON/text fields instead of allowing malformed bodies to raise attribute errors

## Proof
Mutation baseline on upstream `main`:
- **2 failed, 2 passed**: mismatched membership and missing membership were both accepted and anchored rows

After fix:
- `python -m pytest -q tests/test_provenance_anchor_membership.py` -> **4 passed**
- `python -m pytest -q tests/test_provenance_anchor_membership.py tests/test_receipt_watch_url.py tests/test_admin_key_env.py` -> **8 passed**
- `python -m py_compile bottube_server.py tests/test_provenance_anchor_membership.py` -> clean
- `git diff --check` -> clean

Closes #1915

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -23399,17 +23399,53 @@ def admin_provenance_anchor_result():
     if not _ts_admin_ok():
         return jsonify({"ok": False, "error": "unauthorized"}), 401
     _provenance_ensure_anchor_columns()
-    data = request.get_json(silent=True) or {}
-    batch_id = (data.get("batch_id") or "").strip()
-    chain = (data.get("chain") or "ergo").strip()[:32]
-    tx_hash = (data.get("tx_hash") or "").strip()[:128]
+    data, body_error = _admin_json_body()
+    if body_error:
+        return jsonify({"ok": False, "error": body_error}), 400
+
+    batch_id, field_error = _admin_text_field(data, "batch_id")
+    if field_error:
+        return jsonify({"ok": False, "error": field_error}), 400
+    chain, field_error = _admin_text_field(data, "chain", default="ergo", max_length=32)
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1918, opened 2026-09-01
perf: remove discarded profile aggregates

### LINKED ISSUE #1917: Agent profile runs three discarded social aggregate queries
## Summary
`GET /api/agents/<agent_name>` executes three aggregate social-graph queries (incoming commenters, incoming likers, and outgoing interactions) but never reads or returns any of their results. The same data already has a dedicated `/api/agents/<agent_name>/interactions` endpoint.

Every profile request therefore pays for two grouped joins plus a correlated-subquery scan whose results are discarded. The cost grows with comments, votes, videos, and agents while the response is byte-for-byte unaffected.

## Reproduction
A focused SQL-trace regression test seeds three agents and representative interactions, requests `/api/agents/alice`, and observes the discarded `GROUP BY ... LIMIT 8` and `ORDER BY comments_given + likes_given` statements. Baseline: 1 failed.

## Expected
The profile endpoint should execute only queries needed to construct its response. Social aggregates should remain available from the dedicated interactions endpoint.

## Scope
Ordinary performance/correctness cleanup only; no response schema, auth, or security behavior changes.

nwinkelman2:
## Summary
- remove three social aggregate queries from `GET /api/agents/<agent_name>` whose results were assigned and then discarded
- leave the dedicated `/api/agents/<agent_name>/interactions` endpoint unchanged
- add a SQL-trace regression proving the profile route does not reintroduce the discarded grouped/correlated scans

## Proof
- mutation baseline on upstream main: focused SQL-trace test **failed**
- after fix: profile query + both social graph shape tests **3 passed**
- Python compile clean
- `git diff --check` clean

The full `test_social_graph.py` file also contains six pre-existing expectation failures because current error responses include the documented `param` field; this PR neither changes nor relies on those validation responses.

Closes #1917

## Payout

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -8740,40 +8740,6 @@ def get_agent(agent_name):
     )
     agent_badges = _list_agent_badges(db, int(agent["id"]))
 
-    # Agent-to-agent interaction data
-    aid = agent["id"]
-    interaction_commenters = db.execute(
-        """SELECT a2.agent_name, a2.display_name, a2.avatar_url, COUNT(*) AS cnt
-           FROM comments c JOIN videos v ON c.video_id = v.video_id
-           JOIN agents a2 ON c.agent_id = a2.id
-           WHERE v.agent_id = ? AND c.agent_id != ?
-             AND COALESCE(v.is_removed, 0) = 0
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1930, opened 2026-09-01
[TRAVIS] Make BAN upload rewards idempotent

### LINKED ISSUE #1929: Upload completion retries can credit duplicate BAN rewards
## Summary
`award_ban_upload` is not idempotent. Replaying the successful-upload delivery step for the same agent and public video ID inserts another credited `video_upload` reward every time, inflating the custodial BAN ledger.

## Reproduction
On current `main`:

1. Initialize the Banano tables in SQLite.
2. Call `award_ban_upload(db, 17, "video-public-id")` twice.
3. Query credited reward rows for agent 17 / `video-public-id` / reason `video_upload`.

## Expected
Exactly one row totaling `REWARDS["upload"]`; retries return no additional reward. A different video ID remains independently rewardable.

## Actual
Two identical credited rows are written, totaling twice the configured upload reward. The function has no duplicate check, unique key, or atomic insert guard.

This matters because upload completion is an effect edge that can be retried after response loss/recovery. The adjacent generation-reward path already tries to prevent duplicate delivery; upload rewards do not.

## Environment
- Windows 11
- Python 3.14
- SQLite in-memory reproduction
- Repository `main` at clone time, 2026-09-01

## Non-duplicate check
Searched open and closed BoTTube issues for duplicate Banano/BAN upload rewards and found no matching report.

nwinkelman2:
## Summary
- make BAN upload reward delivery idempotent per agent/public-video pair
- use one atomic `INSERT ... WHERE NOT EXISTS` effect instead of a read/write split
- return the delivered amount on first execution and `0.0` on replay
- prove different videos remain independently rewardable

Closes #1929.

## Proof
On unmodified `main`, the duplicate-delivery regression failed with two identical 1 BAN rows.

With this patch:
- focused + adjacent Banano amount/tip/withdraw suites: **27 passed**
- `python -m py_compile banano_blueprint.py tests/test_banano_upload_reward_idempotency.py` -> clean
- `git diff --check` -> clean

The single SQLite insert statement keeps concurrent writers serialized at the ledger effect instead of relying on a race-prone pre-read.

## Payout

### DIFF
--- a/banano_blueprint.py
+++ b/banano_blueprint.py
@@ -326,14 +326,17 @@ def get_platform_balance() -> dict:
 # ---------------------------------------------------------------------------
 
 def award_ban_upload(db, agent_id: int, video_id: str):
-    """Award BAN for a successful video upload."""
+    """Award BAN once for a successful agent/video upload pair."""
     amount = REWARDS["upload"]
-    db.execute(
+    cursor = db.execute(
         "INSERT INTO ban_transactions (agent_id, tx_type, amount_ban, reason, video_id, status, created_at) "
-        "VALUES (?, 'reward', ?, 'video_upload', ?, 'credited', ?)",
-        (agent_id, amount, video_id, time.time()),
+        "SELECT ?, 'reward', ?, 'video_upload', ?, 'credited', ? "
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1936, opened 2026-09-01
fix: validate transcript trigger input

### LINKED ISSUE #1934: Transcript trigger crashes or misreads malformed force values
## Bug

`POST /api/videos/<video_id>/transcript/trigger` assumes every parsed JSON value is an object and coerces the optional `force` field with Python truthiness.

## Reproduction

- A top-level JSON array reaches `.get(...)` and raises `AttributeError`, producing an internal failure.
- `{ "force": "false" }` is coerced to `True`, so a caller trying to disable replacement can unintentionally request forced retranscription.
- Numbers, arrays, and objects are also silently converted to booleans and enqueue work.

## Expected

The route should accept an absent body or JSON object only, require `force` to be a JSON boolean when present, reject malformed types with HTTP 400, and never enqueue work for invalid input.

## Scope

Ordinary request-contract correctness for the existing transcription trigger; no authentication or security-control changes.

nwinkelman2:
## Summary
- reject non-object JSON bodies on the transcript trigger instead of raising `AttributeError`
- require `force` to be a JSON boolean instead of coercing strings, numbers, arrays, and objects
- ensure invalid requests never enqueue transcription work
- cover both malformed body shapes and five misleading `force` values

Fixes #1934

## Mutation proof
Before the fix, all six new contracts failed: top-level arrays raised `AttributeError`, while every invalid `force` value reached `enqueue_transcription` after truthiness coercion.

## Verification
- `50 passed, 10 skipped` — full `tests/test_whisper_transcription.py`
- `7 passed` — focused trigger input and adjacent not-found contracts
- `python -m py_compile whisper_transcription_blueprint.py tests/test_whisper_transcription.py`
- `git diff --check`

## Payout

### DIFF
--- a/tests/test_whisper_transcription.py
+++ b/tests/test_whisper_transcription.py
@@ -594,6 +594,49 @@ def test_trigger_transcription_video_not_found(flask_client):
     assert resp.status_code == 404
 
 
+def test_trigger_transcription_rejects_non_object_json(flask_client, monkeypatch):
+    import whisper_transcription_blueprint as wtb
+
+    monkeypatch.setattr(wtb, "_get_video_filename", lambda _video_id: "clip.mp4")
+    monkeypatch.setattr(
+        wtb.wt,
+        "enqueue_transcription",
+        lambda *_args, **_kwargs: pytest.fail("malformed body enqueued work"),
+    )
+
+    resp = flask_client.post(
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1938, opened 2026-09-01
fix: validate discover search enums

### LINKED ISSUE #1937: Discover search silently ignores invalid category and sort values
## Bug

`GET /discover/api/search` documents finite enums for `category` and `sort`, but does not validate either one.

## Reproduction

- `?category=definitely-not-a-category` skips the category predicate and queries the entire catalog while echoing the invalid category.
- `?q=robot&sort=definitely-not-a-sort` silently falls back to relevance ordering.

Both malformed requests reach the database instead of returning a client error, making filters misleading and potentially returning much broader results than requested.

## Expected

Reject unknown non-empty categories and sort modes with HTTP 400 before querying. Return the allowed values so callers can repair the request.

## Scope

Ordinary Discover request-validation and result-consistency behavior only.

nwinkelman2:
## Summary
- reject unknown non-empty Discover category filters before querying
- reject unsupported Discover sort modes instead of silently treating them as relevance
- return the allowed enum values with HTTP 400 responses
- prove invalid options never reach the database

Fixes #1937

## Mutation proof
Both focused contracts failed on upstream main: the invalid category and invalid sort each reached the database query path.

## Verification
- `24 passed` — query-validation, thumbnail, and null-bio Discover suites
- `python -m py_compile search_blueprint.py tests/test_search_blueprint_query_validation.py`
- `git diff --check`

## Payout

### DIFF
--- a/search_blueprint.py
+++ b/search_blueprint.py
@@ -102,6 +102,19 @@ def api_search():
 
     if not query and not category:
         return jsonify({"error": "Query or category required"}), 400
+
+    if category and category not in VIDEO_CATEGORIES:
+        return jsonify({
+            "error": "Invalid 'category' parameter.",
+            "allowed": VIDEO_CATEGORIES,
+        }), 400
+
+    allowed_sorts = ('relevance', 'newest', 'views', 'likes')
+    if sort not in allowed_sorts:
+        return jsonify({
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1940, opened 2026-09-01
[TRAVIS] Serialize BAN ledger debit admission

### LINKED ISSUE #1939: Concurrent BAN withdrawals can reserve more than the available balance
## Summary
BAN withdrawal admission is a check-then-write race. Two concurrent requests can both read the same spendable ledger balance before either pending withdrawal is visible, then each insert a reservation that individually passed the check but collectively exceeds the balance.

## Reproduction
A deterministic concurrent regression on current `main`:

1. Credit agent 1 with 10 BAN.
2. Start two authenticated `/ban/withdraw` requests for 8 BAN each.
3. Synchronize both requests immediately after their unlocked balance query.
4. Allow both requests to continue.

## Expected
One request returns 200 and reserves 8 BAN. The other re-reads the post-reservation balance and returns 400. Total pending withdrawals remain 8 BAN.

## Actual
Both requests return 200 and the ledger holds 16 BAN of pending withdrawals against a 10 BAN balance.

The balance query and pending insert are separate operations with no write transaction or account-level reservation.

## Environment
- Windows 11
- Python 3.14
- SQLite / Flask threaded test clients
- Repository `main` at clone time, 2026-09-01

## Non-duplicate check
Searched open and closed BoTTube issues for concurrent/racing Banano withdrawals and found no matching report.

nwinkelman2:
## Summary
- acquire a SQLite write reservation before reading spendable BAN for both tips and withdrawals
- keep each balance admission and debit insertion inside one transaction
- roll back explicitly when the post-lock balance is insufficient
- add deterministic two-request concurrency regressions for both debit routes

Closes #1939.

## Proof
On unmodified `main`, two simultaneous 8 BAN withdrawals both returned 200 and reserved 16 BAN against a 10 BAN balance. The same unlocked check/write pattern also allowed two simultaneous 8 BAN tips to spend the same 10 BAN.

With this patch:
- concurrency + adjacent Banano amount/tip/withdraw suites: **27 passed**
- `python -m py_compile banano_blueprint.py tests/test_banano_withdrawal_atomicity.py` -> clean
- `git diff --check` -> clean

## Payout

### DIFF
--- a/banano_blueprint.py
+++ b/banano_blueprint.py
@@ -553,6 +553,12 @@ def ban_tip():
     if recipient["id"] == user_id:
         return jsonify({"error": "Cannot tip yourself"}), 400
 
+    # Reserve the sender ledger before checking spendable balance.  Tips and
+    # withdrawals debit the same balance, so an unlocked check-then-insert
+    # lets concurrent requests each spend funds the other has not committed
+    # yet.
+    db.execute("BEGIN IMMEDIATE")
+
     # Withdrawals are written as 'pending' and only ever advance to
     # 'sent'/'failed' -- never 'credited'. Filtering on 'credited' alone
     # dropped those rows before the CASE could subtract them, leaving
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1942, opened 2026-09-01
fix: preserve failed generation jobs for retry

### LINKED ISSUE #1941: Canceling a failed generation job destroys its retry state
## Bug

A failed generation job remains eligible for `POST /api/generation/jobs/<job_id>/retry`, but the cancel endpoint also accepts that terminal `failed` state and overwrites it with `canceled`.

## Reproduction

1. Let an owned generation job reach `failed`.
2. POST its `/cancel` route.
3. Observe HTTP 200 and status mutation to `canceled`.
4. The retry route then rejects it because it is no longer `failed`.

## Expected

Like completed and already-canceled jobs, failed jobs are terminal for cancellation and should return HTTP 409 without mutation. This preserves the explicit retry transition.

## Scope

Ordinary workflow/state-machine correctness only; no authentication or security-control changes.

nwinkelman2:
## Summary
- treat `failed` as a terminal state for cancellation
- preserve the documented failed -> retry transition
- return the same idempotent HTTP 409 contract used for completed/canceled jobs
- prove no state mutation occurs for a failed job

Fixes #1941

## Mutation proof
On upstream main, the regression reached `update_job(... status="canceled")` and failed because the terminal job was mutated.

## Verification
- `8 passed` — full `tests/test_generation_routes.py`
- `python -m py_compile generation/routes.py tests/test_generation_routes.py`
- `git diff --check`

## Payout

### DIFF
--- a/generation/routes.py
+++ b/generation/routes.py
@@ -268,7 +268,11 @@ def cancel_generation_job(job_id: str):
         return jsonify({"error": "Job not found"}), 404
     if job.get("owner_user_id") != str(g.agent["id"]):
         return jsonify({"error": "Not your job"}), 403
-    if job["status"] in (JobStatus.completed.value, JobStatus.canceled.value):
+    if job["status"] in (
+        JobStatus.completed.value,
+        JobStatus.failed.value,
+        JobStatus.canceled.value,
+    ):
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1944, opened 2026-09-01
Return 404 for unknown relationship event histories

### LINKED ISSUE #1943: GET relationship events returns 200 for an unknown relationship
## Problem

GET /api/beef/relationships/{id}/events currently returns HTTP 200 with an empty array when the relationship ID does not exist. That response is indistinguishable from an existing relationship that has no events, so API clients cannot tell a typo/stale ID from a valid empty event history.

## Reproduction

Request an ID absent from agent_relationships. The endpoint queries only relationship_events and serializes the empty result as [].

## Expected

Return HTTP 404 with a stable relationship-not-found error when the parent relationship does not exist. Preserve HTTP 200 [] for an existing relationship whose event history is empty.

## Proposed fix

Check parent relationship existence before reading events and add a regression test for the missing-parent contract.

nwinkelman2:
Closes #1943.

## What changed
- verify the parent relationship exists before loading its event history
- return a stable HTTP 404 error for an unknown relationship ID
- retain the existing HTTP 200 event array for valid relationships
- add regression coverage for the missing-parent contract

## Mutation proof
Before the production fix, the new regression failed because the endpoint returned 200 instead of 404.

## Validation
- focused event contracts: 2 passed
- full tests/test_beef_system.py: 37 passed, 1 unrelated pre-existing failure in test_admin_kill because the generated admin key is not supplied by the test and the route returns 403
- git diff --check: clean

## Payout

### DIFF
--- a/agent_relationships.py
+++ b/agent_relationships.py
@@ -420,6 +420,13 @@ def get_events(rel_id: int):
         The result value.
     """
     db = get_db()
+    relationship = db.execute(
+        "SELECT 1 FROM agent_relationships WHERE id=?",
+        (rel_id,),
+    ).fetchone()
+    if relationship is None:
+        return jsonify({"error": "relationship not found"}), 404
+
     rows = db.execute(
         "SELECT * FROM relationship_events WHERE relationship_id=? ORDER BY created_at DESC",
         (rel_id,),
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1946, opened 2026-09-01
Keep provider failover rotation stable across restarts

### LINKED ISSUE #1945: Provider failover rotation changes after every process restart
## Problem

ProviderRegistry.get_ordered documents deterministic rotation by job ID, but it uses Python's built-in hash(). String hashes are randomized per process, so the same job ID can select a different first provider after a worker restart.

That breaks sticky retries and makes failover/load-distribution behavior depend on the process hash seed rather than the job identity.

## Reproduction

Register three healthy providers and call get_ordered for the same job under two different Python hash seeds. The provider order changes.

## Expected

The same job ID and provider registry order should produce the same rotation across process restarts.

## Proposed fix

Derive the rotation index from a stable digest of the job ID and add a regression that proves changing the process hash seed cannot affect provider selection.

nwinkelman2:
Closes #1945.

## What changed
- replace Python's process-randomized string hash with a stable SHA-256-derived rotation index
- preserve the existing per-job distribution and provider-order semantics
- add a regression proving process hash changes cannot alter the selected order

## Mutation proof
Before the production change, simulating two process hash seeds changed the same job's order from alpha/beta/gamma to beta/gamma/alpha and the regression failed.

## Validation
- tests/test_video_providers.py: 6 passed
- git diff --check: clean

## Payout

### DIFF
--- a/tests/test_video_providers.py
+++ b/tests/test_video_providers.py
@@ -1,4 +1,6 @@
 # SPDX-License-Identifier: MIT
+import builtins
+
 from video_providers import ProviderRegistry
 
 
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1954, opened 2026-09-01
Preserve news RSS XML when content contains CDATA terminators

### LINKED ISSUE #1953: News RSS becomes malformed when content contains a CDATA terminator
## Problem

The news RSS route interpolates video titles, descriptions, and creator names directly into CDATA sections. If any value contains the valid text sequence ]]>, it closes the CDATA block early and the entire RSS document becomes malformed XML.

## Reproduction

Render /news/rss with a news item whose title, description, or display name contains ]]>. A standard XML parser rejects the feed.

## Impact

One ordinary piece of creator content can make every feed item unreadable to RSS clients until it leaves the feed window.

## Expected

The feed should remain well-formed and preserve the original text exactly.

## Proposed fix

Split embedded CDATA terminators using the standard adjacent-CDATA representation for all three CDATA-backed fields and add a parser-level regression.

nwinkelman2:
Closes #1953.

## What changed
- add one CDATA-safe text renderer that splits embedded terminators into adjacent sections
- apply it consistently to RSS title, description, and creator fields
- preserve the original reader-visible text
- add a parser-level regression covering all three fields

## Mutation proof
Before the production fix, ElementTree rejected the generated feed as malformed XML at the first embedded CDATA terminator.

## Validation
- news RSS + DB-path suites: 3 passed
- git diff --check: clean

## Payout

### DIFF
--- a/news_routes.py
+++ b/news_routes.py
@@ -14,6 +14,11 @@ BASE_DIR = Path(os.environ.get("BOTTUBE_BASE_DIR", str(Path(__file__).resolve().
 DB_PATH = BASE_DIR / "bottube.db"
 
 
+def _cdata_text(value) -> str:
+    """Render arbitrary text safely inside an XML CDATA section."""
+    return str(value or "").replace("]]>", "]]]]><![CDATA[>")
+
+
 def _get_db():
     """Retrieve db. Returns a SQLite database connection.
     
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1956, opened 2026-09-01
Allow automatic mood transitions under frequent updates

### LINKED ISSUE #1955: Frequent mood updates prevent every automatic transition
## Problem

MoodEngine uses time since last_updated for its one-hour transition gate, but every sub-hour update advances last_updated while refreshing intensity. A scheduler calling update_mood every few minutes therefore keeps resetting the gate and the agent can never transition automatically.

## Reproduction

Force an energetic state, call update_mood every ten minutes for an hour, and make the transition draw deterministic. At the one-hour mark the state is still energetic because the gate sees only ten minutes since the last refresh.

## Expected

Intensity refreshes should use last_updated, while transition eligibility should use the age of the current state (started_at).

## Proposed fix

Separate state age from refresh age and add a clock-controlled regression covering frequent scheduler calls.

nwinkelman2:
Closes #1955.

## What changed
- gate transition eligibility on current-state age (started_at)
- keep intensity decay based on the incremental refresh age (last_updated)
- add a deterministic clock regression modeling ten-minute scheduler calls

## Mutation proof
Before the production fix, six ten-minute updates left the agent energetic after a full hour because each refresh restarted the transition timer.

## Validation
- mood regression + complete mood-engine suite: 16 passed
- git diff --check: clean

## Payout

### DIFF
--- a/mood_engine.py
+++ b/mood_engine.py
@@ -522,9 +522,12 @@ class MoodEngine:
             self._save_mood(agent_id, new_mood)
             return new_mood
         
-        # Check if enough time has passed for mood change
+        # Transition eligibility is based on the age of the current state.
+        # ``last_updated`` advances on every intensity refresh, so using it as
+        # the gate would let frequent callers postpone transitions forever.
+        time_in_mood = current_time - current_mood.started_at
         time_since_update = current_time - current_mood.last_updated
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1958, opened 2026-09-01
Use actual watch timestamps for recommendation affinity

### LINKED ISSUE #1957: Recommendation affinity decays from upload time instead of watch time
## Problem

Recommended mode builds watch history with v.created_at AS watched_at. That is the video's publication timestamp, not when the user watched it (w.created_at).

Category affinity therefore treats a freshly watched older video as an old preference and can heavily decay or effectively discard the user's real recent interest.

## Reproduction

Give a user two views with known view timestamps that differ from the videos' upload timestamps, enter recommended mode, and inspect the history passed to the ranking engine. It receives the upload times.

## Expected

Affinity decay must use the timestamp of the user's view event.

## Proposed fix

Select w.created_at as watched_at and add an integration regression that captures the exact history passed into recommendation scoring.

nwinkelman2:
Closes #1957.

## What changed
- select the view event timestamp as watched_at for recommendation history
- preserve upload timestamps only for video freshness scoring
- add an integration regression capturing the exact history given to the ranker

## Mutation proof
Before the production fix, known view times 200 and 100 were replaced by the videos' much newer publication timestamps.

## Validation
- complete recommendation-engine suite: 35 passed
- git diff --check: clean

### DIFF
--- a/recommendation_engine.py
+++ b/recommendation_engine.py
@@ -493,7 +493,7 @@ def get_feed_recommendations(
     if mode == "recommended" and agent_id:
         # Get user's watch history for affinity
         watch_history = db.execute(
-            """SELECT v.category, v.created_at AS watched_at
+            """SELECT v.category, w.created_at AS watched_at
                FROM views w
                JOIN videos v ON w.video_id = v.video_id
                WHERE w.agent_id = ?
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1960, opened 2026-09-01
[TRAVIS] Claim BAN payouts before the irreversible send edge

### LINKED ISSUE #1959: BAN payout workers can replay an irreversible withdrawal
## Financial integrity defect

`banano_payout.py` discovers pending withdrawals and calls the irreversible `wallet.send()` without first atomically claiming each row. Two overlapping cron processes can select the same pending transaction and both send it. A process/network failure after submission but before the local `sent` update also leaves the row pending, causing the next cron run to replay a potentially successful transfer.

## Required invariant

- Exactly one payout worker may cross the send edge for a pending row.
- A transport result whose on-chain outcome is unknown must not be automatically replayed.
- Successful/permanent-failure transitions must apply only to the worker's claimed row.

## Proof shape

A compare-and-set regression must show the first worker claiming the row and the second losing the claim. A simulated send timeout followed by another processor run must produce one send attempt total and leave the row quarantined for reconciliation, not pending for replay.

nwinkelman2:
## Summary - atomically compare-and-set each pending BAN withdrawal to `processing` before the irreversible send edge - let only the winning worker send when cron processes overlap - quarantine transport-uncertain outcomes instead of silently replaying a possibly successful block - scope sent/failed transitions to the claimed processing row  Closes #1959.  ## Proof - claim competition: first connection wins, second connection returns false - timeout replay: two processor runs produce exactly one `wallet.send()` call and durable status `uncertain` - focused regressions: **2 passed** - compile and diff-integrity checks: clean  ## Payout Native RTC wallet: `RTCe625bc2d5c25d4348236a4fcb74e5d0ef01f6d6d` 
Follow-up hardening: `processing` and `uncertain` withdrawals now remain reserved in public balance, tip admission, and second-withdrawal admission, preventing the new safe payout states from making committed BAN spendable again. Six payout/ledger regressions pass.

### DIFF
--- a/banano_payout.py
+++ b/banano_payout.py
@@ -80,6 +80,27 @@ def _ban_rpc(action_data: dict) -> dict:
     return {"error": "all RPC endpoints failed"}
 
 
+def _claim_withdrawal(db, tx_id: int) -> bool:
+    """Atomically claim one pending withdrawal for this payout worker.
+
+    The pending list is only discovery.  Multiple cron processes may discover
+    the same row, so the irreversible send edge must be guarded by a
+    compare-and-set under a SQLite write reservation.
+    """
+    try:
+        db.execute("BEGIN IMMEDIATE")
+        claimed = db.execute(
+            "UPDATE ban_transactions SET status = 'processing', processed_at = ? "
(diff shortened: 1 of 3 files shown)


Scottcjn:

Scottcjn (approved):

PR #1962, opened 2026-09-01
Accept finite fractional Unix query timestamps

### LINKED ISSUE #1961: Shared timestamp validators reject fractional Unix seconds
## Problem

Both shared timestamp parsers document support for integer or float Unix timestamps, but the implementation tries int() and then treats every failure as either invalid or ISO-8601. Fractional epoch values such as 1700000000.125 therefore return HTTP 400.

This drops sub-second cursor precision and contradicts the public helper contract.

## Expected

Accept finite fractional Unix seconds, preserve integer return behavior for integer input, and reject non-finite numeric values.

## Proposed fix

Try a finite float after integer parsing and before the ISO fallback, with regressions for both parser variants and nan/inf rejection.

nwinkelman2:
Closes #1961.

## What changed
- preserve the fast integer path and integer return behavior
- parse finite fractional epoch values before the ISO-8601 fallback
- reject nan and infinities explicitly
- cover both public timestamp parsers and non-finite values

## Mutation proof
Before the production fix, both public parsers returned 400 for 1700000000.125.

## Validation
- complete shared query-validation suite: 15 passed
- git diff --check: clean

### DIFF
--- a/bottube_validators/validators.py
+++ b/bottube_validators/validators.py
@@ -15,6 +15,7 @@ request that did not supply the parameter at all.
 
 from __future__ import annotations
 
+import math
 from datetime import datetime, timezone
 from typing import Any, Iterable, Optional, Tuple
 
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1964, opened 2026-09-01
[TRAVIS] Claim Base wRTC withdrawals before minting

### LINKED ISSUE #1963: Base wRTC payout workers can mint the same withdrawal twice
## Defect
`POST /api/base-bridge/process-withdrawals` reads a batch of `pending` withdrawals, then each worker performs an unconditional `UPDATE ... SET status='processing' WHERE id=?` before minting on Base.

Two concurrent admin/cron invocations can read the same row, both mark it processing, and both reach the irreversible `send_raw_transaction` edge. A transport failure after submission is also recorded as ordinary `failed`, even though the node may already have accepted the transaction.

## Impact
A single queued withdrawal can be minted more than once. Ambiguous send outcomes lose the distinction required for safe operator reconciliation.

## Proposed repair
- atomically claim with `WHERE id=? AND status='pending'` and proceed only when exactly one row changed
- scope terminal transitions to the claimed `processing` row
- mark exceptions after submission begins as `uncertain`, never implicitly retryable
- add regression proof that only one claim can win

nwinkelman2:
## Summary - atomically claim each pending Base wRTC withdrawal before building or broadcasting its mint transaction - allow only the worker whose `pending -> processing` transition changed one row to reach the irreversible send edge - scope completion/failure writes to that claimed processing state - preserve ambiguous post-submission failures as `uncertain` for reconciliation instead of treating them as safely retryable  ## Proof - `python -m pytest tests/test_base_wrtc_bridge_validation.py test_bridge_admin_auth.py -q` — 25 passed - `python -m py_compile base_wrtc_bridge_blueprint.py` - `git diff --check` - regression proves a second claim for the same queued withdrawal cannot win  Closes #1963  Native RTC wallet: RTCe625bc2d5c25d4348236a4fcb74e5d0ef01f6d6d 
Follow-up integration proof: ambiguous `uncertain` withdrawals remain included in bridge liability count and volume instead of disappearing from operator statistics. The focused/adjacent gate is now 26 passed.

### DIFF
--- a/base_wrtc_bridge_blueprint.py
+++ b/base_wrtc_bridge_blueprint.py
@@ -133,6 +133,20 @@ def init_base_wrtc_tables(db):
     db.commit()
 
 
+def _claim_base_withdrawal(db, withdrawal_id):
+    """Atomically claim one queued withdrawal before any on-chain side effect."""
+    cursor = db.execute(
+        """
+        UPDATE base_wrtc_withdrawals
+        SET status = 'processing'
+        WHERE id = ? AND status = 'pending'
+        """,
+        (withdrawal_id,),
+    )
+    db.commit()
+    return cursor.rowcount == 1
+
+
 def _json_object_body():
     """Return request JSON when it is an object, or a 400 response tuple."""
     data = request.get_json(silent=True)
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1966, opened 2026-09-01
fix(syndication): reject malformed partner templates

### LINKED ISSUE #1965: Partner syndication crashes on malformed payload templates
## Problem
`PartnerAPIAdapter.validate_config()` accepts a malformed JSON `payload_template`, while `syndicate()` parses that template before entering its guarded request block. A single bad partner configuration therefore raises `JSONDecodeError` and can escape the adapter's documented `SyndicationResult` failure contract, interrupting the poll cycle.

## Reproduction
Configure a partner adapter with `payload_template='{"title": "{{video_title}}"'`. Validation currently returns `true`, and syndicating a payload raises before any HTTP request is made.

## Expected
Malformed templates fail configuration validation and direct syndication calls return a failed `SyndicationResult` without making an outbound request.

nwinkelman2:
## Summary
- reject malformed partner `payload_template` JSON during adapter validation
- preserve the `SyndicationResult` failure contract for direct calls
- prove malformed templates never trigger an outbound request

## Mutation proof
Before the fix, `test_invalid_payload_template_fails_closed` failed because validation returned `True`; continuing to syndication raised an uncaught `JSONDecodeError` before the request guard.

## Tests
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --no-project --with pytest --with requests pytest -q tests/test_syndication_adapter.py` (31 passed)
- `git diff --check`

Fixes #1965

### DIFF
--- a/syndication_adapter.py
+++ b/syndication_adapter.py
@@ -11,6 +11,7 @@ Adapters are stateless and focused solely on platform-specific API calls.
 Queue management and retry logic are handled by syndication_queue.py.
 """
 
+import json
 import logging
 from abc import ABC, abstractmethod
 from dataclasses import dataclass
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1968, opened 2026-09-01
[TRAVIS] Claim Solana wRTC withdrawals before transfer

### LINKED ISSUE #1967: Solana wRTC workers can double-pay and refund an ambiguous transfer
## Defect
The Solana wRTC withdrawal processor selects `pending` rows but never claims them before running `send_wrtc.mjs`. Concurrent admin/cron invocations can therefore execute the same irreversible token transfer more than once.

The same route treats every non-zero helper exit as definitely failed and refunds the user. Because the helper can fail after transaction submission or confirmation ambiguity, that can leave the user with both the on-chain transfer and restored platform balance.

## Impact
Duplicate wRTC payouts and payout-plus-refund loss are possible under concurrent workers or ambiguous Solana RPC transport failures.

## Proposed repair
- atomically claim each row using a status-qualified `pending -> processing` update
- only the worker whose claim changed one row may invoke the transfer helper
- scope terminal writes to the claimed processing state
- preserve any ambiguous helper/transport failure as `uncertain` without automatic refund
- add regression proof that only one claim can win

nwinkelman2:
## Summary
- atomically claim each pending Solana wRTC withdrawal before invoking the transfer helper
- allow only the worker whose `pending -> processing` transition changed one row to reach the irreversible send edge
- scope completion and uncertainty writes to the claimed processing state
- fail closed on helper errors/timeouts as `uncertain` without restoring platform balance, because an on-chain submission may already exist

## Proof
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_wrtc_payout_claim.py -q` — 1 passed
- direct SQLite proof: first claim succeeds, second claim for the same row fails, state remains `processing`
- `python -m py_compile wrtc_bridge.py`
- `git diff --check`

Closes #1967

### DIFF
--- a/tests/test_wrtc_payout_claim.py
+++ b/tests/test_wrtc_payout_claim.py
@@ -1,0 +1,26 @@
+# SPDX-License-Identifier: MIT
+"""Regression proof for Solana wRTC payout admission."""
+
+import sqlite3
+
+import wrtc_bridge
+
+
+def test_only_one_worker_can_claim_a_pending_withdrawal():
+    db = sqlite3.connect(":memory:")
+    db.executescript(wrtc_bridge.WRTC_SCHEMA)
+    db.execute(
+        """
+        INSERT INTO wrtc_withdrawals
+            (agent_id, amount_wrtc, fee_wrtc, net_wrtc, to_address, status, created_at)
+        VALUES (?, ?, ?, ?, ?, 'pending', ?)
+        """,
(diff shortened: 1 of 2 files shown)


nwinkelman2:
Closing after a registration audit found that production registers `wrtc_bridge_blueprint.wrtc_bp`, not this legacy `wrtc_bridge.py` blueprint. The race is real inside the legacy module, but it is not on the active application path, so this should not be presented as a production bounty fix. No reward is claimed for it.

PR #1970, opened 2026-09-01
fix(seo): tolerate malformed video numerics

### LINKED ISSUE #1969: Video JSON-LD crashes on malformed numeric metadata
## Problem
`build_video_jsonld()` converts persisted duration, dimensions, upload time, view count, and comment count with unguarded `float()` / `int()` calls. A malformed legacy value (or a non-finite value such as `NaN`) raises while rendering an otherwise valid watch page, so one metadata field can suppress both the page and its schema markup.

## Reproduction
Call `build_video_jsonld()` with a valid `video_id` and `duration_sec="unknown"`. It raises `ValueError`. `duration_sec="nan"` also raises when converted to `int`; non-finite timestamps fail at `datetime.fromtimestamp()`.

## Expected
Malformed and non-finite numeric metadata should fall back to safe schema defaults while preserving valid video content and JSON-LD.

nwinkelman2:
## Summary
- normalize malformed or non-finite persisted numeric metadata before building VideoObject JSON-LD
- preserve safe duration, dimension, interaction, and upload-date defaults
- cover both invalid strings and `NaN` / infinity regressions

## Mutation proof
Before the fix, both regressions raised from `build_video_jsonld()`: invalid strings failed `float()` conversion, while `NaN` failed integer conversion. The function now returns valid schema data for both cases.

## Tests
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --no-project --with pytest --with flask pytest -q tests/test_seo_video_jsonld.py` (2 passed)
- `python -m py_compile seo_routes.py`
- `git diff --check`

Fixes #1969

### DIFF
--- a/seo_routes.py
+++ b/seo_routes.py
@@ -4,7 +4,7 @@
 # AEO, GEO, E-E-A-T, Semantic Entity Mapping — 2026 Edition
 # ---------------------------------------------------------------------------
 
-import html, json, re, time
+import html, json, math, re, time
 from flask import Blueprint, current_app, request
 from datetime import datetime, timezone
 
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1972, opened 2026-09-01
fix(media): tolerate metadata embedding timeouts

### LINKED ISSUE #1971: Metadata timeout discards prepared video and leaves partial file
## Problem
Media metadata embedding is treated as best-effort when FFmpeg exits with an error, preserving the already-transcoded video. However, the same FFmpeg call has a 120-second timeout that is not caught. A timeout therefore escapes `_embed_metadata()`, fails the entire preparation pipeline, and can leave a partial `.tmp.mp4` behind even though the prepared video is usable.

## Reproduction
Mock the metadata FFmpeg call to raise `subprocess.TimeoutExpired` after creating its temporary output. `_embed_metadata()` propagates the exception and leaves the partial file.

## Expected
A metadata timeout should follow the same best-effort failure path as an FFmpeg exit error: preserve the prepared video, log the failure, and remove the partial temp file.

nwinkelman2:
## Summary
- handle FFmpeg metadata timeouts through the existing best-effort failure path
- preserve the already-prepared video instead of failing the entire pipeline
- remove partial metadata output after either timeout or process failure

## Mutation proof
Before the fix, the new regression raised `subprocess.TimeoutExpired` from `_embed_metadata()` and left `prepared.tmp.mp4` behind. It now returns normally, preserves the original video bytes, and removes the partial file.

## Tests
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --no-project --with pytest pytest -q tests/test_media_prep.py` (22 passed)
- `python -m py_compile media_prep.py`
- `git diff --check`

Fixes #1971

### DIFF
--- a/media_prep.py
+++ b/media_prep.py
@@ -503,7 +503,7 @@ class MediaPrepPipeline:
         try:
             subprocess.run(cmd, capture_output=True, timeout=120, check=True)
             temp_path.replace(video_path)
-        except subprocess.CalledProcessError as e:
+        except (subprocess.CalledProcessError, subprocess.TimeoutExpired) as e:
             log.warning(f"Metadata embedding failed: {e}")
             if temp_path.exists():
                 temp_path.unlink()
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1974, opened 2026-09-01
fix(syndication): close poll watermark gap

### LINKED ISSUE #1973: Syndication poll watermark can skip uploads created during fetch
## Problem
The syndication daemon advances `last_poll_time` with a new timestamp after `/api/feed` returns. A video created after the feed query snapshot but before that timestamp is assigned is absent from the current response and excluded by the next cycle's `since` filter, so it may never be queued.

## Reproduction
Start a poll at time 100, let the feed request complete at time 200, and create a video during the request window. The current daemon records 200 as its next high-water mark, skipping the in-flight upload.

## Expected
Capture the poll high-water mark immediately before issuing the feed request and advance to that value only after processing the response. Uploads created during the request then remain eligible on the next cycle.

nwinkelman2:
## Summary
- capture each feed poll's high-water mark before issuing the request
- advance `last_poll_time` to the request-start boundary after processing
- keep uploads created during an in-flight request eligible for the next cycle

## Mutation proof
The regression advances a controlled clock from 100 to 200 while `fetch_new_videos()` is in flight. Before the fix, the daemon recorded 200 and skipped that window; it now records 100.

## Tests
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 uv run --no-project --with pytest --with requests --with pyyaml pytest -q tests/test_syndication_poller.py -k poll_watermark` (1 passed)
- `python -m py_compile syndication_poller.py`
- `git diff --check`

Fixes #1973

### DIFF
--- a/syndication_poller.py
+++ b/syndication_poller.py
@@ -643,7 +643,10 @@ class SyndicationPoller:
                     time.sleep(min(remaining, 10))
                     continue
 
-                # Poll for new videos
+                # Capture the high-water mark before the request. Videos created
+                # while the feed is being fetched must remain eligible for the
+                # next cycle instead of falling into a timestamp gap.
+                poll_started_at = time.time()
                 new_videos = self.fetch_new_videos(
                     since=self.last_poll_time if self.last_poll_time > 0 else None
                 )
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1978, opened 2026-09-01
[TRAVIS] Keep Python SDK large uploads memory-bounded

### LINKED ISSUE #1977: Python SDK large-upload path still buffers the entire file in RAM
## Problem

The stdlib Python SDK routes files above 100 MB through `_streaming_upload` and promises chunk-bounded memory usage, but the implementation fully materializes the multipart body immediately before `urlopen`:

```python
with tempfile.SpooledTemporaryFile(...) as tmp:
    for part in body_generator():
        tmp.write(part)
    tmp.seek(0)
    req.data = tmp.read()
```

`SpooledTemporaryFile` limits temporary-storage memory while the body is assembled, but `tmp.read()` then allocates one bytes object containing the complete upload. A 2 GB upload therefore still needs roughly 2 GB of contiguous RAM at the send edge, which is the exact failure this method is intended to avoid.

Python's urllib/http.client request path accepts an iterable of byte chunks. The existing generator can be supplied directly while retaining the explicit multipart `Content-Length`.

## Acceptance

- Keep the outbound multipart request body lazy/chunked for the entire send path.
- Preserve the exact multipart `Content-Length`.
- Preserve metadata fields, file bytes, and response handling.
- Add a regression proving the request body is not bytes and the file is yielded according to `chunk_size`.
- Keep existing Python SDK tests green.

nwinkelman2:
## Summary

- pass the existing multipart body generator directly to urllib instead of copying it back into one in-memory bytes object
- retain the explicit multipart `Content-Length`, metadata fields, file headers, and response handling
- add a regression that proves request data stays iterable, file bytes are yielded at the configured chunk size, and the declared length matches the wire body

Closes #1977

## Proof

Before this change, `_streaming_upload` wrote the multipart body to `SpooledTemporaryFile` and then called `tmp.read()`, allocating the entire upload in RAM at the irreversible send edge. The new regression fails against that behavior because `request.data` is bytes rather than a lazy iterable.

The passing regression observes file payload chunks of `3, 3, 3, 1` bytes with a test `chunk_size=3`, joins the complete wire body, and verifies its length exactly equals the request's `Content-Length`.

## Validation

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 py -3 -m pytest python-sdk/tests/test_client.py -q` -> 8 passed
- `py -3 -m compileall -q python-sdk/bottube` -> passed

### DIFF
--- a/python-sdk/bottube/client.py
+++ b/python-sdk/bottube/client.py
@@ -230,14 +230,12 @@ class BoTTubeClient:
                     yield chunk
             yield closing
         
-        req = Request(url, data=None, headers=headers, method="POST")
-        # Use a streaming body via a temporary file to avoid loading all into memory
-        import tempfile
-        with tempfile.SpooledTemporaryFile(max_size=10 * 1024 * 1024) as tmp:
-            for part in body_generator():
-                tmp.write(part)
-            tmp.seek(0)
-            req.data = tmp.read()
+        # urllib/http.client accepts an iterable of bytes as a request body.
+        # Supplying the generator directly keeps large uploads bounded by
+        # ``chunk_size``.  Materialising it through ``tmp.read()`` would copy
+        # the entire (potentially multi-gigabyte) multipart body back into RAM
+        # immediately before sending it, defeating this method's contract.
+        req = Request(url, data=body_generator(), headers=headers, method="POST")
 
         try:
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1980, opened 2026-09-01
[TRAVIS] Preserve JavaScript SDK HTTP response semantics

### LINKED ISSUE #1979: JavaScript SDK loses HTTP status on empty or non-JSON responses
## Problem

The JavaScript SDK calls `res.json()` before it checks `res.ok` in both JSON and multipart request paths. Two normal HTTP cases therefore leak a raw `SyntaxError` instead of preserving API semantics:

1. A successful empty `204 No Content` from a DELETE-style endpoint rejects even though the operation succeeded.
2. A proxy/gateway error with an empty or HTML body (for example `502 Bad Gateway`) rejects while parsing and discards the HTTP status, so callers cannot classify, report, or retry the failure through `BoTTubeError`.

## Acceptance

- Accept a successful empty 204 response for void operations.
- Convert a non-JSON HTTP error into `BoTTubeError` while retaining its status code and status text fallback.
- Continue surfacing malformed JSON on a successful non-204 response rather than silently accepting corruption.
- Apply the behavior consistently to JSON and multipart requests.
- Add regressions for empty 204 and non-JSON 502 responses.

nwinkelman2:
## Summary

- centralize JavaScript SDK response decoding for both JSON and multipart requests
- allow successful empty `204 No Content` responses for void operations
- preserve non-JSON HTTP failures as typed `BoTTubeError` values with the original status and status-text fallback
- keep malformed successful non-204 JSON fail-fast
- add regressions for the empty-204 and HTML/empty-gateway-error paths
- refresh committed CJS and ESM bundles

Closes #1979

## Proof

Before this change both cases failed at `res.json()` before status handling:

- a successful `204` rejected with `SyntaxError: Unexpected end of JSON input`
- a `502 Bad Gateway` with an HTML body rejected with `SyntaxError` and lost status 502

Focused mutations now prove the delete resolves and the gateway response rejects as `BoTTubeError { statusCode: 502, apiError: { error: "Bad Gateway" } }`.

## Validation

- `npx vitest run tests/client.test.ts -t "accepts a successful empty 204 response|preserves the status as a typed SDK error"` -> 2 passed
- CJS and ESM bundle stages completed successfully

Repository baseline notes: the broad suite's existing timeout mock never reacts to `AbortSignal` and times out; declaration bundling/typecheck is currently blocked by the repository's TypeScript 7 + removed `moduleResolution=node10` configuration. Neither failure intersects this response-decoding change; upstream CI remains authoritative.

### DIFF
--- a/js-sdk/dist/index.js
+++ b/js-sdk/dist/index.js
@@ -76,6 +76,21 @@ var BoTTubeClient = class {
     if (this.apiKey) h["X-API-Key"] = this.apiKey;
     return h;
   }
+  async responseData(res) {
+    let data;
+    try {
+      data = await res.json();
+    } catch (err) {
+      if (res.ok && res.status === 204) return void 0;
+      if (!res.ok) {
+        const message = res.statusText?.trim() || `HTTP ${res.status}`;
+        throw new BoTTubeError(res.status, { error: message });
+      }
+      throw err;
+    }
+    if (!res.ok) throw new BoTTubeError(res.status, data);
+    return data;
+  }
   async request(method, path, body) {
     const url = `${this.baseUrl}${path}`;
     const controller = new AbortController();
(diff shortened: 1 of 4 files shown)


Scottcjn (approved):

PR #1982, opened 2026-09-01
[TRAVIS] Keep feeds available for invalid epoch timestamps

### LINKED ISSUE #1981: RSS and Atom feeds crash on out-of-range video timestamps
## Summary

The public RSS and Atom feed routes can return HTTP 500 when a video carries an out-of-range numeric `created_at` value. Both timestamp helpers detect numeric strings before their protected ISO parsing branch and call `datetime.fromtimestamp(...)` without handling `OverflowError`, `OSError`, or non-finite values.

## Reproduction

On current `main`:

```python
from feed_blueprint import _to_rfc2822, _to_iso8601

_to_rfc2822("9" * 400)
_to_iso8601(float("inf"))
```

The first call raises `OverflowError: timestamp out of range for platform time_t`; non-finite numeric values fail similarly. A single malformed or legacy timestamp returned by `/api/videos` therefore aborts `/feed/rss`, `/feed/atom`, and their per-agent variants instead of allowing the feed to remain available.

## Expected

Unsupported epoch values use the same safe current-time fallback already used for malformed ISO timestamps. The feed remains valid XML and available.

## Actual

Numeric branches bypass the existing exception fallback and can take the complete feed offline.

## Environment

- Windows 11
- Python 3.14
- Repository `main` at `dcb0f84`

## Non-duplicate check

Searched current open BoTTube issues for RSS, Atom, feed, timestamp, and date handling. Existing reports cover social-feed routing, activity-feed loading, CDATA termination, mood timing, and fractional validators; none cover out-of-range feed epochs.

nwinkelman2:
## Summary

- bound epoch conversion for RSS and Atom timestamps
- reject non-finite and platform-out-of-range epochs through the existing current-time fallback
- cover both timestamp helpers and all four public feed routes

## Mutation proof

Before the fix, `_to_rfc2822("9" * 400)` raised `OverflowError: timestamp out of range for platform time_t`. The new route regression exercises that same value through RSS, Atom, and both per-agent variants.

## Tests

```text
8 passed in 3.21s
python -m py_compile feed_blueprint.py tests/test_feed_blueprint.py
```

The complete existing `tests/test_feed_blueprint.py` run is 29 passing tests plus one unrelated pre-existing failure in `test_fetch_videos_builds_filtered_request_and_normalizes_response`; that test calls `_fetch_videos` without a Flask request context even though current `main` reads `request.host` before invoking its mocked HTTP client. This patch does not touch that path.

Closes #1981

### DIFF
--- a/feed_blueprint.py
+++ b/feed_blueprint.py
@@ -3,6 +3,7 @@
 # Author: @AUTHENSOR
 # BCOS-Tier: L1
 import datetime
+import math
 import os
 from email.utils import format_datetime
 
@@ -59,6 +60,17 @@ def _video_description_html(fields):
     return f'<img src="{thumb}" /><p>{desc}</p>'
 
 
+def _epoch_datetime_or_now(value):
+    """Return a bounded UTC datetime for an epoch-like value."""
+    try:
+        epoch = float(value)
+        if not math.isfinite(epoch):
+            raise ValueError("epoch must be finite")
+        return datetime.datetime.fromtimestamp(epoch, tz=datetime.timezone.utc)
+    except (OverflowError, OSError, TypeError, ValueError):
+        return datetime.datetime.now(datetime.timezone.utc)
+
+
 def _to_rfc2822(value):
     """Convert various timestamp formats to RFC 2822 for RSS pubDate."""
     if value is None or value == "":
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #1987, opened 2026-09-01
fix(mood): reject non-object JSON bodies

### LINKED ISSUE #1986: Mood write endpoints crash on non-object JSON bodies
### Problem
Both authenticated mood-write endpoints assume every truthy JSON body is an object:

- `POST /api/v1/agents/<agent>/mood/update`
- `POST /api/v1/agents/<agent>/mood/signal`

A valid caller can send a JSON array (for example `["not", "an", "object"]`). Flask parses it successfully, then each handler calls `.get(...)` on the list and raises `AttributeError`, producing an internal server error instead of a bounded client response.

### Reproduction
Using a valid agent API key, POST an array as the JSON body to either route. The update route fails at `data.get("force_state")`; the signal route fails at `data.get("signal_type")`.

Focused mutation result: 2 failures, both `AttributeError: 'list' object has no attribute 'get'`.

### Expected
Both endpoints should require a JSON object and return the repository's established `400 {"error": "JSON body must be an object"}` contract for arrays, strings, and other non-object JSON values. Existing ownership checks and valid object behavior must remain unchanged.

### Proposed fix
Validate the parsed body type immediately after authorization in both handlers and add parametrized regressions covering the two endpoints alongside their existing ownership tests.

nwinkelman2:
## Summary
- require JSON objects on both authenticated mood-write endpoints
- return the existing 400 client-error contract for arrays/strings instead of raising on `.get`
- preserve optional empty-body behavior for mood updates and existing authorization behavior
- add parametrized regressions for update and signal routes

Fixes #1986

## Mutation proof
Before the fix, the new two-case regression failed twice with `AttributeError: 'list' object has no attribute 'get'`, at the update route's `force_state` lookup and the signal route's `signal_type` lookup.

## Validation
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_mood_write_authorization.py -q` -> 10 passed
- `python -m py_compile bottube_server.py tests/test_mood_write_authorization.py`
- `git diff --check`

### DIFF
--- a/tests/test_mood_write_authorization.py
+++ b/tests/test_mood_write_authorization.py
@@ -146,6 +146,19 @@ def test_owner_can_still_record_own_signal(client, victim):
     assert resp.status_code == 200, resp.get_json()
 
 
+@pytest.mark.parametrize("endpoint", ["update", "signal"])
+def test_mood_writes_reject_non_object_json(client, victim, endpoint):
+    """Arrays must produce a client error instead of crashing on ``.get``."""
+    resp = client.post(
+        f"/api/v1/agents/{victim['agent_name']}/mood/{endpoint}",
+        json=["not", "an", "object"],
+        headers={"X-API-Key": victim["api_key"]},
+    )
+
+    assert resp.status_code == 400, resp.get_json()
+    assert resp.get_json() == {"error": "JSON body must be an object"}
+
+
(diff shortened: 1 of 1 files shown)


Scottcjn:

Scottcjn (approved):

PR #1989, opened 2026-09-01
fix(feed): preserve limit in Atom self link (#1988)

### LINKED ISSUE #1988: Atom feed self link drops the requested limit filter
## Bug

`GET /feed/atom` supports `?limit=N` and passes that value to `_fetch_videos()`, but the generated Atom `rel="self"` URL only preserves the `agent` and `category` query parameters.

For example, a request for:

`/feed/atom?limit=5`

returns a feed built with 5 videos, while its self link is:

`https://bottube.ai/feed/atom`

That self URL represents the default 20-item feed, not the current 5-item resource.

## Reproduction from current source

In `feed_blueprint.py`, `atom_feed()` builds `self_params` from only `agent` and `category`, even though `limit = _parse_limit()` is part of the effective request state.

## Expected

The Atom self link should preserve all effective filters that define the current feed, including `limit`, e.g. `https://bottube.ai/feed/atom?limit=5`, and combined filters should be correctly URL-encoded.

## Impact

Feed readers/caches following `rel="self"` can resolve a different resource than the one originally fetched, changing the item count and losing the caller's limit filter.

## Suggested fix

Build the self-link query string from the effective agent/category/limit parameters using standard URL encoding, and add regression coverage for `limit` alone and combined filters.

## Environment

Static reproduction against current `Scottcjn/bottube` `main`, `feed_blueprint.py`. No state-changing requests were used.

AI assistance was used in identifying and preparing this report.

mdmcl-pixel:
## Summary

Fixes #1988.

The global Atom feed accepts `?limit=N` and serves that representation, but its `rel="self"` URL previously omitted the explicit limit. Feed readers following the self link could therefore resolve the default 20-item feed instead of the representation they fetched.

## Changes

- preserve an explicitly supplied `limit` in the Atom self link;
- URL-encode agent/category/limit query parameters before XML escaping the self-link attribute;
- preserve the existing canonical URL when `limit` is omitted;
- add focused regression coverage for explicit limit, combined filters with reserved characters, and the no-limit default.

## Validation

Focused GitHub Actions validation passed:

`python -m pytest -q tests/test_feed_atom_self_link.py`

Result: 3 passed.

Final branch is rebuilt as one clean commit on current upstream `main`, with only `feed_blueprint.py` and `tests/test_feed_atom_self_link.py` changed.

AI assistance was used in identifying, implementing, and testing this contribution.

### DIFF
--- a/feed_blueprint.py
+++ b/feed_blueprint.py
@@ -5,6 +5,7 @@
 import datetime
 import os
 from email.utils import format_datetime
+from urllib.parse import urlencode
 
 import requests
 from flask import Blueprint, Response, jsonify, request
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):
Self link now carries the served limit and is built with urlencode + one escape; first fix for #1988.

PR #1993, opened 2026-09-02
fix(feed): preserve limit param in Atom self link

### LINKED ISSUE #1988: Atom feed self link drops the requested limit filter

rafaio1:
Closes #1988

## Problem
The Atom feed `rel="self"` link dropped the requested `limit` filter, causing readers following the self link to resolve a different resource (default 20 items) than the one originally fetched.

## Fix
- Include `limit` in `self_params` when non-default (not 20)
- Add regression tests for limit preservation and default omission

## Verification
```bash
curl -sS 'https://bottube.ai/feed/atom?limit=5' | grep 'rel="self"'
# Should now include ?limit=5 in the self href
```

Claim: /claim (already filed in #1988)
Wallet: RTC1e9bf7a2a60aac9bcbc5a0df0c65e9501e932861

### DIFF
--- a/feed_blueprint.py
+++ b/feed_blueprint.py
@@ -260,6 +260,8 @@ def atom_feed():
         self_params.append(f"agent={escape_xml(agent)}")
     if category:
         self_params.append(f"category={escape_xml(category)}")
+    if limit is not None and limit != 20:
+        self_params.append(f"limit={limit}")
     self_qs = f"?{'&amp;'.join(self_params)}" if self_params else ""
     self_href = f"https://bottube.ai/feed/atom{self_qs}"
 
--- a/tests/test_atom_self_link_limit.py
+++ b/tests/test_atom_self_link_limit.py
@@ -1,0 +1,41 @@
+import re
+from feed_blueprint import atom_feed
+
+class FakeRequest:
+    def __init__(self, args):
+        self.args = args
+
+def test_atom_self_link_preserves_limit(monkeypatch):
+    """Regression: Atom self link must include limit when non-default."""
+    from flask import Flask
+    app = Flask(__name__)
+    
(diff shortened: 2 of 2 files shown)


rafaio1:
/claim

Wallet: RTC1e9bf7a2a60aac9bcbc5a0df0c65e9501e932861

Scottcjn:
Closing as a duplicate: #1988 was filed by @mdmcl-pixel and their fix #1989 (Sept 1) predates this PR. First sound fix wins. — Sophia

PR #1874, opened 2026-09-01
[TRAVIS] Prevent HTML search pagination overflow

nwinkelman2:
Closes #1873.

## What changed
- compute the HTML search offset once
- reject non-empty searches whose OFFSET exceeds SQLite signed 64-bit range
- preserve ordinary result pagination and the empty-search landing page

## Repro / mutation proof
Before the fix, GET /search?q=x&page=9223372036854775807 raised OverflowError: Python int too large to convert to SQLite INTEGER from the LIMIT/OFFSET query. The focused regression file failed 1 test and passed 2 controls.

## Validation
- C:\Python314\python.exe -m pytest -q tests/test_search_page_html_offset_overflow.py -> 3 passed
- existing API search overflow assertions also passed; Windows-only fixture teardown retained a DB handle after those assertions, unrelated to this diff
- C:\Python314\python.exe -m py_compile bottube_server.py
- git diff --check

This is ordinary public UI pagination correctness and does not change authentication or security controls.

## Payout
Native RTC wallet: `RTCe625bc2d5c25d4348236a4fcb74e5d0ef01f6d6d`

Native RTC wallet: RTCe625bc2d5c25d4348236a4fcb74e5d0ef01f6d6d

Issue #1873, opened 2026-09-01
HTML search page crashes on oversized pagination offset


nwinkelman2:
## Summary\nThe public HTML search route (GET /search) accepts arbitrarily large positive page values. When a non-empty query is present, it multiplies page by the fixed page size and binds the result as a SQLite OFFSET. Values beyond SQLite's signed 64-bit range raise OverflowError and surface as a server error.\n\n## Reproduction\nGET /search?q=x&page=9223372036854775807\n\n## Expected\nReturn a clear 400 for a page whose computed offset cannot be represented by SQLite. Ordinary pages and the empty-search landing page should remain unchanged.\n\n## Scope\nOrdinary public UI pagination correctness; no authentication or security-control changes.

rafaio1:
/claim

Wallet: RTC1e9bf7a2a60aac9bcbc5a0df0c65e9501e932861

Will fix HTML /search pagination overflow with bounded OFFSET validation and regression tests per issue spec.

PR #1997, opened 2026-09-02
[TRAVIS] Bound HTML search pagination offsets

### LINKED ISSUE #1996: HTML search crashes on out-of-range pagination offset
## Summary

The HTML `/search` route accepts an unbounded positive `page` value and multiplies it by its 24-row page size before binding the result to SQLite. A page at SQLite's signed 64-bit ceiling produces an out-of-range OFFSET and crashes with `OverflowError: Python int too large to convert to SQLite INTEGER` instead of returning a bounded client response.

This is distinct from the already-fixed JSON `/api/search` overflow path: the server-rendered `/search` handler implements its own pagination parsing and query.

## Reproduction

```text
GET /search?q=video&page=9223372036854775807
```

On current `main`, a Flask client raises at the `LIMIT ? OFFSET ?` query in `search_page`; with production error handling this becomes HTTP 500.

## Expected

Reject a page whose computed OFFSET cannot fit SQLite's signed 64-bit integer range with HTTP 400. Ordinary and large-but-safe positive page values should continue to render normally.

## Proposed fix

Validate `(page - 1) * per_page` against SQLite's maximum signed integer before executing the query, return a clear 400 for overflow, and add focused regressions for the failing boundary plus a large safe page.

nwinkelman2:
## Summary

- bound the server-rendered search page's SQLite OFFSET before query execution
- reuse the validated offset in the query
- add regressions for the crashing int64 boundary and a large safe page

Closes #1996

## Mutation proof

Before the fix, this request raises `OverflowError: Python int too large to convert to SQLite INTEGER` in `search_page` before a response is produced:

```text
GET /search?q=video&page=9223372036854775807
```

After the fix it returns HTTP 400, while `page=1000000000` still renders HTTP 200.

## Validation

```text
python -m pytest tests/test_search_page_html_offset_overflow.py -q
2 passed
python -m py_compile bottube_server.py
git diff --check
```

A broader discoverability run also exposed six failures already present on current main (uploads not entering search results, two missing routes, and a legacy template expectation); they are unrelated to this two-line pagination guard.

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -14104,6 +14104,9 @@ def search_page():
     except (TypeError, ValueError):
         page = 1
     per_page = 24
+    offset = (page - 1) * per_page
+    if offset > _SQLITE_MAX_INT64:
+        abort(400, description="page out of range")
     # Category checkboxes submit repeated ?cat=<id> params.
     valid_cat_ids = {c["id"] for c in VIDEO_CATEGORIES}
     selected_categories = [c for c in request.args.getlist("cat") if c in valid_cat_ids]
@@ -14133,7 +14136,7 @@ def search_page():
                WHERE {where}
                ORDER BY v.views DESC, v.created_at DESC
                LIMIT ? OFFSET ?""",
-            params + [per_page, (page - 1) * per_page],
+            params + [per_page, offset],
         ).fetchall()
 
     pages = max(1, (total + per_page - 1) // per_page)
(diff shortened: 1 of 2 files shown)


nwinkelman2:
Withdrawing this PR: the same defect was already reported in #1873 and fixed in #1874. This duplicate was created after a title/keyword search failed to surface the earlier issue; no bounty credit is requested for #1997.

PR #1999, opened 2026-09-02
[TRAVIS] Restore bounded public search suggestions

### LINKED ISSUE #1998: Search suggestions contract is shipped but the endpoint returns 404
## Summary

BoTTube's discoverability contract includes a search-suggestions API and current regression coverage calls `GET /api/search/suggestions?q=Py`, but current `main` has no route for it. The endpoint returns HTTP 404, so clients cannot offer the title/category/creator/tag suggestions represented by the shipped discoverability tests.

This is separate from full search at `/api/search`: suggestions are bounded prefix/substring candidates used before users submit a full query.

## Reproduction

```text
GET /api/search/suggestions?q=Py
```

Current result: HTTP 404.

The existing `tests/test_discoverability.py::TestSearchEnhancements::test_search_suggestions` also fails at the same edge.

## Expected

Return a bounded JSON object with `suggestions`, `categories`, `agents`, and `tags`, drawing only from public/non-removed videos and non-banned agents. Empty or too-short queries should return empty arrays without scanning the catalog.

## Scope

Ordinary public discovery/API completeness only. No authentication, payment, or security-control changes.

nwinkelman2:
## Summary

- restore the shipped `GET /api/search/suggestions` discoverability contract
- return bounded public video, category, creator, and tag matches
- exclude removed content and banned creators
- escape SQL LIKE metacharacters and avoid catalog scans for sub-two-character queries
- add focused response, bounds, and wildcard regressions

Closes #1998

## Mutation proof

On current `main`, `GET /api/search/suggestions?q=Py` returns HTTP 404 and the existing discoverability regression fails at the route edge. This patch returns the documented result groups from seeded public catalog data.

## Validation

```text
python -m pytest tests/test_search_suggestions_contract.py -q
3 passed
python -m py_compile bottube_server.py
git diff --check
```

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -14090,6 +14090,88 @@ def join_page():
     return render_template("join.html")
 
 
+@app.route("/api/search/suggestions")
+def api_search_suggestions():
+    """Return bounded public catalog suggestions for a partial query."""
+    query = request.args.get("q", "").strip()
+    empty_result = {"suggestions": [], "categories": [], "agents": [], "tags": []}
+    if len(query) < 2:
+        return jsonify(empty_result)
+    if len(query) > 64:
+        return jsonify({"error": "q must be at most 64 characters"}), 400
+
+    # Treat wildcard characters as ordinary query text. Besides producing more
+    # useful suggestions, this prevents a partial-query request from turning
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2001, opened 2026-09-02
fix(admin): validate report hours windows

### LINKED ISSUE #2000: Admin reports silently coerce invalid hours windows
## Reproduction

Two authenticated admin reporting endpoints parse their documented `hours` query window with Werkzeug's `type=int` fallback and then clamp it:

- `GET /api/admin/visitors`
- `GET /api/admin/scan-content`

On current `main`, each returns HTTP 200 for `hours=abc`, `hours=1.5`, `hours=0`, `hours=-1`, and `hours=169`. Malformed values silently become the 24-hour default, while out-of-range values silently become 1 or 168.

An isolated temporary-database regression matrix has 10 failing cases on unmodified `main`; the documented boundary values 1 and 168 still pass.

## Expected

Both routes should use the existing `_parse_positive_int_query` contract and return structured HTTP 400 errors for malformed or out-of-range windows. The default should remain 24 hours and boundary values 1 and 168 should remain valid.

## Actual

Invalid input is silently accepted and can produce a report for a different time window than the caller requested.

## Scope

This is an ordinary input-validation/correctness fix. It does not change admin authentication, permissions, data mutation, or security controls.

nwinkelman2:
Resolves https://github.com/Scottcjn/bottube/issues/2000

## What changed

- route both documented admin `hours` windows through the existing strict positive-integer query parser
- reject malformed, fractional, non-positive, and over-168-hour values with a structured HTTP 400 response
- preserve the 24-hour default and the documented 1/168-hour boundaries
- add a 14-case integrated regression matrix across both endpoints

## How it was checked

- mutation baseline on unmodified `main`: 10 invalid-window assertions failed while 4 boundary assertions passed
- `python -m pytest tests/test_admin_hours_query_validation.py -q` -> 14 passed
- `python -m py_compile bottube_server.py tests/test_admin_hours_query_validation.py`
- `git diff --check`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -15088,7 +15088,9 @@ def admin_visitors():
     if err:
         return err
 
-    hours = min(168, max(1, request.args.get("hours", 24, type=int)))
+    hours, error = _parse_positive_int_query("hours", 24, max_value=168)
+    if error:
+        return error
     cutoff = time.time() - hours * 3600
 
     stats = {
@@ -16195,7 +16197,9 @@ def admin_scan_content():
     if err:
         return err
 
-    hours = min(168, max(1, request.args.get("hours", 24, type=int)))
+    hours, error = _parse_positive_int_query("hours", 24, max_value=168)
+    if error:
+        return error
     cutoff = time.time() - hours * 3600
 
     db = get_db()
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2003, opened 2026-09-02
fix(discover): make result cards keyboard accessible

### LINKED ISSUE #2002: Discover result cards are inaccessible to keyboard users
## Summary
The Discover page renders its dynamic agent, category, and video results as clickable `<div>` elements. These controls have no native interactive semantics, cannot receive keyboard focus, and provide no Enter/Space activation path, so keyboard-only users cannot open the same destinations/actions available to pointer users.

Affected render paths in `bottube_templates/discover.html`:
- `.agent-card` creator results
- category `.video-card` results
- `.video-card` video results

## Reproduction
1. Open `/discover` and load creator, category, or video results.
2. Use Tab/Shift+Tab without a pointing device.
3. Observe that the result cards are skipped entirely.
4. Inspect the DOM: navigation/actions are attached only through inline `onclick` handlers on non-focusable `<div>` elements.

## Expected
Every pointer-activatable Discover result is keyboard reachable, exposes the correct native/ARIA semantics, and activates exactly once with Enter (and Space where button semantics are appropriate).

## Impact
A core discovery path is unavailable to keyboard-only users. This is an ordinary accessibility/correctness defect; it does not involve authentication or security controls.

nwinkelman2:
## Summary
- replace mouse-only Discover creator/video result `<div>`s with native navigation links
- render category actions as native buttons so Enter and Space work without custom key handlers or double activation
- preserve card styling and add a visible `:focus-visible` treatment
- add focused regression coverage for semantics, labels, single activation path, and focus visibility

Fixes #2002

## Proof
- mutation baseline against `origin/main`: semantic-card contract fails
- `python -m pytest -q tests/test_discover_keyboard_cards.py tests/test_discover_agent_card_escaping.py tests/test_discover_thumbnail_urls.py` — **11 passed**
- `python -m pytest -q tests/test_accessibility.py tests/test_aria_labels.py` — **31 passed, 1 unrelated pre-existing failure** (`explore.html:126`, untouched by this PR)
- `git diff --check` — clean

### DIFF
--- a/bottube_templates/discover.html
+++ b/bottube_templates/discover.html
@@ -104,17 +104,34 @@
     }
 
     .video-card {
+        display: block;
+        width: 100%;
         background: var(--bg-secondary);
         border: 1px solid var(--border);
         border-radius: 12px;
         overflow: hidden;
         cursor: pointer;
+        color: inherit;
+        font: inherit;
+        text-align: left;
+        text-decoration: none;
         transition: transform 0.2s, box-shadow 0.2s;
     }
 
     .video-card:hover {
         transform: translateY(-2px);
         box-shadow: 0 4px 12px rgba(0,0,0,0.3);
+    }
+
+    .video-card:focus-visible,
+    .agent-card:focus-visible {
+        outline: 3px solid var(--accent);
+        outline-offset: 3px;
+    }
+
+    .category-card {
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2005, opened 2026-09-02
fix(generate): make completed history results accessible

### LINKED ISSUE #2004: Completed generation history results are mouse-only controls
## Summary
Completed rows in the Gemini generation history are rendered as `<div>` elements whose only activation path is an inline `onclick` calling `window.open(...)`. They are skipped by keyboard navigation and expose no link semantics. Non-completed rows use the same interactive-looking card styling even though clicking them does nothing.

## Reproduction
1. Open the generation page with at least one completed Gemini job and one pending/failed job.
2. Tab through the page without a pointing device.
3. Observe that the completed history result cannot receive focus or be opened with Enter.
4. Inspect the row: navigation is attached only to a non-focusable `<div onclick=...>`.

## Expected
- Completed jobs are native, keyboard-focusable links that safely open the job result in a new tab.
- Incomplete jobs are truthful non-interactive status records.
- Existing visual presentation is preserved and focused completed rows have a visible focus indicator.

## Impact
Keyboard-only users cannot open completed generation results, and the UI does not communicate which history rows are actionable. This is an ordinary accessibility/correctness defect.

nwinkelman2:
## Summary
- render completed generation-history jobs as native links with safe new-tab behavior
- leave pending/generating/failed rows as truthful non-interactive status records
- preserve existing visual styling and add visible focus treatment to actionable rows
- encode job IDs and add focused regression coverage for both semantic branches

Fixes #2004

## Proof
- mutation baseline against `origin/main`: native-link/non-action contract fails
- `python -m pytest -q tests/test_generation_history_keyboard.py` — **3 passed**
- combined accessibility run — **34 passed, 1 unrelated pre-existing failure** (`explore.html:126`, untouched)
- `git diff --check` — clean

### DIFF
--- a/bottube_templates/generate.html
+++ b/bottube_templates/generate.html
@@ -130,6 +130,17 @@
         padding: 10px 14px; background: var(--bg-secondary);
         border-radius: 6px; margin-bottom: 6px;
         border: 1px solid var(--border);
+        color: inherit; text-decoration: none;
+    }
+    a.gen-history-item {
+        cursor: pointer;
+    }
+    a.gen-history-item:hover {
+        border-color: var(--accent);
+    }
+    a.gen-history-item:focus-visible {
+        outline: 3px solid var(--accent);
+        outline-offset: 2px;
     }
     .gen-history-item .prompt-text {
         flex: 1; color: var(--text-primary); font-size: 0.9rem;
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2008, opened 2026-09-02
fix(activity): use native video links

### LINKED ISSUE #2006: Activity-feed video cards are mouse-only controls
## Problem

Upload entries in the activity feed render their video destination as a `<div onclick=...>`. The control is not exposed as a link, cannot receive keyboard focus, and cannot be activated with Enter, so keyboard and assistive-technology users cannot open the uploaded video from the feed.

## Reproduction

1. Open `/activity`.
2. Tab through an upload entry.
3. Observe that the video card is skipped; only a mouse click can open it.

## Expected

The uploaded-video destination should be a native link with the same card styling and a visible focus state. Non-destination status rows should remain non-interactive.

## Proposed fix

Render upload cards as native links and add a focused regression test that rejects the current mouse-only markup.

nwinkelman2:
## Summary
- render activity-feed upload destinations as native links instead of mouse-only clickable divs
- safely encode video IDs in watch URLs
- preserve card styling and add an explicit keyboard focus indicator
- add a regression covering native semantics, encoding, focus visibility, and removal of the click-only handler

## Verification
- `python -m pytest tests/test_accessibility.py -q`
- 26 passed, 10 subtests passed
- `git diff --check`

Closes #2006

### DIFF
--- a/bottube_templates/activity_feed.html
+++ b/bottube_templates/activity_feed.html
@@ -111,6 +111,16 @@
         align-items: center;
     }
 
+    a.activity-video {
+        color: inherit;
+        text-decoration: none;
+    }
+
+    a.activity-video:focus-visible {
+        outline: 3px solid var(--accent);
+        outline-offset: 2px;
+    }
+
     .video-thumb {
         width: 80px;
         height: 45px;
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2010, opened 2026-09-02
fix(activity): synchronize feed tab state

### LINKED ISSUE #2009: Activity-feed filter state depends on browser-global event
## Problem

The Activity Feed's `loadFeed()` clears the active class from every filter tab, then tries to restore it through the browser-global `event.target`. Initial loading and the 30-second polling path have no click event, and browser-global event behavior is not portable, so the visible tab state can disappear or follow the wrong control. The controls also expose no tab-list/selected semantics to assistive technology.

## Reproduction

1. Open `/activity` and let the initial load run.
2. Observe all `.feed-tab` controls after `loadFeed('all')` clears their state.
3. Wait for automatic polling or invoke `loadFeed()` without a click event; there is no reliable target from which to restore the selected tab.

## Expected

Filter state should be derived from the explicit `filter` argument, not ambient global state. Exactly one tab should remain selected, with `aria-selected` synchronized for every user click and background refresh.

nwinkelman2:
## Summary
- derive the selected Activity Feed filter from the explicit `filter` argument instead of browser-global `event`
- keep visual active state, `aria-selected`, and roving tab focus synchronized during clicks, initial load, and background polling
- expose the filter controls as a named tab list
- add a focused regression for the state contract

## Verification
- `python -m pytest tests/test_accessibility.py::TestAccessibilityAttributes::test_activity_feed_tabs_derive_state_from_filter -q`
- 1 passed
- `git diff --check`

Closes #2009

### DIFF
--- a/bottube_templates/activity_feed.html
+++ b/bottube_templates/activity_feed.html
@@ -181,11 +181,11 @@
         <p>Real-time updates from AI agents across BoTTube</p>
     </div>
 
-    <div class="feed-tabs">
-        <button class="feed-tab active" onclick="loadFeed('all')">All Activity</button>
-        <button class="feed-tab" onclick="loadFeed('uploads')">📤 Uploads</button>
-        <button class="feed-tab" onclick="loadFeed('comments')">💬 Comments</button>
-        <button class="feed-tab" onclick="loadFeed('tips')">💰 Tips</button>
+    <div class="feed-tabs" role="tablist" aria-label="Activity type">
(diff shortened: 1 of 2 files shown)


Scottcjn:

nwinkelman2:
Rebased against the current main branch. The requested behavior and regression coverage are already present upstream, leaving this PR with no distinct diff. Closing the redundant branch rather than preserving a stale/conflicting duplicate.

PR #2012, opened 2026-09-02
fix(receipts): reject invalid batch limits

### LINKED ISSUE #2007: Receipt batch endpoint silently rewrites invalid limit values
## Summary
`GET /api/agents/<agent_name>/receipts.zip` silently converts malformed `limit` values to the default and clamps zero/negative values to one. A client typo such as `?limit=abc`, `?limit=0`, or `?limit=-5` therefore returns HTTP 200 and generates a ZIP instead of identifying the invalid request.

The route currently catches `Exception` around `int(...)` and assigns `limit = 200`, while `max(1, ...)` silently rewrites non-positive values.

## Reproduction
For an existing agent with anchored receipts, request:

- `/api/agents/example/receipts.zip?limit=abc`
- `/api/agents/example/receipts.zip?limit=0`
- `/api/agents/example/receipts.zip?limit=-5`

All are accepted. The malformed value becomes 200; zero and negative values become 1.

## Expected
Malformed and out-of-range pagination values return a stable HTTP 400 JSON response before ZIP generation. Omitted `limit` keeps its documented default, and valid values through the existing batch maximum continue to work.

## Impact
Clients cannot distinguish an invalid request from a successful batch and may download a much larger archive than requested. This is an ordinary API input-validation defect.

nwinkelman2:
## Summary
- validate receipt-batch `limit` before rate-limit, provenance, database, or ZIP work
- reject malformed, non-positive, and over-cap limits with stable HTTP 400 JSON responses
- preserve the default of 200 and valid range 1–500
- add regression coverage proving invalid inputs do not cross the receipt side-effect edge

Fixes #2007

## Proof
- mutation baseline against `origin/main`: strict pagination contract fails (broad fallback/clamping present)
- `python -m pytest -q tests/test_receipt_zip_limit_validation.py` — **1 passed** (8 request cases)
- `python -m pytest -q tests/test_receipt_watch_url.py` — **1 passed**
- `git diff --check` — clean

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -24396,6 +24396,15 @@ def api_agent_receipts_zip(agent_name):
     if not re.fullmatch(r"[A-Za-z0-9_.\-]{1,64}", agent_name or ""):
         return jsonify({"ok": False, "error": "invalid agent name"}), 400
 
+    limit, error = _parse_positive_int_query(
+        "limit",
+        200,
+        min_value=1,
+        max_value=_RECEIPTS_BATCH_MAX,
+    )
+    if error:
+        return error
+
     ip = _get_client_ip()
     if not _rate_limit(f"receipts_zip:{ip}", 6, 600):
         return jsonify({"ok": False, "error": "rate limited"}), 429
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2016, opened 2026-09-02
fix(tips): validate pagination parameters

### LINKED ISSUE #2014: Video tips API silently rewrites malformed pagination
## Summary
The public `GET /api/videos/<video_id>/tips` endpoint silently rewrites malformed and non-positive `page` / `per_page` parameters through Flask's `type=int` defaulting and `max(...)` clamping.

Examples on current main for an existing video:

- `?page=abc` behaves as `page=1`
- `?page=0` and `?page=-3` behave as `page=1`
- `?per_page=abc` behaves as `per_page=10`
- `?per_page=0` and `?per_page=-3` behave as `per_page=1`
- `?per_page=999` behaves as `per_page=50`

## Expected
Malformed and out-of-range pagination values return stable HTTP 400 JSON responses. Omitted parameters preserve the existing defaults (`page=1`, `per_page=10`), and valid bounds (`page>=1`, `1<=per_page<=50`) remain unchanged.

## Impact
Clients cannot detect bad pagination requests and may repeatedly retrieve the wrong page or unexpected batch size. This is distinct from the already-merged large-offset fix, which only rejects a successfully parsed page after offset arithmetic overflows SQLite's integer range.

## Scope
Ordinary public API input validation only; no payment mutation or security behavior is involved.

nwinkelman2:
## Summary
- reject malformed/non-positive page and per_page values before video lookup or tip synchronization
- reject per_page values above the existing 50-row bound instead of silently clamping
- preserve page/per_page defaults and valid bounds
- retain the existing SQLite offset-overflow guard

Fixes #2014

## Proof
- mutation baseline against `origin/main`: strict pagination contract fails (Flask type-defaulting and bounds clamping present)
- new 8-case route regression — **passed**
- existing default/empty-totals route regression — **passed**
- `git diff --check` — clean

### DIFF
--- a/tests/test_video_tips_api.py
+++ b/tests/test_video_tips_api.py
@@ -9,6 +9,35 @@ def test_video_tips_returns_404_for_unknown_video(client):
     assert response.get_json() == {"error": "Video not found"}
 
 
+def test_video_tips_rejects_invalid_pagination_before_lookup(client):
+    cases = [
+        ({"page": "abc"}, "page must be an integer"),
+        ({"page": "0"}, "page must be >= 1"),
+        ({"page": "-3"}, "page must be >= 1"),
+        ({"per_page": "abc"}, "per_page must be an integer"),
+        ({"per_page": "0"}, "per_page must be >= 1"),
+        ({"per_page": "-3"}, "per_page must be >= 1"),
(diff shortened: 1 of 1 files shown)


Scottcjn:

Scottcjn (approved):

PR #2022, opened 2026-09-02
Fix giveaway countdown timer accumulation

### LINKED ISSUE #2019: Giveaway countdown accumulates animation-frame loops
## Summary
The giveaway countdown starts both a recursive `requestAnimationFrame(tick)` loop and a `setInterval(tick, 1000)` loop. The recursive path updates four DOM nodes on every display frame even though the values only change once per second, and every interval tick starts another persistent animation-frame chain.

## Reproduction
1. Open `/giveaway` while the giveaway start/end timestamp is in the future.
2. Instrument or count calls to `tick()` / `requestAnimationFrame`.
3. The initial call creates one ~60Hz chain.
4. Each one-second interval callback calls `tick()` again and creates another independent ~60Hz chain.
5. The number of active chains grows every second until the target time.

## Expected
The countdown updates once immediately and then once per second, with one timer that is cleared when the target is reached. No animation-frame loop is needed for whole-second display values.

## Impact
A page left open accumulates redundant callbacks and DOM writes, consuming progressively more CPU/battery for no visual benefit. This is an ordinary client-side performance/correctness defect.

nwinkelman2:
## Summary
- render the giveaway countdown immediately with a single one-second interval
- tear down that interval when the countdown reaches zero
- remove recursive animation-frame scheduling so callback work remains bounded
- add an executable JavaScript regression covering initial render, timer ownership, zero teardown, and no recursive scheduling

## Verification
- mutation baseline reproduced on origin/main: recursive requestAnimationFrame plus setInterval and no teardown
- node tests/js/giveaway_countdown.test.cjs
- python -m pytest -q tests/test_giveaway_countdown_runtime.py (1 passed)
- git diff --check

Fixes #2019

### DIFF
--- a/bottube_templates/giveaway.html
+++ b/bottube_templates/giveaway.html
@@ -435,6 +435,7 @@
     var el = document.getElementById("countdown");
     if (!el) return;
     var target = parseFloat(el.dataset.target) * 1000;
+    var timerId;
     function tick() {
         var diff = Math.max(0, target - Date.now());
         var d = Math.floor(diff / 86400000);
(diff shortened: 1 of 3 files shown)


Scottcjn (approved):

PR #2023, opened 2026-09-02
fix(bridge): synchronize chain tab state

### LINKED ISSUE #2021: Bridge tab ARIA selection diverges from visible chain
## Problem

Bridge chain tabs initialize with `aria-selected`, but `switchChain()` only changes CSS classes. After selecting ERG, BTC, or another chain, assistive technology still hears wRTC selected and the newly active tab unselected. The function also discovers the selected tab by parsing `onclick.toString()`, which is brittle under CSP/tooling transformations.

## Reproduction

1. Open `/bridge/wrtc`.
2. Activate the ERG tab.
3. Inspect the tabs: ERG has the visual `active` class, but its `aria-selected` remains `false`; wRTC remains `true`.

## Expected

Selection should be derived from stable `data-chain` values, and every transition should synchronize the active class, `aria-selected`, and roving tab stop.

nwinkelman2:
## Summary
- identify bridge tabs through stable `data-chain` values instead of parsing inline-handler source
- synchronize visual active state, `aria-selected`, and the roving tab stop on every chain change
- preserve panel switching, URL replacement, and funnel tracking
- add a focused bridge accessibility regression

## Verification
- `python -m pytest tests/test_bridge_template_accessibility.py -q`
- 4 passed
- `git diff --check`

Closes #2021

### DIFF
--- a/bottube_templates/bridge.html
+++ b/bottube_templates/bridge.html
@@ -289,27 +289,27 @@
 
     <!-- Chain Tabs -->
     <div class="chain-tabs" role="tablist">
-        <button class="chain-tab active" onclick="switchChain('wrtc')" role="tab" aria-label="Select wRTC (Solana) bridge" aria-selected="true">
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2026, opened 2026-09-02
fix(collaboration): lock pending invite actions

### LINKED ISSUE #2025: Pending invite actions allow duplicate conflicting requests
## Problem

Pending collaboration invites leave both Accept and Decline enabled while the response request is in flight. Repeated clicks or a fast Accept/Decline sequence can dispatch multiple conflicting POST requests for the same invite, producing race-dependent UI and unnecessary server work.

## Reproduction

1. Open pending collaboration invites.
2. Confirm Accept and click it repeatedly before the request returns, or click Decline immediately afterward.
3. Observe multiple POST requests to the same invite endpoint.

## Expected

After confirmation, both response controls for that invite should become disabled and expose a busy state until the request fails or completes. A failed response should restore the controls so the user can retry.

nwinkelman2:
## Summary
- disable both response controls for one pending invite after confirmation
- expose the in-flight state with `aria-busy`
- restore both controls after provider or transport failure so a retry remains possible
- make response buttons explicitly non-submitting
- add a focused UI regression

## Verification
- `python -m pytest tests/test_collaboration_invite_ui.py -q`
- 1 passed
- `git diff --check`

Closes #2025

### DIFF
--- a/bottube_templates/collaboration_pending_invites.html
+++ b/bottube_templates/collaboration_pending_invites.html
@@ -247,10 +247,10 @@
 
             <div class="invite-actions">
                 {% if not invite.is_expired %}
-                <button class="btn btn-primary" onclick="respondToInvite('{{ invite.invite_id }}', 'accept')">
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2032, opened 2026-09-02
Keep repeated code copies byte-identical

### LINKED ISSUE #2029: Repeated code copy corrupts the copied snippet
## Problem

The /badges and /embed-guide copy helpers derive clipboard payloads from the entire .code-block text and remove literal button labels. After the first click changes the button label to Copied!, clicking again before the 1.5-second reset can leave an ied! prefix in the copied snippet. The clipboard no longer contains valid Markdown or HTML.

## Reproduction

1. Open /badges or /embed-guide.
2. Click any Copy button twice within 1.5 seconds.
3. Paste the clipboard.
4. The second value can start with ied! instead of the snippet itself.

## Expected

Clipboard extraction must exclude the control element entirely so its transient label can never become part of the payload. Repeated clicks should copy the exact same code and status timers must not race.

nwinkelman2:
## Summary
- derive badge and embed-guide clipboard payloads from cloned code blocks with copy controls removed
- keep transient Copied! labels out of copied Markdown and HTML
- cancel each prior label-reset timer before scheduling its replacement
- add executable regressions proving rapid repeated clicks copy identical text on both pages

## Verification
- mutation baseline reproduced against origin/main: button text is scraped into clipboard source
- node tests/js/badges_copy_code.test.cjs
- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_badges_copy_code_runtime.py (1 passed)
- staged diff check clean

Fixes #2029

### DIFF
--- a/bottube_templates/badges.html
+++ b/bottube_templates/badges.html
@@ -416,10 +416,19 @@ function showTab(btn, panelId) {
 
 function copyCode(btn) {
     var block = btn.parentElement;
-    var text = block.textContent.replace("Copy", "").trim();
+    var copy = block.cloneNode(true);
+    copy.querySelectorAll(".copy-btn").forEach(function(control) { control.remove(); });
+    var text = copy.textContent.trim();
+    if (btn._copyResetTimer !== undefined) {
+        clearTimeout(btn._copyResetTimer);
+        btn._copyResetTimer = undefined;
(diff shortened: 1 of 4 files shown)


Scottcjn (approved):

PR #2037, opened 2026-09-02
fix(channel): synchronize subscribe semantics

### LINKED ISSUE #2035: Channel subscribe accessibility state remains stale after toggle
## Problem

The channel Subscribe/Following button is server-rendered with correct `aria-pressed` and an action-specific label, but `toggleSubscribe()` updates only its text and CSS class. After a successful subscribe or unfollow action, the visible state changes while assistive technology continues to receive the stale pressed state and stale action label.

## Reproduction

1. Open another agent's channel while signed in.
2. Activate Subscribe.
3. Observe visible text changes to Following, but `aria-pressed` remains `false` and the label still says Subscribe.
4. Toggle again; the inverse stale state occurs.

## Expected

Every successful response should synchronize text, CSS class, `aria-pressed`, and the next-action accessible label from the server-returned `following` value.

nwinkelman2:
## Summary
- synchronize `aria-pressed` with the server-returned following state
- update the action-specific accessible label after subscribe and unfollow
- preserve current text, class, subscriber-count, login, and retry behavior
- add a focused regression for the full visible/accessibility state contract

## Verification
- `python -m pytest tests/test_channel_subscribe_accessibility.py -q`
- 1 passed
- `git diff --check`

Closes #2035

### DIFF
--- a/bottube_templates/channel.html
+++ b/bottube_templates/channel.html
@@ -707,6 +707,7 @@
         {% if current_user and current_user.agent_name != agent.agent_name %}
         <button id="subscribe-btn" class="subscribe-btn {{ 'following' if is_following else 'not-following' }}"
                 onclick="toggleSubscribe()" type="button"
+                data-agent-label="{{ agent.display_name or agent.agent_name }}"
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2042, opened 2026-09-02
Synchronize RTC tip preset selection

### LINKED ISSUE #2040: Custom RTC tip leaves a stale preset selected
## Problem

The watch-page tip presets only toggle a visual .selected class. After choosing a preset and then typing a different custom amount, the old preset remains highlighted even though a different amount will be submitted. The buttons also expose no programmatic pressed state.

## Reproduction

1. Open the RTC tip panel on a watch page.
2. Select the 0.10 preset.
3. Change the custom amount input to 0.25.
4. The 0.10 button still appears selected while the form will submit 0.25.

## Expected

Preset selection should expose an exactly-one ria-pressed state. Editing the custom amount should clear every preset's visual and programmatic selection so displayed state matches the submitted value.

nwinkelman2:
## Summary
- expose RTC tip presets as a labelled pressed-state group
- keep exactly one visual and programmatic preset selected
- clear all preset state as soon as a custom amount is entered without changing that amount
- add static input wiring and executable selection-transition regressions

## Verification
- mutation baseline reproduced against origin/main: preset visual state remained stale after custom input
- node tests/js/tip_preset_state.test.cjs
- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_tip_preset_state_runtime.py (2 passed)
- staged diff check clean

Fixes #2040

### DIFF
--- a/bottube_templates/watch.html
+++ b/bottube_templates/watch.html
@@ -2569,16 +2569,16 @@
                 </div>
 
                 <div class="tip-panel" id="tip-panel">
-                    <div class="tip-amounts">
-                        <button class="tip-amount-btn" aria-label="Send 0.01 RTC tip" onclick="selectTipAmount(0.01)">0.01</button>
-                        <button class="tip-amount-btn" aria-label="Send 0.05 RTC tip" onclick="selectTipAmount(0.05)">0.05</button>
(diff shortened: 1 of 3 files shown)


Scottcjn (approved):

PR #2044, opened 2026-09-02
fix(analytics): expose selected ranking metric

### LINKED ISSUE #2041: Analytics ranking controls expose no selected metric
## Problem

Analytics Top Videos offers three mutually exclusive sorting controls (Views, Engagement, Tips), but only changes an `active` CSS class. The group has no accessible name and the controls expose no selected/pressed state, so screen-reader users cannot tell which ranking is currently applied.

## Reproduction

1. Open `/analytics`.
2. Activate By Engagement or By Tips.
3. Observe the visual active style changes and the table reloads, but the controls expose no programmatic current state.

## Expected

The controls should be a named group of toggle buttons. Exactly one button should expose `aria-pressed="true"`, synchronized from the same metric selection that reloads the table.

nwinkelman2:
## Summary
- expose the Top Videos ranking controls as a named toggle-button group
- synchronize the visual active class and `aria-pressed` from one selection decision
- preserve existing metric loading behavior
- add a focused regression proving exactly one selected metric

## Verification
- `python -m pytest tests/test_analytics_metric_accessibility.py -q`
- 1 passed
- `git diff --check`

Closes #2041

### DIFF
--- a/bottube_templates/analytics.html
+++ b/bottube_templates/analytics.html
@@ -291,10 +291,10 @@
     <div class="analytics-section">
         <div class="section-header">
             <h2>🏆 Top Videos</h2>
-            <div class="metric-tabs">
-                <button class="metric-tab active" data-metric="views">By Views</button>
-                <button class="metric-tab" data-metric="engagement">By Engagement</button>
-                <button class="metric-tab" data-metric="tips">By Tips</button>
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2045, opened 2026-09-02
Validate admin list pagination

### LINKED ISSUE #2043: Admin referral and badge lists silently coerce invalid pagination
## Problem

/api/admin/referrals and /api/admin/badges parse page/per_page with Flask's typed default and then clamp. Malformed values such as page=abc are silently rewritten to defaults, while non-positive or over-limit values are silently changed. This hides client defects and makes the echoed pagination contract disagree with the request.

## Reproduction

With a valid admin key, request either endpoint with page=abc, page=0, per_page=bad, or per_page=101. Each request succeeds with a substituted/clamped value instead of returning a controlled validation error.

## Expected

Both admin list endpoints should use the server's canonical strict integer query parser, reject malformed/non-positive/over-limit values with HTTP 400 before database work, and preserve their documented defaults and valid ranges.

nwinkelman2:
## Summary
- route referral and badge admin pagination through the canonical strict integer parser
- reject malformed, non-positive, over-cap per-page, and excessively deep page values before database access
- preserve omitted defaults and valid pagination values
- add 16 route regressions covering both endpoints and the no-database-work failure edge

## Verification
- mutation baseline reproduced against origin/main: both routes default/clamp Flask typed query values
- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_badges.py -k 'admin_badges_rejects_invalid_pagination_before_db or admin_badges_preserves_valid_pagination' (8 passed)
- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_referrals.py -k 'admin_referrals_rejects_invalid_pagination_before_db or admin_referrals_preserves_valid_pagination' (8 passed)
- python -m py_compile bottube_server.py tests/test_badges.py tests/test_referrals.py
- staged diff check clean

Fixes #2043

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -5864,12 +5864,17 @@ def admin_referrals():
     if err:
         return err
 
+    page, error = _parse_positive_int_query("page", 1, max_value=10000)
+    if error:
+        return error
+    per_page, error = _parse_positive_int_query("per_page", 20, max_value=100)
+    if error:
+        return error
+
     db = get_db()
     status_filter = (request.args.get("status", "") or "").strip().lower()
     track_filter = (request.args.get("track", "") or "").strip().lower()
     ref_code = _normalize_ref_code(request.args.get("code", ""))
-    page = max(1, request.args.get("page", 1, type=int))
-    per_page = min(100, max(1, request.args.get("per_page", 20, type=int)))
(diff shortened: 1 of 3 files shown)


Scottcjn (approved):

PR #2051, opened 2026-09-02
fix(translations): label search and isolate new tabs

### LINKED ISSUE #2050: Translations search lacks a persistent label and safe new-tab contract
## Problem

On the video-translations page, the search control relies on placeholder text as its only name. Placeholder text is not a persistent programmatic label, so screen-reader and voice-control users cannot reliably identify the field.

Each generated video link also opens a new tab with `target="_blank"` but omits `rel="noopener noreferrer"`, unlike a bounded safe-new-tab contract.

## Reproduction

1. Render `templates/translations.html` with at least one translation.
2. Inspect `#searchInput`: there is no associated `<label>`, `aria-label`, or `aria-labelledby`.
3. Inspect the result link: it has `target="_blank"` without a protective `rel` value.

## Expected

The search field has a persistent accessible name while preserving the existing visual layout, and generated video links retain their current new-tab behavior with explicit opener/referrer isolation.

## Scope

A focused template change plus executable regression coverage for both contracts. No route or data-model behavior changes.

nwinkelman2:
## Summary
- give the translations search field a persistent programmatic name without changing its visual layout
- preserve new-tab video navigation while isolating opener and referrer context
- add focused executable regression coverage for both template contracts

## Verification
- mutation baseline against `origin/main`: both the accessible name and safe-new-tab relation were absent
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_translations_template_accessibility.py -q` (2 passed)
- staged diff check clean

Fixes #2050

### DIFF
--- a/templates/translations.html
+++ b/templates/translations.html
@@ -93,14 +93,14 @@
     </div>
 
     <div class="search-box">
-        <input type="text" placeholder="Search translations by language or content..." id="searchInput">
+        <input type="text" placeholder="Search translations by language or content..." id="searchInput" aria-label="Search translations">
     </div>
 
     <div id="translationsContainer">
         {% for translation in translations %}
         <div class="translation-card">
             <div class="language-tag">{{ translation.language }}</div>
-            <p><strong>Video:</strong> <a href="{{ translation.video_url }}" class="video-link" target="_blank">{{ translation.video_url }}</a></p>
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2052, opened 2026-09-02
feat: restore search suggestions API

### LINKED ISSUE #1998: Search suggestions contract is shipped but the endpoint returns 404

nwinkelman2:
Closes #1998.

- restores the shipped `/api/search/suggestions` contract
- returns bounded title, category, creator, and tag suggestions
- filters removed videos and banned creators
- avoids catalog scans for empty/one-character queries
- repairs the existing multipart-invalid regression setup and adds the short-query gate

Verification: focused suggestion tests passed (2 passed); `python -m py_compile bottube_server.py`; `git diff --check`.

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -8547,6 +8547,76 @@ def web_subscribe(agent_name):
 # Search
 # ---------------------------------------------------------------------------
 
+@app.route("/api/search/suggestions")
+def search_suggestions():
+    """Return bounded public-catalog suggestions for a partial query."""
+    q = request.args.get("q", "").strip()
+    empty = {"query": q, "suggestions": [], "categories": [], "agents": [], "tags": []}
+    if len(q) < 2:
+        return jsonify(empty)
+
+    db = get_db()
+    like_q = f"%{q}%"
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2054, opened 2026-09-02
fix(grazer): isolate external resource tabs

### LINKED ISSUE #2053: Grazer resource links omit safe new-tab relation
## Problem

The Grazer page opens its GitHub, npm, and PyPI resource links in new tabs, but none of the three links declares `rel="noopener noreferrer"`.

## Reproduction

1. Render `bottube_templates/grazer.html`.
2. Inspect the three resource anchors under the command examples.
3. Each uses `target="_blank"` without an explicit opener/referrer isolation contract.

## Expected

Keep the same destinations, styling, and new-tab behavior while adding `rel="noopener noreferrer"` to every external resource link. Add focused regression coverage so future resources cannot silently regress.

## Scope

Template-only navigation hardening and an executable test; no route or API changes.

nwinkelman2:
## Summary
- preserve all three Grazer resource destinations and their existing new-tab behavior
- add explicit opener and referrer isolation to the GitHub, npm, and PyPI links
- add a focused regression that covers the complete resource-link set

## Verification
- mutation baseline against `origin/main`: 3 of 3 links omitted the safe relation
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_grazer_template_navigation.py -q` (1 passed)
- staged diff check clean

Fixes #2053

### DIFF
--- a/bottube_templates/grazer.html
+++ b/bottube_templates/grazer.html
@@ -96,17 +96,17 @@
 </div>
 
 <div style="text-align: center; padding: 40px 0;">
-    <a href="https://github.com/Scottcjn/grazer-skill" target="_blank"
+    <a href="https://github.com/Scottcjn/grazer-skill" target="_blank" rel="noopener noreferrer"
        style="display: inline-block; padding: 12px 32px; background: var(--accent); color: #000;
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2057, opened 2026-09-02
fix(chat): expose messages as a live log

### LINKED ISSUE #2055: Live chat messages are not exposed as an accessible log
## Problem

The standalone live-chat page appends chat and system messages to `#messages`, but that container has no log/live-region semantics. New messages therefore are not reliably announced to screen-reader users. The composer input also relies on placeholder text as its only name.

## Reproduction

1. Render `templates/chat.html` and connect to a live stream.
2. Inspect `#messages`: it is a generic `<div>` despite receiving ordered appended messages.
3. Inspect `#msg-input`: it has no associated label or ARIA name.
4. Receive a message or system event; the DOM changes but exposes no live-region contract.

## Expected

Expose the message stream as a polite append-only log with additions announced, and give the composer a persistent accessible name. Preserve existing socket events, visual layout, scrolling, and send behavior.

## Scope

Template semantics plus focused executable regression coverage; no socket or payment behavior changes.

nwinkelman2:
Summary:
- expose appended chat and system messages through a polite ARIA log
- give the message composer a persistent accessible name
- add focused template-contract regressions

Verification:
- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_chat_template_accessibility.py -q (2 passed)
- git diff --check

Fixes #2055

### DIFF
--- a/templates/chat.html
+++ b/templates/chat.html
@@ -42,7 +42,7 @@
             <span>Live Chat</span>
             <span class="viewers" id="viewer-count">0 watching</span>
         </div>
-        <div class="chat-messages" id="messages"></div>
+        <div class="chat-messages" id="messages" role="log" aria-live="polite" aria-relevant="additions text"></div>
         {% if is_mod %}
         <div class="mod-bar">
             <button onclick="toggleSlowMode()">Slow Mode</button>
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2064, opened 2026-09-02
fix(trust-safety): support fresh database initialization

### LINKED ISSUE #2059: [BUG] Trust-and-safety schema migration crashes on a fresh configured database
## Summary

The trust-and-safety schema initializer assumes the core gents table already exists. With a fresh database path, an authenticated request to a trust-and-safety admin endpoint fails during lazy initialization with sqlite3.OperationalError: no such table: agents instead of reaching the endpoint's bounded validation/response path.

## Reproduction

1. Set BOTTUBE_DB_PATH to a new, empty file path.
2. Import ottube_server with a valid BOTTUBE_ADMIN_KEY.
3. POST an empty JSON object to /admin/blocklist/add with that key.
4. Observe the request crash inside _ensure_ts_schema() while adding TOS columns to the absent gents table.

Existing regression coverage reproduces it:

`powershell
='1'
python -m pytest -q tests/test_admin_key_env.py
` 

## Impact

Fresh/test deployments cannot use the lazy trust-and-safety initializer independently, and the authenticated gate returns an internal error rather than the expected bounded validation response.

## Suggested fix

Keep creating the trust-and-safety tables unconditionally, but only apply the optional TOS column migration when the canonical gents table exists. Preserve idempotence and cover both ephemeral and explicit admin-key paths.

## Funding

Reported for the ongoing BoTTube bug pool: https://github.com/Scottcjn/rustchain-bounties/issues/71

nwinkelman2:
## Summary
- keep trust-and-safety table creation usable before the core agents table exists
- retry the agent TOS/suspension column migration once core schema is available
- cover a complete valid blocklist write on a fresh configured database

## Verification
- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_admin_key_env.py tests/test_malformed_json_admin.py — 14 passed
- python -m py_compile bottube_server.py
- git diff --check

Closes #2059

Funded-program context: Scottcjn/rustchain-bounties#71.

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -20153,24 +20153,30 @@ def _ensure_ts_schema():
                 """
             )
 
-            # Add TOS columns to agents (idempotent).
+            # Add TOS columns to agents (idempotent).  Trust-and-safety tables
+            # are also used by fresh/test deployments where the core schema
+            # may not have been initialized yet.  PRAGMA returns no columns
+            # for an absent table, so do not try to ALTER it in that state.
+            # Leave the combined readiness flag false so a later call retries
+            # the agent-column migration after the core schema is created.
             cols = {row[1] for row in conn.execute("PRAGMA table_info(agents)").fetchall()}
-            if "tos_version_accepted" not in cols:
-                conn.execute("ALTER TABLE agents ADD COLUMN tos_version_accepted TEXT DEFAULT ''")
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2065, opened 2026-09-02
fix(admin): validate moderation queue pagination

### LINKED ISSUE #2062: Admin moderation queues silently rewrite invalid pagination
### Problem

The three admin queue endpoints /api/admin/reports, /api/admin/reward-holds, and /api/admin/moderation-holds parse page and per_page with 
equest.args.get(..., type=int) and immediately clamp the result. Werkzeug returns each default when conversion fails, so inputs such as page=abc or per_page= are silently served as valid page 1 / default-size requests.

The same paths also accept oversized page values until SQLite receives an out-of-range OFFSET, producing a server error instead of a deterministic client response.

### Expected

All three related admin queues should reuse the canonical positive-integer parser, reject malformed/non-positive/out-of-bounds values with stable HTTP 400 JSON, preserve omitted defaults and valid bounds, and perform no database query for invalid input.

### Scope

This is one shared admin-queue pagination contract, not separate claims per endpoint. Add focused regression coverage for each endpoint and representative invalid values.

nwinkelman2:
Summary:
- apply the canonical bounded positive-integer parser to reports, reward holds, and moderation holds
- reject malformed, non-positive, route-over-cap, and overflow pagination before database work
- preserve omitted defaults and each route's existing valid maximum

Verification:
- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_admin_queue_pagination_validation.py -q (1 passed; all endpoint/value cases covered in one fixture)
- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_quests.py::test_report_threshold_queues_hold_without_auto_removal tests/test_quests.py::test_admin_resolve_report_defaults_to_coach_without_deleting_comment tests/test_quests.py::test_like_reward_hold_can_be_credited_by_admin tests/test_quests.py::test_moderation_hold_release_restores_video -q (4 passed)
- git diff --check

Fixes #2062

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -17620,11 +17620,15 @@ def admin_reports():
     if not ADMIN_KEY or admin_key != ADMIN_KEY:
         return jsonify({"error": "Unauthorized"}), 401
 
-    db = get_db()
     status_filter = request.args.get("status", "pending")
-    page = max(1, request.args.get("page", 1, type=int))
-    per_page = min(50, max(1, request.args.get("per_page", 20, type=int)))
+    page, error = _parse_positive_int_query("page", 1, max_value=10000)
+    if error:
+        return error
+    per_page, error = _parse_positive_int_query("per_page", 20, max_value=50)
+    if error:
+        return error
     offset = (page - 1) * per_page
+    db = get_db()
 
     rows = db.execute(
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2070, opened 2026-09-02
fix(wallet): announce operation results

### LINKED ISSUE #2069: Wallet operation results are visual-only and not announced
### Problem

/settings/wallet updates three status/result containers after async operations: #linked-wallet-result, #local-wallet-status, and #local-wallet-result. They are generic empty div elements with no live-region/status semantics.

Saving a receiving wallet, generating/importing/clearing a local wallet, loading wallet state, or encountering an API/network/runtime error changes only visible text. Screen-reader users receive no operation completion or failure feedback while focus stays on the triggering button.

### Expected

Expose the three persistent result containers as atomic polite statuses without changing the current visual layout or wallet behavior. Add focused regression coverage for every status container and for success/error text written through the shared setText path.

nwinkelman2:
Summary:
- expose linked-wallet, local-wallet state, and local-wallet action feedback as persistent atomic polite statuses
- preserve the existing layout and wallet behavior
- exercise every status container and the shared result writer

Verification:
- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_wallet_status_live_region.py -q (2 passed)
- git diff --check

Fixes #2069

### DIFF
--- a/bottube_templates/settings_wallet.html
+++ b/bottube_templates/settings_wallet.html
@@ -20,7 +20,7 @@
               style="background:var(--accent);color:var(--bg-primary);border:none;border-radius:8px;padding:10px 14px;font-weight:800;cursor:pointer;"
               onclick="saveLinkedWallet()" aria-label="Save linked wallet address">Save</button>
     </div>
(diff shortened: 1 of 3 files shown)


Scottcjn (approved):

PR #2077, opened 2026-09-02
fix(upload): preserve HTTP error semantics

### LINKED ISSUE #2074: Anonymous upload hides non-JSON HTTP failures
### Problem

The API-key upload form unconditionally calls 
esponse.json() before checking HTTP status. Reverse proxies and upload limits commonly return HTML or empty bodies for failures such as 413/502. In those cases JSON parsing rejects and the UI reports only an Unexpected token/parse error, losing the actual HTTP status and any usable response context.

### Reproduction

Mock /api/upload to return 413 Payload Too Large with a non-JSON body and submit the logged-out upload form. The catch path displays a JSON parser failure rather than HTTP 413, even though the response itself is valid evidence.

### Expected

Parse JSON on a best-effort basis, retain 
esponse.ok and 
esponse.status as the transport truth, surface provider JSON errors when present and a stable HTTP fallback otherwise, and restore the submit button on every failure. Preserve the successful redirect behavior. Add executable regression coverage for JSON failure, non-JSON HTTP failure, and success.

nwinkelman2:
Summary:
- parse upload JSON on a best-effort basis while retaining transport ok/status truth
- prefer provider JSON errors and fall back to stable HTTP status messages for HTML/empty failures
- preserve successful redirects and restore the submit button on all failure paths

Verification:
- PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest tests/test_upload_http_response.py -q (1 passed; success, JSON failure, and non-JSON HTTP failure covered)
- git diff --check

Fixes #2074

### DIFF
--- a/bottube_templates/upload.html
+++ b/bottube_templates/upload.html
@@ -468,12 +468,18 @@ document.querySelector('.upload-form').addEventListener('submit', function(e) {
         headers: { 'X-API-Key': apiKey },
         body: formData
     })
-    .then(function(r) { return r.json(); })
-    .then(function(data) {
-        if (data.ok) {
+    .then(function(r) {
+        return r.json().catch(function() { return {}; }).then(function(data) {
+            return { ok: r.ok, status: r.status, data: data };
+        });
+    })
+    .then(function(resp) {
+        var data = resp.data || {};
+        if (resp.ok && data.ok) {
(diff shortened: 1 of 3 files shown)


Scottcjn (approved):

PR #2078, opened 2026-09-02
fix(security): enable TLS verification by default in telegram_bot

xxzzzzy:
## Summary

`telegram_bot.py`'s `BoTTubeAPI` hard-coded `requests.Session.verify = False` because the default `BOTTUBE_BASE` historically pointed at a self-signed local IP (`50.28.86.153:8097`). The Telegram bot is a public, user-facing surface, so disabling TLS verification silently exposes every BoTTube API call (browse, search, agent profile, watch-time telemetry, /tip wallet lookup) to active on-path attackers.

## Fix

- Default to TLS verification on (`verify_ssl=True`).
- Accept an explicit `verify_ssl` argument on `BoTTubeAPI` for the legacy self-signed workflow.
- Honor the env opt-out `BOTTUBE_TG_INSECURE_SKIP_TLS_VERIFY` (truthy values: `1` / `true` / `yes` / `on`) so operators on the legacy self-signed deployment can disable verification without editing the file.

## Tests

- Added 6 unit tests in `tests/test_telegram_bot.py::TestBoTTubeAPI` covering: default-true, env opt-out truthy `1` and `yes`, env opt-out falsy, explicit `verify_ssl=False`, and explicit override of the env opt-out.
- All 24 tests in `tests/test_telegram_bot.py` pass (`pytest -q`).

## Compatibility

- Behaviour for end users on the default deployment is unchanged (the public `bottube.ai` URL already serves a valid certificate).
- Behaviour for the legacy self-signed local IP deployment is preserved via the `BOTTUBE_TG_INSECURE_SKIP_TLS_VERIFY` env var.

## Notes

- Defensive hardening only; no API contracts changed.
- Companion spirit to prior merged fixes that similarly tightened TLS defaults (`fix(security): default to HTTPS in health-check.py`, `fix(security): enable TLS verification by default in beacon_client.py`, etc.).

### DIFF
--- a/telegram_bot.py
+++ b/telegram_bot.py
@@ -104,11 +104,26 @@ def _escape(text: str) -> str:
 class BoTTubeAPI:
     """REST client for BoTTube."""
 
-    def __init__(self, base_url: str = BOTTUBE_BASE, timeout: int = 10):
-        """Set up a requests session against the BoTTube API (self-signed cert, TLS verify disabled)."""
+    def __init__(
+        self,
+        base_url: str = BOTTUBE_BASE,
+        timeout: int = 10,
+        verify_ssl: Optional[bool] = None,
+    ):
+        """Set up a requests session against the BoTTube API.
+
+        TLS verification is enabled by default. To preserve the previous
+        self-signed-cert workflow (NOT recommended for production), set the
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):
TLS verification on by default for the Telegram bot with an explicit env opt-out; tests updated.

PR #2081, opened 2026-09-02
fix(pi): validate authentication payloads

### LINKED ISSUE #2075: Pi authentication crashes on malformed JSON payloads
### Problem

`POST /pi/auth` assumes both the JSON body and `access_token` are string-shaped. A valid JSON array reaches `data.get(...)` and raises `AttributeError`; an object with a numeric token reaches `.strip()` and also raises. The authentication endpoint returns a server error instead of a bounded client response.

### Reproduction

- `POST /pi/auth` with a JSON array body
- `POST /pi/auth` with an object whose `access_token` is numeric

Both fail before any Pi Network request is made.

### Expected

Reject non-object JSON and non-string `access_token` with deterministic HTTP 400 JSON errors, while preserving the existing required-token and valid string authentication flow. Add focused regression coverage.

nwinkelman2:
## Summary
- validate `/pi/auth` request bodies as JSON objects before field access
- reject non-string `access_token` values with a deterministic 400 response
- cover malformed bodies, typed tokens, required-token behavior, and the valid Pi `/v2/me` flow

## Validation
- `python -m pytest tests/test_pi_auth_input_validation.py -q` (4 passed)
- `python -m py_compile bottube_server.py tests/test_pi_auth_input_validation.py`
- `git diff --check`

Fixes #2075

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -5252,8 +5252,13 @@ def pi_auth():
     find-or-create a BoTTube agents account keyed on pi_uid and establish a session.
     No Pi API key is required for this flow (mirrors /auth/google). Username scope
     only -- this never initiates a Pi payment."""
-    data = request.get_json(silent=True) or {}
-    access_token = (data.get("access_token") or "").strip()
+    data, body_error = _json_object_body()
+    if body_error:
+        return body_error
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2082, opened 2026-09-02
fix: preserve engineering data on refresh errors

### LINKED ISSUE #2080: Engineering dashboard silently accepts failed or incomplete refreshes
## Bug

The engineering dashboard's 30-second refresh currently calls `response.json()` for every response and silently swallows all failures. An HTTP 500 HTML response is therefore indistinguishable from a successful refresh to the operator, and a successful but incomplete JSON response can overwrite the generated timestamp with the literal text `undefined`.

## Reproduction

1. Open `/engineering`.
2. Have `/api/engineering` return a non-2xx HTML error page (or a JSON object without `generated_at`).
3. Let the countdown reach zero.

The countdown resets, the failed refresh is silent, and incomplete data may be rendered.

## Expected

- Non-2xx responses are rejected before JSON parsing.
- The existing dashboard data remains intact unless the response contains valid replacement fields.
- The refresh indicator reports failure accessibly and retries on the next bounded cycle.
- Tests cover HTTP failure, invalid payload, and a valid update.

## Scope

Ordinary dashboard correctness and error handling only; no API or authentication changes.

nwinkelman2:
## Summary

- reject non-2xx engineering refreshes before parsing their bodies
- validate the complete node/timestamp payload before replacing last-known-good data
- surface success and retryable failure through an accessible status region
- add Node-backed regressions for HTTP failure, incomplete JSON, and a valid update

## Verification

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_engineering_refresh.py` — 4 passed
- `git diff --check` — clean

Fixes #2080

### DIFF
--- a/bottube_templates/engineering.html
+++ b/bottube_templates/engineering.html
@@ -41,7 +41,7 @@
 
 <div class="eng-shell">
   <h1>BoTTube Engineering</h1>
-  <p class="eng-sub">Live operational visibility — provider health, request latency, queue depth, and active experiments. <span class="eng-refresh"><span class="spin-dot"></span><span id="eng-tick">refreshing every 30s</span></span></p>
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2083, opened 2026-09-02
fix(mood): validate history limits

### LINKED ISSUE #2079: Mood history silently rewrites invalid limits
`GET /api/v1/agents/<agent_name>/mood/history` silently rewrites malformed and out-of-range `limit` values. For example, `?limit=abc` becomes the default 20, `?limit=0` becomes 1, and `?limit=101` becomes 100. That hides client bugs and makes this endpoint inconsistent with the API's strict bounded-integer contract.

Expected: reject malformed, non-positive, and over-100 values with a stable HTTP 400 response before calling the mood-history engine, while preserving the default of 20 and valid range 1–100.

nwinkelman2:
Closes #2079

## What changed

- routes mood-history `limit` through the API's strict bounded-integer parser
- rejects malformed, fractional, non-positive, and over-100 values before mood-engine work
- preserves the default of 20 and valid range 1–100
- adds invalid-edge and valid-control regressions

## Verification

- `python -m pytest tests/test_mood_write_authorization.py -q` (16 passed)
- `python -m py_compile bottube_server.py tests/test_mood_write_authorization.py`
- `git diff --check` (clean)

RustChain payout address: `RTCe625bc2d5c25d4348236a4fcb74e5d0ef01f6d6d`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -7523,7 +7523,9 @@ def get_mood_history(agent_name):
     if not agent:
         return jsonify({"error": "Agent not found"}), 404
     
-    limit = min(100, max(1, request.args.get("limit", 20, type=int)))
+    limit, error = _parse_positive_int_query("limit", 20, max_value=100)
+    if error:
+        return error
     
     engine = get_mood_engine(str(DB_PATH))
     history = engine.get_mood_history(agent["id"], limit)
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2088, opened 2026-09-02
fix: keep Beacon Atlas available on metadata failure

### LINKED ISSUE #2085: Beacon Atlas discards valid graph when optional agent metadata fails
## Bug

Beacon Atlas treats the beacon directory and the supplemental agent-metadata API as one all-or-nothing request. Because the first two requests run in `Promise.all`, a temporary failure from `/api/agents` aborts the entire visualization even when `/api/beacon/directory` returned the canonical graph nodes successfully.

The same requests currently parse non-2xx HTML responses as JSON without checking HTTP status.

## Reproduction

1. Open `/beacon/atlas`.
2. Return 200 JSON from `/api/beacon/directory`.
3. Return 500 from `/api/agents?limit=100&sort=popular`.

The page displays `Failed to load data` and renders no graph, discarding the valid beacon directory.

## Expected

- The beacon directory remains the required source for graph nodes.
- Optional agent metadata enriches nodes when available but cannot take down the atlas.
- Non-2xx responses are rejected before JSON parsing.
- Regression coverage proves required-source failure remains terminal while optional-source failure degrades gracefully.

## Scope

Ordinary data-loading reliability only; no API, authentication, or graph-visualization behavior changes.

nwinkelman2:
## Summary

- keep the canonical Beacon directory as the required Atlas data source
- treat agent-list pages as optional enrichment so their outage cannot erase the graph
- reject non-2xx and invalid JSON payload shapes explicitly
- add executable regressions for required-source failure, graceful metadata degradation, and partial pagination

## Verification

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_beacon_atlas_data_loading.py tests/test_seo_audit_hygiene.py::test_template_and_beacon_atlas_assets_have_no_console_log` — 4 passed
- `git diff --check` — clean

Fixes #2085

### DIFF
--- a/bottube_templates/beacon_atlas.html
+++ b/bottube_templates/beacon_atlas.html
@@ -256,6 +256,49 @@
 </div>
 
 <script>
+async function _atlasFetchJson(url) {
+  const response = await fetch(url);
+  if (!response.ok) {
+    throw new Error(`Atlas request failed with HTTP ${response.status}`);
+  }
+  const payload = await response.json();
+  if (!payload || typeof payload !== 'object' || Array.isArray(payload)) {
+    throw new Error('Atlas request returned an invalid payload');
+  }
+  return payload;
+}
+
+async function _atlasLoadDirectoryData() {
+  const [beaconResult, agentResult] = await Promise.allSettled([
+    _atlasFetchJson('/api/beacon/directory'),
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2092, opened 2026-09-02
fix: preserve unread state when mark-all fails

### LINKED ISSUE #2090: Mark-all notification failure falsely clears unread state
## Bug

The global notification dropdown clears its unread badge after every response from `POST /api/notifications/read`, including HTTP 4xx/5xx failures. Its promise handler never checks `response.ok`, so an expired session, rejected CSRF token, or server failure is presented as a successful "mark all read" action even though unread state was not changed.

## Reproduction

1. Open the notification panel with unread notifications.
2. Have `POST /api/notifications/read` return HTTP 403 or 500.
3. Activate `Mark all as read`.

The UI immediately hides the unread badge and reloads the list as if the write succeeded.

## Expected

- Only a successful 2xx response may clear/reload unread state.
- Failed responses preserve the last-known unread state.
- Repeated activation is disabled while the write is in flight and restored afterward.
- A focused runtime regression proves success and failure semantics.

## Scope

Ordinary notification state consistency only; no endpoint, authentication, or CSRF changes.

nwinkelman2:
## Summary

- require a successful response before clearing or reloading notification state
- keep the last-known unread badge intact on HTTP failure
- prevent duplicate mark-all writes while one request is in flight
- add a Node-backed runtime regression for failure, success, and double activation

## Verification

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_notification_mark_all_ui.py tests/test_notifications.py` — 4 passed
- `git diff --check` — clean

Fixes #2090

### DIFF
--- a/bottube_static/base.js
+++ b/bottube_static/base.js
@@ -93,6 +93,7 @@
     if (!wrapper || !bell || !panel || !badge || !list || !markAll) return;
 
     var noneText = getMeta("bt-notif-none") || "No notifications";
+    var markAllInFlight = false;
 
     function setUnreadBadge(unread) {
       var count = Number(unread || 0);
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2094, opened 2026-09-02
fix(notifications): validate pagination

### LINKED ISSUE #2091: Notification APIs silently rewrite invalid pagination
Both notification list contracts silently rewrite invalid pagination: the API-key route (`/api/agents/me/notifications`) and the session routes (`/api/notifications`, including their aliases) use Flask typed defaults plus min/max clamps. Malformed page values become 1, non-positive values become 1, and `per_page=51` becomes 50.

Expected: both contracts should reject malformed, fractional, non-positive, and over-50 values with HTTP 400 before notification queries, preserve their defaults, and reject page values whose computed SQLite offset exceeds the signed 64-bit range.

nwinkelman2:
Closes #2091

## What changed

- applies strict bounded pagination to both API-key and session notification lists
- rejects malformed, fractional, non-positive, and over-50 values
- rejects page values whose computed SQLite offset exceeds the signed 64-bit bound
- preserves current defaults and valid boundaries across both authentication modes
- adds an integrated 42-request regression covering invalid and valid contracts

## Verification

- `python -m pytest tests/test_notifications.py -q` (3 passed)
- `python -m py_compile bottube_server.py tests/test_notifications.py`
- `git diff --check` (clean)

RustChain payout address: `RTCe625bc2d5c25d4348236a4fcb74e5d0ef01f6d6d`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -10904,8 +10904,16 @@ def subscription_feed():
 @require_api_key
 def my_notifications():
     """List notifications for the authenticated agent."""
-    page = max(1, request.args.get("page", 1, type=int))
-    per_page = min(50, max(1, request.args.get("per_page", 20, type=int)))
+    page, error = _parse_positive_int_query("page", 1)
+    if error:
+        return error
+    per_page, error = _parse_positive_int_query(
+        "per_page", 20, max_value=50,
+    )
+    if error:
+        return error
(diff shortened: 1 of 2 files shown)


Scottcjn:

nwinkelman2:
Rebased against the current main branch. The requested behavior and regression coverage are already present upstream, leaving this PR with no distinct diff. Closing the redundant branch rather than preserving a stale/conflicting duplicate.

PR #2097, opened 2026-09-02
Validate Sophia chat JSON input

### LINKED ISSUE #2093: Sophia chat crashes on malformed JSON and typed fields
### Problem

`POST /api/sophia` assumes the parsed JSON body, embedded `agent_api_key`, and `message` are string-shaped. A valid JSON array reaches `.get(...)` in caller resolution and raises `AttributeError`. Numeric or structured values for the embedded API key or message likewise reach `.strip()` and raise before authentication, rate limiting, or any model call.

### Expected

- Reject non-object JSON with a deterministic HTTP 400 response.
- Reject non-string `agent_api_key` and `message` fields with field-specific HTTP 400 errors.
- Validate before authentication or inference while preserving CORS preflight, header/session authentication, public-chat behavior, and valid string messages.
- Add focused regressions proving malformed requests never call the Sophia backend.

nwinkelman2:
## Summary
- parse and validate the Sophia request body before caller lookup or model inference
- reject non-object JSON and non-string `agent_api_key` / `message` values with stable 400 responses
- preserve CORS preflight, public chat, authentication, and valid string-message behavior

## Verification
- `python -m pytest tests/test_sophia_input_validation.py -q` (3 passed)
- `python -m py_compile sophia_blueprint.py tests/test_sophia_input_validation.py`
- `git diff --check`

Fixes #2093

### DIFF
--- a/sophia_blueprint.py
+++ b/sophia_blueprint.py
@@ -174,12 +174,11 @@ def _conn():
     return c
 
 
-def _resolve_caller():
+def _resolve_caller(body):
     """Return (agent_id, api_key, name, is_human) for an API agent or a logged-in human,
     or ('__error__', ...) on DB failure, else None for genuine no-auth."""
     api_key = request.headers.get("X-API-Key", "")
     if not api_key:
-        body = request.get_json(silent=True) or {}
         api_key = (body.get("agent_api_key") or "").strip()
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2098, opened 2026-09-02
fix: synchronize Discover tab state and navigation

### LINKED ISSUE #2095: Discover filter tabs expose stale selection and no keyboard navigation
## Bug

The Discover page advertises its five content filters as an ARIA tablist, but the selected state is only updated visually for three of the five controls. `aria-selected` never changes, Categories and Tags do not update the active tab at all, every tab remains in the keyboard tab order, and there is no Arrow/Home/End navigation or tab-to-panel relationship.

## Reproduction

1. Open `/discover` and activate `For You`, `Creators`, `Categories`, or `Tags`.
2. Inspect the tablist or navigate it with a keyboard/screen reader.

`Trending` continues to expose `aria-selected="true"`; the activated control remains `false`. Arrow keys do nothing, and assistive technology cannot identify which panel each tab controls.

## Expected

- Exactly one active tab exposes `aria-selected="true"` and `tabindex="0"`.
- Each tab has a unique ID and `aria-controls` relationship to a named tabpanel.
- Click, Enter/Space, Arrow keys, Home, and End use the same activation path.
- Categories and Tags participate in the same state contract as the three primary content filters.
- Focused executable coverage proves selection and keyboard behavior.

## Scope

Discover-page tab state and keyboard behavior only; no search API or data changes.

nwinkelman2:
## Summary

- give all five Discover filters unique tab-to-panel relationships
- keep active styling, `aria-selected`, and the roving tab stop synchronized
- route Categories and Tags through the same selection state as other filters
- add wrapping Arrow navigation plus Home/End with focused executable coverage

## Verification

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_discover_tabs.py tests/test_accessibility.py` — 27 passed, 10 subtests passed
- `git diff --check` — clean

Fixes #2095

### DIFF
--- a/bottube_templates/discover.html
+++ b/bottube_templates/discover.html
@@ -270,27 +270,27 @@
 
     <!-- Filter Tabs -->
     <div class="filter-tabs" role="tablist" aria-label="Content filters">
-        <button class="filter-tab active" data-filter="all" onclick="loadContent('trending')" role="tab" aria-selected="true">🔥 Trending</button>
-        <button class="filter-tab" onclick="loadContent('foryou')" role="tab" aria-selected="false">✨ For You</button>
-        <button class="filter-tab" onclick="loadContent('agents')" role="tab" aria-selected="false">👤 Creators</button>
-        <button class="filter-tab" onclick="showCategories()" role="tab" aria-selected="false">📁 Categories</button>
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2105, opened 2026-09-02
Fix analytics period request ordering

### LINKED ISSUE #2102: Analytics period changes can render stale out-of-order responses
## Summary

Rapidly changing a Creator Analytics period can render a response for an older selection after the newer response has already rendered. The selected period button and displayed data can therefore disagree.

## Reproduction

1. Open `/analytics` while authenticated.
2. Throttle the views or engagement endpoint so a `7d` request is slow.
3. Select **7 Days**, then immediately select **90 Days**.
4. Let the `90d` response complete first, followed by the older `7d` response.

## Actual behavior

Every request updates the DOM when it resolves. The late `7d` response overwrites the selected `90d` result.

The initial state has the same consistency problem: both selectors mark **7 Days** active while the page loads `30d` data.

## Expected behavior

Only the newest request for each analytics panel may update that panel. Earlier responses should be ignored, including earlier failures, without changing the currently selected period.

## Suggested fix

Track a monotonically increasing request generation separately for the views and engagement panels. Capture it before each fetch and verify it is still current before any DOM mutation. Load the same initial period shown by the active selectors. Add a regression that resolves two requests in reverse order and proves the newest selection remains rendered.

This is an ordinary UI/data-consistency report under the ongoing RustChain bounty pool; no production exploitation was performed.

nwinkelman2:
## Summary

- keep independent request generations for the views and engagement panels
- ignore out-of-order responses before they can mutate either panel
- align initial API requests with the visibly active 7-day selectors
- add a browser-runtime regression that resolves requests in reverse order

## Verification

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_analytics_request_order.py` — 2 passed
- Jinja template parse — passed
- `git diff --check` — passed

Fixes #2102

### DIFF
--- a/bottube_templates/analytics.html
+++ b/bottube_templates/analytics.html
@@ -355,6 +355,9 @@
     integrity="sha384-jb8JQMbMoBUzgWatfe6COACi2ljcDdZQ2OxczGA3bGNeWe+6DChMTBJemed7ZnvJ"
     crossorigin="anonymous"></script>
 <script>
+    let viewsRequestGeneration = 0;
+    let engagementRequestGeneration = 0;
+
     // Fetch summary stats
     async function loadSummary() {
         try {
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2113, opened 2026-09-02
Make playlist append position allocation atomic

### LINKED ISSUE #2111: Concurrent playlist appends can reuse the same position
## Summary

Both playlist append endpoints calculate the next item position with a separate `SELECT MAX(position)` followed by `INSERT`. Concurrent requests can read the same maximum and commit two different videos at the same position, leaving playlist order ambiguous and unstable.

Affected routes:

## Reproduction

1. Create a playlist with at least one item.
2. Send two authenticated add requests for two different visible videos concurrently.
3. Both handlers can read the same current maximum before either insertion commits.
4. Because only `(playlist_id, video_id)` is unique, both inserts may succeed with the same `position`.

The API may then return the same position to both callers, and reads ordered only by `position` have no deterministic order for those items.

## Expected behavior

Appending is atomic per playlist. Each successful request receives a distinct monotonically increasing position, and concurrent duplicate-video requests still return the documented conflict response instead of surfacing an integrity error.

## Suggested fix

Move next-position allocation into the same SQLite `INSERT ... SELECT COALESCE(MAX(position), 0) + 1` statement, shared by both routes. Handle the existing unique-video constraint as a conflict and cover sequential/duplicate/concurrent append behavior with focused regressions.

This is an ordinary data-integrity/concurrency report under the ongoing RustChain bounty pool; no production data was modified.

nwinkelman2:
## Summary

- allocate each playlist position inside the same SQLite write statement as the insert
- share the append operation across API and browser routes
- convert a concurrent same-video uniqueness race into the documented 409 conflict
- add an eight-writer regression proving positions remain unique and contiguous

## Verification

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_playlist_visibility.py tests/test_authenticated_json_object_validation.py` — 8 passed
- `python -m py_compile bottube_server.py tests/test_playlist_visibility.py` — passed
- `git diff --check` — passed

Fixes #2111

### DIFF
--- a/tests/test_playlist_visibility.py
+++ b/tests/test_playlist_visibility.py
@@ -3,7 +3,9 @@ import os
 import sqlite3
 import sys
 import time
+from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
+from threading import Barrier
 
 import pytest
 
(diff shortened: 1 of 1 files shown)


Scottcjn:

Scottcjn (approved):

PR #2114, opened 2026-09-02
Report playlist removal outcomes

### LINKED ISSUE #2110: Playlist removal failures are silent and repeatable
### Problem

The owner-only removal action on a playlist provides no result when the server rejects the request or the network fails. `removeItem()` parses JSON without checking `response.ok`, silently ignores payloads where `ok` is false, and has no rejection handler. The button remains actionable during the request, so repeated clicks can also submit duplicate removals.

A successful removal updates the list, but every failure path is invisible to both visual users and assistive technology.

### Expected

- Gate the removal on both the HTTP response and the JSON success contract.
- Surface server and network failures in an atomic live status without removing the item.
- Disable the selected remove button while its request is in flight and restore it on failure.
- Preserve confirmation, successful removal, and item renumbering.
- Add focused executable regressions for success, HTTP failure, and duplicate-click suppression.

### Duplicate check

Searched all open/closed BoTTube issues and PRs for playlist removal failure, remove response handling, and playlist action feedback; no matching issue or implementation exists. No open PR modifies `bottube_templates/playlist.html`.

nwinkelman2:
## Summary
- require both an HTTP-success and JSON-success result before removing a playlist item
- announce success/server/network failures in an atomic live status
- suppress duplicate removals while the selected button is in flight and restore it after failure
- preserve confirmation, successful removal, and renumbering

## Verification
- `python -m pytest tests/test_playlist_remove_feedback.py -q` (2 passed)
- `python -m py_compile tests/test_playlist_remove_feedback.py`
- `git diff --check`

Fixes #2110

### DIFF
--- a/bottube_templates/playlist.html
+++ b/bottube_templates/playlist.html
@@ -187,6 +187,21 @@
         color: var(--red);
     }
 
+    .pl-remove-btn:disabled {
+        cursor: wait;
+        opacity: 0.55;
+    }
+
+    .playlist-action-status {
+        margin-bottom: 10px;
+        color: var(--text-secondary);
+        font-size: 13px;
+    }
+
+    .playlist-action-status[data-error="true"] {
+        color: var(--red);
+    }
+
     @media (max-width: 768px) {
         .playlist-header {
             flex-direction: column;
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2117, opened 2026-09-02
fix: validate admin media backfill requests

### LINKED ISSUE #2112: Admin media backfills lack a bounded JSON contract
## Summary

Three authenticated media-maintenance routes do not have a reliable request contract:

- `POST /admin/visual/backfill`
- `POST /admin/embeddings/backfill`
- `POST /admin/renditions/backfill`

A JSON array, scalar, or boolean reaches `.get(...)` and raises an exception. Invalid `limit` / `concurrency` values are silently replaced or clipped, and non-string `since_video_id` values can reach `.strip()` and raise. These are operator APIs that launch bounded background work, so silently changing the requested workload is unsafe and a malformed request must not become a server error.

## Expected

- require a JSON object when a body is supplied;
- return deterministic HTTP 400 JSON for malformed integer and string fields;
- require documented positive integer bounds instead of silently rewriting them;
- preserve omitted defaults and valid boundary values;
- validate before any database or media-processing effect.

## Verification scope

Add a shared request parser and executable route regressions across all three endpoints, covering non-object bodies, invalid field types, out-of-range values, defaults, and valid bounds. This is ordinary admin API correctness; it does not weaken authentication or media availability checks.

## Funding

Reported for the active BoTTube bug pool: https://github.com/Scottcjn/rustchain-bounties/issues/71

nwinkelman2:
## Summary

- add one strict JSON-object request contract shared by all three admin media backfill operators
- reject booleans, strings, out-of-range batch limits/concurrency, malformed cursors, and invalid targeted video IDs before schema, database, or ffmpeg effects
- preserve documented defaults and valid boundary values while replacing silent coercion/clamping with actionable 400 responses

## Verification

- `python -m pytest tests/test_admin_media_backfill_contract.py -q --disable-warnings --maxfail=1` (3 passed; 24 contract cases)
- `python -m py_compile bottube_server.py tests/test_admin_media_backfill_contract.py`
- `git diff --check`

Closes #2112

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -20954,21 +20954,102 @@ def _uv_cache_warm():
     return True
 
 
+def _parse_admin_media_batch_request(
+    *,
+    default_limit,
+    max_limit,
+    default_concurrency=None,
+    max_concurrency=None,
+):
+    """Validate the shared request contract for media backfill operators."""
+    data, error = _json_object_body()
+    if error:
+        return None, error
+
+    def positive_int(field, default, maximum):
+        value = data.get(field, default)
+        if isinstance(value, bool) or not isinstance(value, int):
+            return None, (
+                jsonify({"ok": False, "error": f"{field} must be an integer"}),
+                400,
+            )
+        if value < 1 or value > maximum:
+            return None, (
+                jsonify({
+                    "ok": False,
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2119, opened 2026-09-02
Fix concurrent same-voter vote races

### LINKED ISSUE #2116: Concurrent same-voter requests can corrupt vote counters
## Problem

## Reproduction

## Proposed fix

nwinkelman2:
## Summary

- acquire a SQLite write transaction before reading each voter's current state
- apply the serialized transition to both API-key and browser-session video/comment vote routes
- keep duplicate simultaneous requests idempotent instead of raising a unique-constraint error
- add deterministic two-writer regressions for video and comment votes

## Mutation proof

With the transaction-owner calls removed, both new regressions fail with the reproduced unique constraints:

- `UNIQUE constraint failed: votes.agent_id, votes.video_id`
- `UNIQUE constraint failed: comment_votes.agent_id, comment_votes.comment_id`

## Verification

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_vote_concurrency.py tests/test_quests.py tests/test_public_interaction_visibility.py` — 22 passed
- `python -m py_compile bottube_server.py tests/test_vote_concurrency.py` — passed
- `git diff --check` — passed

Fixes #2116

### DIFF
--- a/tests/test_vote_concurrency.py
+++ b/tests/test_vote_concurrency.py
@@ -1,0 +1,144 @@
+# SPDX-License-Identifier: MIT
+import os
+import sqlite3
+import sys
+import threading
+import time
+from pathlib import Path
+
+import pytest
+
+
+ROOT = Path(__file__).resolve().parents[1]
+if str(ROOT) not in sys.path:
+    sys.path.insert(0, str(ROOT))
+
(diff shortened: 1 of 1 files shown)


Scottcjn:

Scottcjn (approved):

PR #2127, opened 2026-09-02
Bound Studio job polling failures

### LINKED ISSUE #2125: Studio polling hides repeated job endpoint failures
## Problem

After Studio charges RTC and starts an asynchronous video/model job, `poll()` parses every response as JSON without checking its HTTP status. Non-JSON responses and HTTP errors are silently swallowed, and polling continues for up to six minutes while the page still implies generation is progressing. Repeated failures are indistinguishable from a healthy queued job.

This is particularly damaging after a successful charge: the user cannot tell whether the job endpoint is unavailable, the response is malformed, or generation is genuinely still running.

## Reproduction

1. Start a video or 3D Studio generation so `_doGenerate()` calls `poll(status_url)`.
2. Make the status endpoint return a non-JSON 404/5xx response or repeated network failures.
3. Observe the UI continue to report generation/progress until the generic timeout, with no truthful polling failure state.

## Expected

- Check HTTP status before accepting a job payload.
- Handle non-JSON responses without throwing away the HTTP error.
- Keep transient failures bounded and visible while retrying.
- Stop after a small consecutive-failure budget, restore controls, and tell the user the job may still be running and where to look.
- Reset the failure counter after a valid response.
- Cover successful, terminal-job, transient-recovery, and exhausted-error paths with executable regressions.

nwinkelman2:
## Summary
- reject failed HTTP and malformed/non-object Studio job responses instead of treating them as healthy progress
- expose bounded transient retry state and stop after three consecutive failures
- reset the failure budget after a valid job response and suppress overlapping poll requests
- restore generation controls on terminal, timeout, and retry-exhaustion outcomes
- announce Studio job status through an atomic live region

## Tests
- `python -m pytest tests/test_studio_poll_contract.py -q` (3 passed)
- `python -m py_compile tests/test_studio_poll_contract.py`
- `git diff --check`

Fixes #2125

### DIFF
--- a/bottube_templates/studio.html
+++ b/bottube_templates/studio.html
@@ -124,7 +124,7 @@
     <button class="gen" id="i2v-btn" onclick="stGenerate('i2v')">Animate Image · <span class="btnc-i2v">5</span> RTC</button>
   </div>
 
-  <div class="st-status" id="st-status"></div>
+  <div class="st-status" id="st-status" role="status" aria-live="polite" aria-atomic="true"></div>
   <div class="st-result" id="st-result"></div>
 
   <div class="st-foot">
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2130, opened 2026-09-02
fix(comments): reject hidden video targets

### LINKED ISSUE #2129: Comment write routes accept removed and banned-owner videos
## Problem

Both comment-write routes validate a target with `SELECT id FROM videos WHERE video_id = ?`. That accepts videos which are no longer public:

- videos with `is_removed = 1`;
- videos owned by a banned agent.

The public metadata, stream, watch, description, and comment-read routes correctly return 404 for those targets, so the write edge can create comments, notifications, quest progress, and reward activity against content that no public reader can access.

Affected routes:

- `POST /api/videos/<video_id>/comment`
- `POST /api/videos/<video_id>/web-comment`

## Expected behavior

Both comment-write routes should use the canonical public-video predicate before any content validation, insert, reward, notification, or quest side effect. Removed or banned-owner targets should return the same 404 as the public read routes and leave the comments table unchanged.

## Proposed fix

Join the target video's owner and reuse `_public_video_filter_sql()` in both target lookups. Add an executable regression covering both authentication paths and both hidden-video states, including proof that no comment row is inserted.

nwinkelman2:
Closes #2129.

## What changed

- API-key comment creation now resolves its target through the canonical public-video predicate.
- Web-session comment creation applies the identical target contract.
- Removed videos and banned-owner videos return `404 Video not found` before validation, insertion, rewards, quest progress, or notifications.
- Added a regression spanning both authentication paths and both hidden-video states, including zero-row side-effect proof.

## Proof

Mutation baseline (old target lookup): the regression failed with `201 CREATED` for a removed video.

Fixed verification:

- `python -m pytest tests/test_quests.py -q` — 18 passed
- `python -m py_compile bottube_server.py tests/test_quests.py`
- `git diff --check`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -7761,7 +7761,12 @@ def add_comment(video_id):
         return jsonify({"error": "Comment rate limit exceeded. Try again later."}), 429
 
     db = get_db()
-    video = db.execute("SELECT id FROM videos WHERE video_id = ?", (video_id,)).fetchone()
+    video = db.execute(
+        f"""SELECT v.id
+            FROM videos v JOIN agents a ON v.agent_id = a.id
+            WHERE v.video_id = ? AND {_public_video_filter_sql()}""",
+        (video_id,),
+    ).fetchone()
     if not video:
         return jsonify({"error": "Video not found"}), 404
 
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2133, opened 2026-09-02
Make webhook delivery-window accounting atomic

### LINKED ISSUE #2132: Concurrent webhook deliveries lose rate-window accounting
## Summary

`fire_webhooks()` starts one delivery thread per event. Every thread reads `event_window_start` and `event_count` before sending, then writes an absolute `event_count + 1` only after success. Concurrent successful deliveries for the same hook can therefore reserve the same count and overwrite each other, so persisted delivery accounting falls behind the actual number sent and the documented 100-events/hour cap is not deterministic.

## Reproduction

1. Register one active webhook with an event count below 100.
2. Trigger multiple matching events concurrently.
3. Let two delivery workers read the same count before either updates it.
4. Both send successfully and both write the same next count.

The receiver gets two events while the database records one. Near the boundary, multiple workers can all pass the `event_count >= 100` check and deliver beyond the configured window cap.

## Expected behavior

Each eligible delivery atomically reserves one distinct slot for its webhook and window before performing network I/O. A full window rejects further delivery workers, and an expired window rolls over exactly once without losing concurrent increments.

## Proposed fix

Move window rollover, capacity checking, and slot increment into one conditional SQLite update, then deliver only when that reservation succeeds. Add focused multi-connection regressions for unique slot allocation, the 100-slot boundary, and expired-window rollover.

This is an ordinary webhook delivery-accounting and concurrency defect under the ongoing RustChain bounty pool; no production data was modified.

nwinkelman2:
## Summary
- atomically reserve each webhook delivery slot before network I/O
- serialize window rollover, capacity checks, and increments across delivery workers
- stop full windows exactly at 100 attempts and preserve truthful accounting during receiver failures
- add focused multi-connection concurrency regressions

## Verification
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_webhook_delivery_slots.py tests/test_authenticated_json_object_validation.py` — 5 passed
- `python -m py_compile bottube_server.py tests/test_webhook_delivery_slots.py` — passed
- `git diff --check` — clean

Fixes #2132

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -4120,6 +4120,55 @@ def _canonical_webhook_event(event: str) -> str:
     return mapping.get(event, event)
 
 
+def _reserve_webhook_delivery_slot(
+    conn: sqlite3.Connection,
+    webhook_id: int,
+    now: float,
+    limit: int = 100,
+    window_seconds: int = 3600,
+) -> bool:
+    """Atomically reserve one delivery slot in a webhook's rate window.
+
+    Webhook events are dispatched by independent worker threads.  Serializing
+    the read/rollover/increment operation with ``BEGIN IMMEDIATE`` keeps two
+    workers from observing and then writing the same counter value.  A slot is
+    consumed before network I/O so an unavailable receiver cannot bypass the
+    outbound delivery cap by repeatedly failing requests.
+    """
+    try:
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2137, opened 2026-09-02
Make duplicate subscribe requests atomically idempotent

### LINKED ISSUE #2136: Concurrent duplicate subscribes can surface an integrity failure
## Summary

`POST /api/agents/<agent_name>/subscribe` checks for an existing subscription with a separate `SELECT` and then performs a plain `INSERT` into a table whose primary key is `(follower_id, following_id)`. Two concurrent subscribe requests can both observe no row; one insert wins and the other raises `sqlite3.IntegrityError`, surfacing a 500 even though the requested end state was reached.

## Reproduction

1. Create two agents and authenticate as the prospective follower.
2. Send two subscribe requests for the same target concurrently.
3. Let both requests run the existence check before either commits.
4. Both attempt the same primary-key insert; one request fails instead of returning the endpoint's documented idempotent `Already following` response.

Depending on timing, notification and first-follow quest effects can also diverge from the actual row creator.

## Expected behavior

Subscription creation is atomic and idempotent. Exactly one request creates the row and produces new-subscription side effects; every concurrent loser returns a successful `Already following` result. No request exposes a database integrity error.

## Proposed fix

Use one conflict-tolerant insert and inspect whether that statement created the row. Run notification/quest effects only for the winner, and cover concurrent requests plus the existing sequential idempotency contract with executable regressions.

This is an ordinary data-integrity/concurrency defect under the ongoing RustChain bounty pool; no production data was modified.

nwinkelman2:
## Summary
- replace the subscribe route's check-then-insert race with one conflict-tolerant insert
- detect the single request that created the relationship from SQLite's statement result
- emit notification and first-follow quest effects only for that creator
- return the existing successful `Already following` contract to every concurrent loser
- add focused multi-connection concurrency and idempotency regressions

## Verification
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_subscription_concurrency.py tests/test_subscription_visibility.py` — 7 passed
- `python -m py_compile bottube_server.py tests/test_subscription_concurrency.py` — passed
- `git diff --check` — clean

Fixes #2136

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -10740,6 +10740,22 @@ def update_profile():
 # Subscriptions / Follow
 # ---------------------------------------------------------------------------
 
+def _insert_subscription_once(
+    db: sqlite3.Connection,
+    follower_id: int,
+    following_id: int,
+    created_at: float,
+) -> bool:
+    """Create a subscription atomically, returning whether this call won."""
+    cursor = db.execute(
+        """INSERT OR IGNORE INTO subscriptions
+           (follower_id, following_id, created_at)
+           VALUES (?, ?, ?)""",
+        (follower_id, following_id, created_at),
+    )
+    return cursor.rowcount == 1
+
+
 @app.route("/api/agents/<agent_name>/subscribe", methods=["POST"])
 @require_api_key
 def subscribe_agent(agent_name):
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2144, opened 2026-09-02
Make duplicate comment creation atomic

### LINKED ISSUE #2142: Concurrent identical comments bypass duplicate prevention
## Summary

Both comment creation paths reject a repeated identical comment with a separate `SELECT` before a plain `INSERT`:

- `POST /api/videos/<video_id>/comments`
- `POST /api/videos/<video_id>/web-comment`

Concurrent requests can both observe no matching row and then both create the same comment. Each winner path can also independently run comment rewards and owner notifications, so the duplicate-prevention contract does not protect downstream effects.

## Reproduction

1. Authenticate one agent/user and choose one public video.
2. Send two identical comment bodies concurrently.
3. Let both handlers complete the duplicate query before either insert commits.
4. Both requests insert the same `(video, author, content)` comment and execute new-comment side effects.

## Expected behavior

Duplicate detection and insertion are atomic. Exactly one concurrent request creates the comment and runs rewards/notification work; every other request receives the documented 409 duplicate response and the existing comment ID when available.

## Proposed fix

Share a short SQLite write transaction that checks and inserts the comment as one serialized operation. Leave the transaction open for the winning caller's reward/notification effects, then commit once. Add a multi-connection regression proving one row and one creator under concurrent load.

This is an ordinary content/reward data-integrity defect under the ongoing RustChain bounty pool; no production data was modified.

nwinkelman2:
## Summary
- serialize duplicate lookup and comment insertion with a short SQLite write transaction
- share the atomic creator across API and web comment paths
- preserve reward, quest, and notification work for the single winning request
- return the existing comment ID to every concurrent duplicate
- add focused multi-connection concurrency and author-scope regressions

## Verification
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_comment_creation_concurrency.py tests/test_quests.py -k "comment or concurrency"` — 8 passed
- `python -m py_compile bottube_server.py tests/test_comment_creation_concurrency.py` — passed
- `git diff --check` — clean

Fixes #2142

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -7752,6 +7752,39 @@ def describe_video(video_id):
 # Comments
 # ---------------------------------------------------------------------------
 
+def _insert_comment_once(
+    db: sqlite3.Connection,
+    video_id: str,
+    agent_id: int,
+    parent_id: Optional[int],
+    content: str,
+    comment_type: str,
+    created_at: float,
+) -> Tuple[bool, int]:
+    """Atomically create a comment unless identical content already exists.
+
+    The caller owns the write transaction after this function returns so the
+    winning request can commit rewards and notifications with the new row.
+    """
+    db.execute("BEGIN IMMEDIATE")
+    existing = db.execute(
+        """SELECT id FROM comments
+           WHERE video_id = ? AND agent_id = ? AND content = ?
+           LIMIT 1""",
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2148, opened 2026-09-02
Serialize one-time quest reward refresh

### LINKED ISSUE #2146: Concurrent quest refreshes can race one-time RTC awards
## Summary

`_refresh_agent_quests()` reads quest progress, completion state, `rewarded_at`, and the earnings ledger before it writes the refreshed row and one-time RTC award. Some callers already hold a write transaction, but read-only entry points such as `GET /api/agents/me/quests` and the profile page call it without one.

Concurrent refreshes for the same newly completed quest can both observe a missing/unrewarded row. They may collide on the `(agent_id, quest_key)` primary key or both enter the one-time reward decision before either commit, exposing an integrity failure or duplicate award work.

## Reproduction

1. Create an agent whose profile-completion metric meets a rewarded quest for the first time.
2. Invoke two quest refreshes concurrently on separate SQLite connections.
3. Let both read the quest and agent state before either writes `agent_quests`.
4. Both attempt the completion/upsert/reward path from the same stale snapshot.

## Expected behavior

Quest progress refresh and one-time reward decisions serialize per database. Exactly one canonical progress row exists, exactly one earning is written, and the RTC balance increases once. Callers that already own a write transaction keep using it without a nested-transaction error.

## Proposed fix

Have the canonical refresh owner acquire a short SQLite write transaction before its first snapshot only when the caller does not already have a transaction. Keep the caller-owned commit contract and add a multi-connection regression for concurrent first completion.

This is an ordinary quest/reward data-integrity defect under the ongoing RustChain bounty pool; no production data was modified.

nwinkelman2:
## Summary
- acquire SQLite's write transaction before the first quest snapshot when the caller does not already own one
- serialize first-completion, progress, reward, and balance decisions across concurrent refreshes
- preserve the existing caller-owned transaction contract
- prove ten concurrent connections produce one progress row, one earning row, and one balance increase

## Verification
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_quest_refresh_concurrency.py` (2 passed)
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_quests.py::test_quests_endpoint_unlocks_onboarding_flow tests/test_quests.py::test_quest_rewards_are_idempotent_and_leaderboard_updates` (2 passed)
- `python -m py_compile bottube_server.py tests/test_quest_refresh_concurrency.py`
- `git diff --check`

Fixes #2146

RTC wallet: `RTCe625bc2d5c25d4348236a4fcb74e5d0ef01f6d6d`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -3023,6 +3023,13 @@ def _refresh_agent_quests(
 ) -> List[Dict]:
     """Refresh quest progress for an agent. Checks all active quests and updates completion state."""
     """Refresh quest progress, award one-time RTC, and return quest snapshots."""
+    # Read-only routes also call this mutating owner. Acquire the write lock
+    # before its first snapshot so concurrent first-completion requests cannot
+    # both observe a missing/unrewarded row. Callers that already performed a
+    # write keep their existing transaction and commit all effects together.
+    if not db.in_transaction:
+        db.execute("BEGIN IMMEDIATE")
+
     params: list = []
     where = "WHERE is_active = 1"
     if quest_keys:
--- a/tests/test_quest_refresh_concurrency.py
+++ b/tests/test_quest_refresh_concurrency.py
@@ -1,0 +1,125 @@
+"""Concurrency regressions for one-time quest rewards."""
+
+import sqlite3
+import threading
+
+import bottube_server
+
+
+def _create_quest_db(path):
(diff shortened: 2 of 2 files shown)


Scottcjn (approved):

PR #2151, opened 2026-09-02
Enforce webhook registration quota atomically

### LINKED ISSUE #2150: Concurrent webhook registration can exceed the per-agent limit
## Problem

`POST /api/webhooks` enforces the five-webhook quota with a `SELECT COUNT(*)` followed later by `INSERT`, without a write transaction spanning the decision. Concurrent requests can all observe the same count below five and then commit more than five rows for one agent.

This is a quota/data-integrity defect: the response contract says “Maximum 5 webhooks per agent,” but the database can exceed it under ordinary concurrent registration.

## Reproduction

1. Seed one agent with four webhook rows.
2. Open multiple SQLite connections and synchronize concurrent registrations.
3. Each connection reads count `4` before any insert commits.
4. More than one insert succeeds, leaving over five rows.

## Proposed fix

Make quota admission and insertion one atomic write-serialized operation, preserve validation before lock acquisition, and add a multi-connection regression proving only one of many concurrent fifth-slot attempts succeeds.

Funded under Scottcjn/rustchain-bounties#71.

nwinkelman2:
## Summary
- move the per-agent webhook quota check into one write-serialized admission owner
- keep request/event validation outside the database write-lock window
- preserve caller-owned transaction semantics and roll back helper-owned rejected/error paths
- prove ten concurrent registrations for one remaining slot admit exactly one row

## Verification
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_webhook_quota_concurrency.py` (2 passed)
- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 python -m pytest -q tests/test_authenticated_json_object_validation.py` (2 passed)
- `python -m py_compile bottube_server.py tests/test_webhook_quota_concurrency.py`
- `git diff --check`

Fixes #2150

RTC wallet: `RTCe625bc2d5c25d4348236a4fcb74e5d0ef01f6d6d`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -11449,6 +11449,40 @@ def web_remove_from_playlist(playlist_id):
 WEBHOOK_EVENTS = ["video.uploaded", "video.voted", "comment.created", "agent.created", "comment", "like", "subscribe", "new_video", "mention", "*"]
 
 
+def _insert_webhook_with_quota(
+    db: sqlite3.Connection,
+    agent_id: int,
+    url: str,
+    secret: str,
+    events: str,
+    created_at: float,
+    *,
+    limit: int = 5,
+) -> bool:
+    """Atomically admit one webhook without exceeding an agent's quota."""
+    started_transaction = not db.in_transaction
+    if started_transaction:
+        db.execute("BEGIN IMMEDIATE")
+    try:
+        count = db.execute(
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2152, opened 2026-09-02
Insert community reports atomically

### LINKED ISSUE #2149: Concurrent duplicate reports can inflate moderation thresholds
## Summary

Video and comment reporting use a duplicate `SELECT` followed by a separate `INSERT`, with no uniqueness constraint. Concurrent retries from one reporter can both pass the check and create duplicate pending reports. For videos, duplicate rows also inflate the 3-report auto-flag threshold, allowing one reporter to trigger moderation review as if multiple agents reported the content.

## Affected routes

- `POST /api/videos/<video_id>/report`
- `POST /api/comments/<comment_id>/report`

## Expected

Duplicate detection and insertion must be one atomic SQLite write. The first report should insert; any same-reporter retry should preserve the existing 409 contract without adding a row. Video threshold counts must reflect unique reporters.

## Regression

Exercise the shared insert operation twice for one video/reporter and twice for one comment/reporter. Assert exactly one row per target and inserted/duplicate outcomes. Keep route response and moderation-threshold behavior unchanged for legitimate unique reporters.

nwinkelman2:
Closes #2149

## What changed

- replace video/comment report duplicate-check + insert sequences with one shared atomic SQLite write
- preserve the existing 409 duplicate contracts and release no-op write transactions promptly
- ensure a reporter contributes at most one row to a video's moderation threshold
- keep legitimate unique-reporter threshold behavior unchanged

## Proof

- mutation baseline: focused regression failed because no atomic insert operation existed and both routes used check-then-act writes
- focused report and threshold tests: `10 passed`
- full report + quests suites: `25 passed`
- route/atomic report suite after final integration coverage: `9 passed`
- `python -m py_compile bottube_server.py tests/test_report_input_validation.py`
- `git diff --check`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -17483,6 +17483,43 @@ def _get_reporter_id():
         return g.user['id']
     return None
 
+
+def _insert_report_once(
+    db,
+    *,
+    reporter_id: int,
+    reason: str,
+    details: str,
+    video_id=None,
+    comment_id=None,
+    created_at=None,
+) -> bool:
+    """Atomically insert one report per reporter and target."""
+    if (video_id is None) == (comment_id is None):
+        raise ValueError("exactly one report target is required")
+    target_column = "video_id" if video_id is not None else "comment_id"
+    target_value = video_id if video_id is not None else int(comment_id)
+    timestamp = time.time() if created_at is None else float(created_at)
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2162, opened 2026-09-02
fix(syndication): compare-and-set state transitions

### LINKED ISSUE #2161: Concurrent syndication transitions can overwrite terminal state
## Bug

`SyndicationQueue.update_state()` reads and validates the current state, then updates by `id` alone. Two workers can both validate the same `processing` snapshot and successfully write conflicting terminal outcomes (`completed` and `failed`). Both callers return `True`, while the last writer silently wins, violating the documented queue state machine and making delivery state nondeterministic.

`mark_failed()` has the same stale-state problem across its `processing -> failed -> pending` retry sequence.

## Reproduction

1. Enqueue and dequeue an item so it is `processing`.
2. Synchronize two workers after they both read `processing`.
3. Have one call `mark_completed()` and the other call `mark_failed(..., auto_retry=False)`.
4. Observe both return success even though only one transition can legitimately consume the `processing` state.

## Expected

The validated source state must be part of the update predicate. Exactly one competing transition should affect a row and report success; the loser should return `False` without overwriting the winner.

## Proposed fix

Use compare-and-set updates (`WHERE id = ? AND state = ?`), gate success on affected-row count, and add a deterministic concurrent terminal-transition regression.

nwinkelman2:
## Summary

- make each syndication transition a compare-and-set against its validated source state
- report failure when another worker consumed the transition first
- guard failed-to-pending retries against stale state
- prove competing terminal transitions have exactly one winner

Closes #2161.

## Proof

- Mutation baseline: bypassing affected-row gating makes the contention regression fail because both workers report success.
- `python -m pytest tests/test_syndication_queue.py -q` — 28 passed
- `python -m pytest tests/test_syndication_queue.py tests/test_syndication.py -q` — 50 passed
- `python -m compileall -q syndication_queue.py tests/test_syndication_queue.py`
- `git diff --check`

### DIFF
--- a/tests/test_syndication_queue.py
+++ b/tests/test_syndication_queue.py
@@ -7,6 +7,7 @@ Run with: pytest tests/test_syndication_queue.py -v
 
 import os
 import sys
+import threading
 import time
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


Scottcjn:

Scottcjn (approved):

PR #2164, opened 2026-09-02
fix(gpu): serialize marketplace job claims

### LINKED ISSUE #2163: Concurrent GPU job claims can assign one job twice
## Bug

`POST /api/gpu/jobs/claim` checks a job is `pending`, then updates it by `id` alone. Two GPU providers can both read the same pending job and both receive a successful claim response. The last write silently becomes `provider_id`, while both providers are marked `busy`, so one provider begins work that can never be completed or paid under its identity.

The same route can also let one provider concurrently claim two different jobs because the provider's non-busy check is another read-then-write edge.

## Expected

Job and provider admission must be one transaction:

- exactly one `pending -> claimed` transition wins for a job
- a provider can reserve at most one job at a time
- if provider reservation fails, the job transition rolls back
- losing requests receive a conflict response rather than false success

## Proposed fix

Use conditional affected-row updates for both the job and provider, commit them together, roll back either failure, and add deterministic contention regressions for same-job and same-provider races.

nwinkelman2:
## Summary

- reserve a pending GPU job and non-busy provider in one transaction
- compare-and-set both admission edges and roll back the job when provider reservation loses
- return truthful conflict responses instead of letting two workers start the same job
- cover same-job and same-provider contention with real SQLite connections

Closes #2163.

## Proof

- Mutation baseline: removing the pending-job predicate makes the same-job regression fail because both providers report `claimed`.
- `python -m pytest tests/test_gpu_marketplace_claim_concurrency.py tests/test_gpu_marketplace_validation.py tests/test_gpu_marketplace_auth.py -q` — 29 passed
- `python -m compileall -q gpu_marketplace.py tests/test_gpu_marketplace_claim_concurrency.py`
- `git diff --check`

### DIFF
--- a/gpu_marketplace.py
+++ b/gpu_marketplace.py
@@ -145,6 +145,36 @@ def init_gpu_db(db_path: str = None):
 # HELPERS
 # ---------------------------------------------------------------------------
 
+def _claim_gpu_job_transaction(db, provider_id: str, agent_id: int, job_id: str, now: int) -> str:
+    """Atomically reserve one pending job for one non-busy provider."""
+    claimed = db.execute(
+        """
+        UPDATE gpu_jobs
+        SET status = 'claimed', provider_id = ?, claimed_at = ?
+        WHERE id = ? AND status = 'pending'
+        """,
+        (provider_id, now, job_id),
+    )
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):

PR #2172, opened 2026-09-02
fix(gpu_marketplace): prevent concurrent duplicate settlement on job completion (#2165)

katakata0522:
## Summary
Fixes #2165.

### Problem
`POST /api/gpu/jobs/complete` validated that a job was `running` in a preliminary SELECT, but executed the completion UPDATE with `WHERE id = ?` only, before immediately incrementing provider totals and inserting history.
Two concurrent completion requests could both pass validation, execute the UPDATE, increment `total_jobs` twice, add the payment twice to `total_rtc_earned`, and write duplicate completion history for one job.

### Fix
1. **Atomic Compare-and-Set**: Added `WHERE id = ? AND provider_id = ? AND status = 'running'` to the UPDATE statement, turning the state transition into an atomic admission gate.
2. **Gated Settlement Effects**: Checks `cursor.rowcount`. If 0 (indicating a lost race or invalid state), rolls back the transaction and returns `HTTP 409 Conflict`.
3. **Regression Test**: Added `test_concurrent_complete_job_credits_only_once` in `tests/test_gpu_marketplace_validation.py` using concurrent threads, asserting exactly one 200 OK, `total_jobs == 1`, and exactly 1 history record.

### Verification
Local test suite passing:
```
tests/test_gpu_marketplace_validation.py .......................... [100%]
26 passed in 0.58s
```

PR #2171, opened 2026-09-02
fix(gpu_marketplace): prevent reopening completed jobs via fail endpoint (#2167)

katakata0522:
## Summary
Fixes #2167.

### Problem
`POST /api/gpu/jobs/fail` previously verified only provider ownership and unconditionally executed:
```sql
UPDATE gpu_jobs SET status = 'pending', provider_id = NULL, claimed_at = NULL, error_message = ? WHERE id = ?
```
This allowed settled / completed jobs to be reopened while leaving `rtc_paid`, completion metadata, and provider earnings intact, permitting repeat claims and settlements from a single escrow record.

### Fix
1. **State Validation Guard**: Explicitly rejects `/jobs/fail` requests if the job is not in an active state (`claimed` or `running`), returning HTTP 400.
2. **Atomic Compare-and-Set**: Added `WHERE id = ? AND provider_id = ? AND status IN ('claimed', 'running')` directly into the UPDATE query.
3. **Concurrency Protection**: Evaluates `cursor.rowcount`. If 0, rolls back and returns HTTP 409 Conflict, ensuring only the winning lifecycle transition can release the provider back to `online`.
4. **Timestamp Reset**: Clears `started_at` in addition to `claimed_at` when returning the job to `pending`.
5. **Regression Tests**: Added tests in `tests/test_gpu_marketplace_validation.py` verifying:
   - Completed jobs cannot be failed or reopened (HTTP 400/409, terminal status preserved).
   - Running jobs can be failed and released back to pending.
   - Claimed jobs can be failed and released back to pending.

### Verification
All tests passing locally:
```
tests/test_gpu_marketplace_validation.py ............................ [100%]
28 passed in 0.58s
```

PR #1896, opened 2026-09-01
[TRAVIS] Align subscriber totals across public surfaces

nwinkelman2:
## Summary
- centralize public subscriber totals behind the same banned-follower predicate as subscriber lists
- align channel, watch, analytics, dashboard, and subscribe-response counts
- exclude banned followers from daily subscriber-growth analytics
- add cross-surface contract tests

## Reproduction / mutation proof
With one visible follower and one banned follower, `/api/agents/<name>/subscribers` correctly returned `count: 1`, while the channel, watch page, and analytics reported 2. Both new regressions failed against `main` before the repair.

## Validation
- `C:\Python314\python.exe -m pytest -q tests/test_subscriber_count_consistency.py` -> 2 passed
- `C:\Python314\python.exe -m pytest -q tests/test_subscription_visibility.py` -> 5 passed
- Python compilation clean
- `git diff --check`

Closes #1895

PR #2174, opened 2026-09-03
fix(gpu): prevent concurrent double-settlement via CAS on complete_job (#2165)

### LINKED ISSUE #2165: Concurrent GPU completions can credit one job twice
## Bug

`POST /api/gpu/jobs/complete` validates a `running` snapshot, then updates the job by `id` alone before incrementing provider totals and inserting history. Two concurrent completion requests can both pass validation, both return success, increment `total_jobs` twice, add the RTC amount twice to `total_rtc_earned`, and write duplicate completion history for one job.

## Expected

The `running -> completed` transition must be the atomic settlement admission edge. Exactly one request may persist the job result, provider earnings/statistics, and history row. A losing request must roll back and report a conflict.

## Proposed fix

Compare-and-set the job on both `status = 'running'` and `provider_id`, gate all settlement side effects on affected-row count in the same transaction, and add a real-SQLite concurrent completion regression proving one history row and one provider credit.

rafaio1:
## Summary

Fixes #2165 — Concurrent GPU job completion could result in double-settlement because `complete_job()` did not atomically gate all side effects.

## Changes

- **CAS guard on `complete_job()`**: All side effects (job status update, provider stats increment, history insert) are now gated on `UPDATE gpu_jobs ... WHERE status = 'running' AND provider_id = ?` returning `rowcount > 0`. If the CAS fails (another thread already completed the job), the endpoint returns 409 with no side effects.
- **Test harness fix**: Monkeypatched `bottube_server.DB_PATH` so `init_db()` uses the same temp DB as `get_db()` (which reads `BOTTUBE_DB_PATH` env var). Previously these pointed to different files, causing 404s.
- **Test seed fix**: Added required `created_at` column to `gpu_providers` INSERT (NOT NULL constraint).

## Tests

```
tests/test_gpu_concurrent_complete.py::test_concurrent_complete_only_one_wins PASSED
tests/test_gpu_concurrent_complete.py::test_complete_after_completed_rejected PASSED
```

Both tests verify that exactly one concurrent completion succeeds (200) and the other is rejected (409), and that sequential re-completion is also rejected.

### DIFF
--- a/gpu_marketplace.py
+++ b/gpu_marketplace.py
@@ -637,8 +637,8 @@ def complete_job():
         return jsonify({"error": "Job not found"}), 404
     if row[5] != agent['id']:
         return jsonify({"error": "Not your job"}), 403
-    if row[0] != "running":
-        return jsonify({"error": f"Job not running (status: {row[0]})"}), 400
+    if row[0] != "running" or row[1] != provider_id:
+        return jsonify({"error": "Job not running or ownership mismatch"}), 409
 
     # Calculate payment
     now = int(time.time())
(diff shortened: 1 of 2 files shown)


Scottcjn:

PR #2175, opened 2026-09-03
fix(gpu): prevent stale start from creating ownerless running jobs (#2169)

### LINKED ISSUE #2169: Stale GPU start can create ownerless running jobs
## Bug

`POST /api/gpu/jobs/start` validates a `claimed` snapshot, then updates the job by `id` alone. If the provider failure/release path consumes the claim concurrently, the stale start request can still write `status = 'running'` after the job was returned to `pending` and its `provider_id` cleared. The result is an orphaned running job with no provider owner, while the provider is marked available.

Repeated concurrent starts can also both report success instead of treating the state transition as single-use.

## Expected

Starting must compare-and-set the exact provider-owned `claimed` state. Exactly one request may transition to `running`; a stale or repeated request must return a conflict without changing the job.

## Proposed fix

Guard the update by job id, provider id, and claimed state, gate success on affected-row count, and add regressions proving stale/repeated starts cannot produce ownerless running work.

rafaio1:
## Summary

Fixes #2169 — A stale or duplicate `POST /api/gpu/jobs/start` could transition a job to `running` even after it had been released back to `pending` (e.g. by a concurrent failure path), creating an ownerless running job with no provider assigned.

## Root Cause

The original `start_job()` used a non-atomic read-then-write pattern: it checked that the job was in `claimed` state, then issued an unconditional `UPDATE gpu_jobs SET status = 'running' WHERE id = ?`. Between the check and the update, another request could release the job, and the stale start would still succeed because the UPDATE only filtered on `id`.

## Fix

Replace the read-then-write with a single atomic CAS:

```sql
UPDATE gpu_jobs
SET status = 'running', started_at = ?
WHERE id = ? AND status = 'claimed' AND provider_id = ?
```

Success is gated on `rowcount > 0`. If the CAS fails (job no longer claimed by this provider), return 409 with a descriptive error instead of silently succeeding.

## Tests

Three regression tests added in `tests/test_gpu_stale_start.py`:

- **test_start_claimed_job_succeeds**: Normal claimed→running transition works (200)
- **test_stale_start_after_release_rejected**: Job released to pending between claim and start → 409, no ownerless running job created
- **test_duplicate_start_second_rejected**: Two concurrent starts for same claimed job → exactly one 200, one 409

All tests pass.

### DIFF
--- a/gpu_marketplace.py
+++ b/gpu_marketplace.py
@@ -580,26 +580,32 @@ def start_job():
 
     db = get_db()
 
-    # Verify ownership and job state
-    row = db.execute("""
-        SELECT j.status, j.provider_id, p.agent_id
-        FROM gpu_jobs j
-        JOIN gpu_providers p ON j.provider_id = p.id
-        WHERE j.id = ?
-    """, (job_id,)).fetchone()
+    # Verify ownership via provider -> agent link
+    prov = db.execute(
+        "SELECT agent_id FROM gpu_providers WHERE id = ?", (provider_id,)
+    ).fetchone()
+    if not prov or prov[0] != agent['id']:
+        return jsonify({"error": "Not your provider"}), 403
 
-    if not row:
-        return jsonify({"error": "Job not found"}), 404
-    if row[2] != agent['id']:
-        return jsonify({"error": "Not your job"}), 403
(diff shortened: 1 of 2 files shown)


Scottcjn:

PR #2176, opened 2026-09-03
fix(gpu): atomic CAS for concurrent job claims (#2163)

### LINKED ISSUE #2163: Concurrent GPU job claims can assign one job twice

rafaio1:
## Summary

Fixes #2163 — concurrent GPU job claims could both succeed, allowing double-booking of the same job.

### Root Cause
The original `claim_job()` used a non-atomic read-check-write pattern: it first checked if the job was pending and the provider was idle, then issued an UPDATE in a separate statement. Under concurrency, two providers could both pass the check before either wrote, resulting in both claims succeeding.

### Fix
Replaced with a single atomic CAS (Compare-And-Swap) UPDATE:
```sql
UPDATE gpu_jobs SET status='claimed', provider_id=?, claimed_at=?
WHERE id=? AND status='pending'
  AND NOT EXISTS (SELECT 1 FROM gpu_providers WHERE id=? AND status='busy')
```
Returns 409 on contention (job already claimed) or if the provider became busy between check and write.

### Tests
- `test_same_job_concurrent_claim_exactly_one_wins`: two threads race to claim the same job; exactly one gets 200, the other 409.
- `test_busy_provider_cannot_claim_second_job`: a provider that already holds a claimed job is rejected with 400/409 when attempting a second claim.

Both tests pass against the actual DDL schema.

### DIFF
--- a/gpu_marketplace.py
+++ b/gpu_marketplace.py
@@ -530,26 +530,34 @@ def claim_job():
     if prow[1] == "busy":
         return jsonify({"error": "Provider already busy with a job"}), 400
 
-    # Check job is pending
+    # Atomic CAS: claim job only if still pending AND provider not busy
+    now = int(time.time())
+    cur = db.execute("""
+        UPDATE gpu_jobs
+        SET status = 'claimed', provider_id = ?, claimed_at = ?
+        WHERE id = ? AND status = 'pending'
+          AND NOT EXISTS (
+            SELECT 1 FROM gpu_providers WHERE id = ? AND status = 'busy'
+          )
+    """, (provider_id, now, job_id, provider_id))
+
+    if cur.rowcount == 0:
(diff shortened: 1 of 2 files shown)


Scottcjn:

PR #2177, opened 2026-09-03
fix(playlist): atomic next-position allocation closes #2111 race

### LINKED ISSUE #2111: Concurrent playlist appends can reuse the same position

Vyacheslav-Tomashevskiy:
## fix(playlist): atomic next-position allocation closes #2111 race

Both `POST /api/playlists/<playlist_id>/items` and `POST /playlist/<playlist_id>/add`
allocated the next playlist position with a separate `SELECT COALESCE(MAX(position), 0)`
followed by an `INSERT`. Two concurrent appends for distinct visible videos could both
read the same maximum and commit two rows on the same position, leaving playlist order
ambiguous (and only `(playlist_id, video_id)` uniqueness applied, so the inserts were
both accepted).

This change folds the `MAX(position)` calculation into the same `INSERT...SELECT`
statement so SQLite computes the new position under the same write lock as the row
insert. The two routes now share a single atomic statement:

```sql
INSERT INTO playlist_items (playlist_id, video_id, position, added_at)
SELECT ?, ?, COALESCE(MAX(position), 0) + 1, ?
FROM playlist_items WHERE playlist_id = ?
```

A concurrent duplicate append now hits the existing `idx_playlist_items_uniq` UNIQUE
index and is reported as the documented `409 Video already in playlist` (or
`409 Already in playlist` on the web route) instead of surfacing an `IntegrityError`.
The pre-flight `SELECT 1 FROM playlist_items WHERE playlist_id = ? AND video_id = ?`
check is kept so the common-case duplicate append never has to fall through to the
constraint.

### Tests

New `tests/test_playlist_append_race.py` covers:

- sequential appends keep monotonic positions `1..5`
- single duplicate append returns the documented `409`
- eight threads racing on `INSERT...SELECT` against the same playlist all land on
  distinct positions and produce exactly `[1..8]`
- a dedicated two-thread `test_legacy_two_step_pattern_can_duplicate_positions`
  proves that the **old** `SELECT MAX` + `INSERT` pair still races and produces
  duplicate positions — pinning the failure mode the issue describes
- four concurrent duplicate appends collapse to exactly one `ok` + three `conflict`
  outcomes with a single stored row

`tests/test_playlist_visibility.py` continues to pass unchanged (4/4).

### Risk

The shape of the response is preserved (`{"ok": true, "position": <int>}` for the API
route). The web route still returns `{"ok": true}`. Only the position-allocation
strategy and the duplicate-detection fallback change. The INSERT statement is the
same schema and the `position` column already existed with a covering index
`idx_playlist_items_pl ON playlist_items(playlist_id, position)`.

Fixes #2111.

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -10968,15 +10968,29 @@ def api_add_playlist_item(playlist_id):
     if db.execute("SELECT 1 FROM playlist_items WHERE playlist_id = ? AND video_id = ?", (pl["id"], vid)).fetchone():
         return jsonify({"error": "Video already in playlist"}), 409
 
-    # Get next position
-    max_pos = db.execute("SELECT COALESCE(MAX(position), 0) FROM playlist_items WHERE playlist_id = ?", (pl["id"],)).fetchone()[0]
-    db.execute(
-        "INSERT INTO playlist_items (playlist_id, video_id, position, added_at) VALUES (?,?,?,?)",
-        (pl["id"], vid, max_pos + 1, time.time()),
-    )
+    # Allocate next position atomically in the same statement. Two concurrent
+    # appends previously both read the same MAX(position) in a separate SELECT
+    # and then both INSERTed, colliding on the same position. Folding MAX into
+    # the INSERT lets SQLite compute it under the same write lock as the row
+    # insert so two concurrent requests cannot land on the same position. See
+    # issue #2111.
+    try:
+        db.execute(
+            "INSERT INTO playlist_items (playlist_id, video_id, position, added_at) "
+            "SELECT ?, ?, COALESCE(MAX(position), 0) + 1, ? "
+            "FROM playlist_items WHERE playlist_id = ?",
+            (pl["id"], vid, time.time(), pl["id"]),
+        )
+    except sqlite3.IntegrityError:
+        # (playlist_id, video_id) UNIQUE index — concurrent duplicate append.
+        return jsonify({"error": "Video already in playlist"}), 409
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):
Folds MAX(position)+1 into the INSERT so allocation happens under the write lock; IntegrityError mapped to 409. Closes #2111.

PR #2179, opened 2026-09-03
fix(votes): atomic same-voter writes close #2116 race

### LINKED ISSUE #2116: Concurrent same-voter requests can corrupt vote counters

Vyacheslav-Tomashevskiy:
## Summary

Fixes Scottcjn/bottube#2116. Video and comment vote routes (API-key and web-session variants) issued a SELECT for the existing vote row outside a write transaction, then attempted INSERT/UPDATE. Two concurrent same-voter requests could both pass the existing-vote check, both try to insert the same primary key, and one would fail the UNIQUE constraint on `(agent_id, video_id)` or `(agent_id, comment_id)`, surfacing as an unhandled 500.

## Fix

Wrap the existing-vote read and the subsequent state transition in a `BEGIN IMMEDIATE` write transaction so SQLite serializes competing writers. If the loser of the race still hits the UNIQUE constraint (it should not, but defensive), rollback, re-read the winner's row, and respond with an idempotent 200 carrying `{idempotent: true, your_vote: <winner>}` instead of a 500.

Touches:
- `bottube_server.py`:
  - `vote_comment` (API key, ~line 8061)
  - `web_vote_comment` (session, ~line 8099)
  - `vote_video` (API key, ~line 8232)
  - `web_vote_video` (session, ~line 8316)
- `tests/test_vote_race_2116.py` (new): three regression tests that spin up 8 parallel vote threads per route and assert no 500s, exactly one row in the votes/comment_votes table, and that at least one thread observes `idempotent=True`.

## Verification

Local syntax-check (`python3 -c "import ast; ast.parse(open('bottube_server.py').read())"`) passes. Tests can be run with `pytest tests/test_vote_race_2116.py -v` once the project test fixture (`client`, `app`) is configured per the existing suite.

## Reward claim

Filing under the rustchain-bounties pool Scottcjn/rustchain-bounties#71 (Ongoing Bug Bounty, severity Medium -> 15-50 RTC). Closing #2116.

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -8075,13 +8075,31 @@ def vote_comment(comment_id):
     if vote_val not in (1, -1, 0):
         return jsonify({"error": "vote must be 1 (like), -1 (dislike), or 0 (remove)"}), 400
 
+    # Acquire a write transaction before reading existing vote state to
+    # serialize concurrent same-voter requests (closes #2116).
+    db.execute("BEGIN IMMEDIATE")
     existing = db.execute(
         "SELECT vote FROM comment_votes WHERE agent_id = ? AND comment_id = ?",
         (g.agent["id"], comment_id),
     ).fetchone()
-
-    _apply_comment_vote(db, comment_id, comment["agent_id"], g.agent["id"], vote_val, existing)
-    db.commit()
+    try:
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):
Same-voter vote writes made atomic; regression test exercises the interleaving. Closes #2116.

PR #2180, opened 2026-09-03
fix(referral): atomic first-upload flag close #2153 race

### LINKED ISSUE #2153: Referral first-upload counter can increment twice under concurrency
## Bug

`_referral_mark_first_upload()` reads `referral_first_upload_counted`, then performs an unconditional flag update and counter increment. Two concurrent first-upload requests can both observe `0`, both set the flag, and both increment `referral_codes.first_uploads`, inflating the referral leaderboard and activation evidence.

## Reproduction

1. Create a referred agent with `referral_first_upload_counted = 0`.
2. Run two first-upload completions concurrently so both reads occur before either write.
3. Observe `referral_codes.first_uploads` increment twice for the same invitee.

## Expected

The flag transition is the atomic admission edge: only the request that changes `0 -> 1` may increment the referral counter and refresh the invite state. Later or concurrent calls must be no-ops.

## Proposed fix

Use a conditional update guarded by `referral_first_upload_counted = 0`, gate the counter increment on affected-row count, and add a deterministic concurrent regression test proving exactly-once accounting.

Vyacheslav-Tomashevskiy:
Closes Scottcjn/bottube#2153 — Referral first-upload counter can increment twice under concurrency.

## Problem

`_referral_mark_first_upload()` reads `referral_first_upload_counted` and then performs an unconditional flag UPDATE followed by a counter increment. Two concurrent first-upload requests could both observe `flag = 0`, both flip it, and both increment `referral_codes.first_uploads` — inflating the referral leaderboard and activation evidence.

## Fix

The flag transition is now expressed as a conditional UPDATE (`... WHERE referral_first_upload_counted = 0`) so only the request that actually flips `0 -> 1` may proceed to the counter bump. The first_uploads increment is gated on `cursor.rowcount == 1`. SQLite serialises the UPDATE write lock, so the second concurrent writer sees `flag = 1` and matches 0 rows.

Return value now reports `{"applied": true|false, "reason": ...}` for tests and future audit logging.

## Tests

`tests/test_referral_first_upload_race_2153.py` (new):
- `test_concurrent_first_upload_increments_counter_exactly_once` — 8 concurrent threads → exactly 1 caller flips the flag, counter = 1
- `test_sequential_second_call_is_noop` — second call after flag set is a no-op
- `test_no_referral_code_is_noop` — agent with no `referred_by_code` is a no-op

All 20 existing referral tests still pass.

## Bounty

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -957,32 +957,75 @@ def _referral_apply_signup(db, new_agent_id: int, code: str, *, source: str = "s
 
 
 def _referral_mark_first_upload(db, agent_id: int):
-    """If agent was referred, count their first upload exactly once (best-effort)."""
+    """If agent was referred, count their first upload exactly once (best-effort).
+
+    Issue #2153: the previous SELECT-then-UPDATE pattern let two concurrent
+    first-upload requests both observe `referral_first_upload_counted = 0`,
+    both flip the flag, and both increment `referral_codes.first_uploads` —
+    inflating the referral leaderboard and activation evidence.
+
+    Fix: the flag transition is expressed as a conditional UPDATE
+    (`... WHERE referral_first_upload_counted = 0`) so only the request that
+    actually flips 0->1 may proceed to the counter bump. The first_uploads
+    increment is gated on cursor.rowcount == 1 — if the conditional update
+    affected 0 rows we know another request already won the race and we
+    exit without touching the counter.
+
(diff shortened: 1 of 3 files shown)


Scottcjn (approved):
Atomic first-upload referral flag; stacked on #2179 which is merged. Closes #2153.

PR #1789, opened 2026-08-28
fix: validate tip JSON bodies and finite amounts

Vyacheslav-Tomashevskiy:
## Summary
- add one shared tip payload parser for all four RTC tip handlers
- reject non-object JSON bodies before any `.get(...)` access
- reject boolean, null, NaN, and infinite `amount` values before ledger debit or on-chain forwarding
- keep valid numeric tips working for API-key and web-session routes

## Testing
- `pytest -q tests/test_tip_json_amount_validation.py tests/test_bounty_2161_collab_tip_split.py`
- `python3 -m py_compile bottube_server.py tests/test_tip_json_amount_validation.py`
- `git diff --check`

/claim #1786

PR #2182, opened 2026-09-04
fix(comments): atomic same-voter writes close #2145 race

### LINKED ISSUE #2116: Concurrent same-voter requests can corrupt vote counters

### LINKED ISSUE #2145: Comment vote races can return 500 and drift counters
## Summary

API-key and web comment voting read the voter's existing row before entering `_apply_comment_vote()`, then use check-then-insert/update logic and increment/decrement cached counters. Concurrent requests can both observe no vote, both increment `comments.likes`, and then race on the `(agent_id, comment_id)` primary key; one request escapes as an integrity error while counters can drift from authoritative vote rows.

## Affected routes

- `POST /api/comments/<comment_id>/vote`
- `POST /api/comments/<comment_id>/web-vote`

## Expected

Comment vote application should be idempotent and transaction-safe. Use an upsert/delete for the authoritative vote row, then derive `likes` and `dislikes` from those rows in the same transaction. Identical retries must stay successful without incrementing twice or returning HTTP 500; vote changes and removals must leave cached counters exact.

## Regression

Simulate two callers acting on the same stale `existing=None` snapshot. The current helper raises on the second insert. The fixed helper should preserve one vote row and `(likes, dislikes) == (1, 0)`, then correctly change to dislike and remove.

Vyacheslav-Tomashevskiy:
Closes #2145 — comment vote races can return 500 and drift counters.

## The bug
Both `POST /api/comments/<id>/vote` and `POST /api/comments/<id>/web-vote` read the voter's existing row, then ran a check-then-INSERT/UPDATE on `comment_votes`, and incremented `comments.likes` / `comments.dislikes` outside any locking. Two concurrent same-voter requests could both observe `existing=None`, both increment `comments.likes`, then race on the UNIQUE(agent_id, comment_id) primary key — the loser escaped as HTTP 500 (sqlite3.IntegrityError) and the cached counter drifted from the authoritative vote rows.

## The fix
Wrap the existing-row SELECT and `_apply_comment_vote()` in `BEGIN IMMEDIATE` so SQLite serializes same-voter writers. On the losing race (`sqlite3.IntegrityError`) the handler rolls back, re-reads the winner's row, and re-derives (likes, dislikes) from `comment_votes` in the same transaction, so cached counters stay exact.

This is the same atomic-write pattern that closed #2116 (PR #2179) — applied here to the comment-vote routes.

## Regression test
`tests/test_comment_vote_race_2145.py` — 3 tests, all passing locally:
1. `test_concurrent_same_voter_vote_no_500` — 8 same-voter threads must all return 2xx, leave exactly one `comment_votes` row, and `comments.likes` must equal `SUM(comment_votes.vote == 1)` (no drift).
2. `test_idempotent_flag_observable_on_race` — re-derives (likes, dislikes) from the authoritative vote rows on a losing race.
3. `test_sequential_noop_removes_vote` — like → remove → like must end with exactly 1 like, 0 dislikes, no drift.

Local: `python3 -m pytest tests/test_comment_vote_race_2145.py -q` → **3 passed**.

## Files changed
- `bottube_server.py` — vote_comment, web_vote_comment, _apply_comment_vote (atomic-txn + IntegrityError fallback)
- `tests/test_comment_vote_race_2145.py` — new regression suite

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -48,6 +48,12 @@ from flask import (
     url_for,
 )
 from markupsafe import Markup, escape
+from bottube_validators.validators import (
+    MAX_QUERY_TIMESTAMP,
+    parse_enum as parse_enum_param,
+    parse_positive_int as parse_int_param,
+    parse_timestamp_iso as parse_ts_param,
+)
 from werkzeug.security import check_password_hash, generate_password_hash
 
 # Mood Engine for Agent Mood System (Bounty #2283)
@@ -64,6 +70,7 @@ try:
 except ImportError:
     VISION_SCREENING_ENABLED = False
     def screen_video(video_path, run_tier2=True):
+        """Screen a video file through moderation tiers. Args: video_path: Path to video file. run_tier2: Whether to run tier-2 deep analysis. Returns: Screening result dict with verdict and confidence."""
         return {"status": "pending_review", "tier_reached": 0, "summary": "screening module not available"}
 
 
@@ -87,6 +94,7 @@ _ctr_tracker = None
 _ab_manager = None
 
 def _get_ctr_tracker():
+    """Get or lazily initialize the CTR tracking singleton. Returns: CTR tracker instance."""
     global _ctr_tracker
     if _ctr_tracker is None:
         from thumbnails.ctr_tracker import CTRTracker
@@ -95,6 +103,7 @@ def _get_ctr_tracker():
     return _ctr_tracker
 
 def _get_ab_manager():
+    """Get or lazily initialize the A/B test manager singleton. Returns: AB manager instance."""
     global _ab_manager
     if _ab_manager is None:
         from thumbnails.ab_test import ABTestManager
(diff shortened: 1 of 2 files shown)


Scottcjn:
Holding this one for a rebase, @Vyacheslav-Tomashevskiy. The diff is +400/−170 on bottube_server.py for what the title describes as a filter/atomic-write fix, and it carries content that is not this fix — a batch of one-line docstrings and the debit_rtc helper from #1789. Please rebase onto current `main` so the PR is only the vote/comment change plus its test; then it merges on the next pass like #2177/#2179/#2183 did today. — Sophia

Scottcjn:
Landed via #2190 — slimmed to the comments race fix (the counter re-derivation on a lost race is yours and is what went in); the carried docstrings and older body-parsing took main's side. Authorship preserved. Closes #2145. — Sophia

PR #2184, opened 2026-09-05
fix(referral): atomic fingerprint dedup for #2158 race (closes #2158)

### LINKED ISSUE #2158: Concurrent referral hits can bypass fingerprint deduplication
## Bug

`_referral_touch_hit_unique()` uses a read-then-write sequence for the 24-hour `(referral code, fingerprint)` window. Two concurrent first hits can both observe no row (or the same expired row). Although one `INSERT OR IGNORE` may lose, both requests still increment `referral_codes.hits`, so the supposedly unique referral metric is inflated.

## Reproduction

1. Use the same referral code and request fingerprint from two concurrent requests.
2. Synchronize both reads before either writes.
3. Observe only one uniqueness row/window transition, but two increments of `referral_codes.hits`.

## Expected

The insert or expired-window transition must be the atomic admission edge. Only the request whose insert/update changes a row may increment the referral hit total.

## Proposed fix

- gate a new fingerprint on `INSERT OR IGNORE` affected-row count
- gate an expired fingerprint on a conditional timestamp update that still matches the expired window
- increment the referral total only for the winner
- add deterministic concurrent regressions for both new and expired fingerprints

Vyacheslav-Tomashevskiy:
Closes #2158.

`_referral_touch_hit_unique()` used a read-then-write sequence for the
24-hour `(referral code, fingerprint)` window. Two concurrent first hits
could both observe no row (or the same expired row) and both increment
`referral_codes.hits`, inflating the supposedly unique metric. Even when
the duplicate `INSERT OR IGNORE` lost, the increment still ran for both
requests, so the legacy contract did not protect downstream counters.

This change folds the new-fingerprint insert and the expired refresh
into a single atomic SQLite `INSERT ... ON CONFLICT(code, fp_hash) DO
UPDATE SET last_hit_at = excluded.last_hit_at WHERE
referral_hit_uniques.last_hit_at < cutoff` operation. SQLite only
reports `rowcount == 1` for the request that actually mutated the row,
so the request that elects to refresh (or insert) is the only one that
may increment the referral hit total. A second concurrent request that
finds the row already refreshed within the 24h window is a no-op.

Adds `tests/test_bounty_2158_referral_hit_unique_atomic.py` with five
regression cases:

- first hit increments exactly once
- replaying a fresh fingerprint within 24h is a no-op
- expired fingerprint increments exactly once per window
- 8-thread concurrent first hits yield exactly one increment
- 6-thread concurrent expired refreshes yield exactly one increment

Local: `pytest tests/test_bounty_2158_referral_hit_unique_atomic.py
tests/test_referrals.py` -> 16/16 passed.

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -830,23 +830,22 @@ def _referral_touch_hit_unique(db, code: str):
         fp_hash = hashlib.sha256(fp.encode("utf-8")).hexdigest()
         now = time.time()
         cutoff = now - 86400
-        row = db.execute(
-            "SELECT last_hit_at FROM referral_hit_uniques WHERE code = ? AND fp_hash = ?",
-            (code, fp_hash),
-        ).fetchone()
-        if row and float(row["last_hit_at"] or 0) > cutoff:
+        # Atomic admission edge: a single UPSERT either inserts a new fingerprint
+        # or refreshes an expired one. SQLite returns rowcount == 1 only for the
+        # request that actually changed the row, so concurrent first-hits and
+        # concurrent expired-refreshes each elect exactly one winner.
+        cur = db.execute(
+            "INSERT INTO referral_hit_uniques (code, fp_hash, last_hit_at) "
+            "VALUES (?, ?, ?) "
+            "ON CONFLICT(code, fp_hash) DO UPDATE SET "
+            "  last_hit_at = excluded.last_hit_at "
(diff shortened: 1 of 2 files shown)


Scottcjn (approved):
Atomic fingerprint dedup within the 24h window; the focused fix for #2158 (the 25k-line rewrite that raced it was closed).

PR #2204, opened 2026-09-06
fix(playlists): share one atomic append operation close #2141

### LINKED ISSUE #2141: Playlist add races can return 500 and duplicate positions
## Summary

Both playlist-add routes use a check-then-act sequence for duplicate detection and position assignment. Concurrent requests can pass the duplicate check together; the loser then hits the unique index as an uncaught `sqlite3.IntegrityError` (HTTP 500). Concurrent requests for different videos can also read the same `MAX(position)` and persist duplicate positions.

## Affected routes

## Cause

Each route separately executes:

1. `SELECT 1` to check `(playlist_id, video_id)`
2. `SELECT MAX(position)`
3. a plain `INSERT`

Those statements are not one atomic operation. The unique index protects duplicate membership, but its expected race outcome is not handled. Position allocation has no unique constraint at all.

## Expected

Use one shared atomic insert operation whose position is computed inside the write statement. A duplicate must deterministically return the route's existing 409 contract, and distinct additions must receive consecutive positions. API and web routes should not drift.

## Regression

Exercise the shared operation with an initial add, duplicate add, and second distinct add; assert inserted/duplicate outcomes and positions `1, 2`. Preserve the existing route response contracts.

Vyacheslav-Tomashevskiy:
Closes #2141 — both playlist-add routes (`POST /api/playlists/<playlist_id>/items` and `POST /playlist/<playlist_id>/add`) now delegate to one shared `_append_playlist_item()` helper instead of each carrying its own inline copy of the duplicate check, the position allocation, and the `IntegrityError` translation.

## The bug

Each route ran the same three steps as separate statements:
1. `SELECT 1` to check `(playlist_id, video_id)` membership
2. `SELECT MAX(position)`
3. a plain `INSERT`

Two concurrent appends could pass the duplicate check together, then the loser hit the unique index as an uncaught `IntegrityError` (HTTP 500). Concurrent appends for *different* videos could also both read the same `MAX(position)` and persist duplicate positions. And because the logic was duplicated across both routes, a fix to one silently left the other behind.

## The fix

`_append_playlist_item(db, playlist_row_id, video_id)` wraps the membership check and the insert in one `BEGIN IMMEDIATE` write transaction, with `COALESCE(MAX(position), 0) + 1` computed inside the `INSERT` itself so two concurrent appends can't land on the same position. The `UNIQUE(playlist_id, video_id)` index stays the authority on membership — a duplicate that still loses the race raises `IntegrityError`, translated into the same `None` conflict sentinel instead of escaping as a 500. Both routes call this one helper and keep their existing response contracts (201 + position / 409 for the API route, 200 + ok / 409 for the web route).

Regression tests in `tests/test_playlist_shared_append_2141.py` (5 cases): sequential appends get consecutive positions, a duplicate append returns the conflict sentinel, a burst of concurrent appends for distinct videos never collides on position, and the same for a concurrent duplicate-append race. All pass locally.

## Heads-up on the base

This branch is built on our fork's own current `main` (it had fallen behind upstream, and the token I push with lacks `workflow` scope, which blocks syncing since upstream's history in between touches `.github/workflows/*`). Upstream `main` already carries #2177's position-only fix for the same two routes, done as two separate inline copies rather than a shared helper — so this PR's diff against current upstream `main` will likely show as a textual conflict on those functions rather than a clean fast-forward, even though the two changes solve overlapping problems. Happy to rebase by hand against current main if that's easier to review — the net effect is the same fix, consolidated into one shared helper so the two routes can't drift apart again.

RTC address for payout: `RTCd1554f0f35576faf01d386a6be1c947f560dd0b7`

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -10937,6 +10937,65 @@ def api_delete_playlist(playlist_id):
     return jsonify({"ok": True})
 
 
+def _append_playlist_item(db, playlist_row_id, video_id, now=None):
+    """Append ``video_id`` to a playlist as one atomic operation.
+
+    Returns the allocated ``position`` when a row was inserted, or ``None`` --
+    the conflict sentinel -- when ``(playlist_id, video_id)`` is already a
+    member of the playlist.
+
+    Both playlist append routes previously inlined their own copy of the
+    duplicate check, the position allocation and the ``IntegrityError``
+    handling, so the API and web surfaces were free to drift apart: a fix
+    applied to one route silently left the other behind. Routing both through
+    this helper keeps a single definition of the operation.
+
+    The write is wrapped in ``BEGIN IMMEDIATE`` so the membership check and the
+    insert happen under one write lock -- concurrent appends are serialized
+    instead of interleaving between the check and the act. Within that lock
+    ``COALESCE(MAX(position), 0) + 1`` is computed inside the ``INSERT``
+    statement itself, so two appends can never allocate the same slot. The
+    ``UNIQUE(playlist_id, video_id)`` index remains the authority on
+    membership: a duplicate that still loses the race raises
+    ``IntegrityError``, which is translated into the same ``None`` sentinel as
+    the pre-check rather than escaping as an HTTP 500.
+
+    See issue #2141.
+    """
+    if now is None:
+        now = time.time()
+
+    db.execute("BEGIN IMMEDIATE")
+    try:
+        if db.execute(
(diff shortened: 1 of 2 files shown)


Scottcjn:
@Vyacheslav-Tomashevskiy this conflicts with main after the weekend's merges on `bottube_server.py`. Please rebase; the change itself (one atomic append for playlists, closes #2141) is wanted. — Sophia

Scottcjn:
Thanks — this looks good in principle, but it currently has a merge conflict (mergeable=DIRTY). Please rebase onto main and we'll merge it.

PR #2220, opened 2026-09-08
[claude] Add ModelRunner as a text-to-video provider

hakankaan:
Adds [ModelRunner](https://modelrunner.ai) as an optional text-to-video backend.

**Disclosure:** I work on ModelRunner, and this PR was written with Claude Code. CONTRIBUTING.md invites agent contributions with a name-prefixed title, hence the `[claude]` prefix — happy to change either if you'd prefer.

## What this does

ModelRunner is a hosted inference API with a queue-based interface, so it drops into the existing submit → poll → download shape the fal and Replicate backends already use. It's added in both places providers live:

- `generation/providers/modelrunner.py` — implements `GenerationProvider`, registered in `generation/worker.py`
- `_try_modelrunner()` in `video_gen_blueprint.py` — registered with `requires_key_env="MODELRUNNER_KEY"`

I added it to both because every other backend exists in both, and `video_gen_blueprint` is the one `bottube_server.py` actually registers. Let me know if you'd rather it went in only one.

**No new dependencies** — stdlib `urllib` only, matching the sibling adapters.

## Configuration

| Variable | Purpose |
| --- | --- |
| `MODELRUNNER_KEY` | Enables the provider. Without it, `register()` skips it and nothing changes for existing deployments. |
| `MODELRUNNER_VIDEO_MODEL` | Selects the endpoint. Defaults to `wan-video/wan/v3.0/text-to-video` (2–30s, 480P/720P/1080P). |

## Choices worth flagging

- **Duration is clamped to 2–30s.** `GenerationRequest` permits 1–300s; the model's window is narrower, so requests are clamped rather than rejected.
- **Audio is requested as `false`.** The model returns a real soundtrack, but `_reencode_to_square` replaces the audio track for every provider, so asking for it would mean paying for something that gets discarded. If you ever want an audio-preserving path, this is the line to flip.
- **`cost_tier=3`, not `1`.** Unlike the free-tier backends this is a paid API, and it seemed wrong to let it outrank them in the router's scoring on price.
- **Output parsing accepts several shapes.** Video models return the URL as a bare string today, but some wrap it — handling both is what lets `MODELRUNNER_VIDEO_MODEL` point at a different model without a code change.

## Testing

`tests/test_modelrunner_provider.py` — 32 tests, all offline, no network calls:

- availability gating on the API key, in both registries (including the negative case: no key → not registered)
- duration clamping across the full range `GenerationRequest` allows
- the queue's status vocabulary mapped onto `pending`/`running`/`completed`/`failed`, plus unreachable-queue handling
- output URL extraction across the shapes models return
- submit posts to the configured model with `Key` auth, and fails cleanly when no request id comes back

```
tests/test_modelrunner_provider.py .......  32 passed
tests/test_video_providers.py + test_generation_routes.py
  + test_generation_reliability.py + test_generation_job_duration_validation.py
                                       61 passed total
```

`pycodestyle --max-line-length=120` is clean on both new files.

Five failures elsewhere in the video test suite (`test_video_tips_api`, `test_similar_missing_video`, `test_video_metadata_update_validation`) are present on an unmodified checkout of `main` too — not from this change.

## What I could not verify

I don't have a key on the machine I built this on, so **there is no live end-to-end generation behind this PR**. The payload is validated against the model's published input schema and the output parsing is checked against the shape real completed requests return, but the full round-trip is unproven. Happy to run one and post the result before you merge, if you'd like that first.

---

### 🤝 Partnership & contact

This PR comes from the **ModelRunner** team. Two things beyond the code, in case they're useful:

- **Free API credits for this project.** We run an [Open Source Program](https://modelrunner.ai/oss-program) that grants credits to maintainers of open-source projects — the tiers are on the page, and this project sits comfortably inside them. That offer stands whether or not you merge this PR; it isn't conditional on the integration.
- **Happy to revise anything here** to match your conventions — naming, structure, scope, or splitting it up. Leave a comment and I'll push a change.

Reach us anytime:
- 📧 **support@modelrunner.ai**
- 🌐 https://modelrunner.ai
- 📚 Docs: https://modelrunner.ai/docs


hakankaan:
Two updates.

**Docs added.** I'd missed `docs/VIDEO_GENERATION_GUIDE.md` in the first push — the backend is now in the reference table, the per-backend walkthrough and the environment variable block. One judgement call worth your eye: Part 2 is titled "Free-Tier Video Generation Backends" and ModelRunner is paid, so I marked it clearly as paid and changed the intro to "seven backends... six are free-tier or self-hosted". If you'd rather it sat in its own section outside Part 2, say so and I'll move it.

**On the failing `deployment-drift` check** — I don't believe it's from this PR. The same job fails on `main`'s own CI run at `e0c09a6`, the commit this branch is based on, with `test`, `lint` and `security` green there just as they are here:

- `test_deployment_drift.py::test_repository_configs_are_valid_offline` — `'fail' == 'pass'`
- `test_deployment_drift.py::test_issue_1410_config_reports_exact_live_only_failure_state` — `assert 5 == 4`

Both reproduce on a clean checkout of `main` with none of this branch's code present. Nothing here touches deployment config. Happy to look into it separately if it'd help, but I didn't want to fold an unrelated fix into this PR.

Scottcjn (changes requested):
Independent adversarial review (Codex) + maintainer read: **close, with three blocking fixes.** CI is green against the repaired main, but the code cannot currently produce a video.

1. **ffmpeg re-encode never works.** Output options precede the second input, so ffmpeg rejects the command and `get_result()` always returns `None` (modelrunner.py L151–172). Nothing in the tests exercises `get_result`, which is why this passed.
2. **Regression of the #2211 fix.** Every polling exception becomes `pending`, potentially forever (L111–125) — the exact ComfyUI bug fixed two days ago. The test at L164–174 codifies the wrong behavior. Return a bounded, distinct transient-failure signal like `comfyui_ltx.get_status()` now does.
3. **Pin network destinations.** `urllib` follows redirects and can forward `Authorization` across hosts, and the provider-returned `video_url` is fetched from any host (L92–101, L128–146, L176–188). Reject cross-host redirects and allowlist the HTTPS media host(s).

Good: the key comes only from `MODELRUNNER_KEY` at import and is never logged; base URL is constant and the model is operator config; existing ComfyUI/LTX paths are untouched (worker L154–166 inserts ModelRunner before the title-card fallback); no credit or moderation writes. Add tests for `get_result`/ffmpeg, redirect handling, and the bounded-outage path. Ping when pushed; this is a wanted provider. — Sophia

hakankaan:
Thanks for the thorough pass — all three were real. Pushed on a rebase onto current `main` (the branch had been cut before #2211 landed, so it now sits on the repaired drift check and the `comfyui_ltx` reference).

**1. ffmpeg re-encode.** Confirmed exactly as described — ffmpeg 8.1 rejects the old order with *"Option vf … cannot be applied to input url anullsrc … you are trying to apply an input option to an output file"* and never opens the output. Both `-i` inputs are now declared before any output option, with explicit `-map 0:v:0 -map 1:a:0` so the silent track replaces whatever the model returned (verified: a 440 Hz test tone came out at −91 dB), and the return code is checked and logged instead of trusting `out_path.exists()`. One thing worth knowing: I had copied that argument order from `fal_ai.py`, and `replicate.py`, `huggingface.py` and the blueprint's shared `_reencode_to_square` all carry the same misordering, so those paths can't produce a video either. I left them alone here to keep this PR to its own provider — happy to send that as a separate PR if you'd like.

**2. Poll-error budget.** `get_status()` now mirrors `comfyui_ltx.get_status()`: a per-request counter of consecutive transport/HTTP/malformed-body failures, reset by any poll that reaches the queue, and `"failed"` on the 5th in a row. A genuinely queued job (`IN_QUEUE`, which on a cold start can outlast that budget) never trips it. The old test that codified the wrong behaviour is replaced by four covering the budget, the reset, the malformed-body case and the long-queue case.

**3. Network pinning.** Every call goes through a module opener whose redirect handler raises instead of following — I refused all 3xx rather than only cross-host ones, since the queue API never redirects and that leaves no carve-out to reason about. The finished clip is downloaded only from an HTTPS URL whose parsed hostname is `media.modelrunner.ai` or a subdomain (so `media.modelrunner.ai.evil.example`, `media.modelrunner.ai@evil.example` and a plaintext URL are all refused), with no `Authorization` header on that fetch. The blueprint's `_try_modelrunner` now routes its calls and download through the same helpers so the two entry points can't drift. Tests cover the stock handler forwarding the key off-host (the hazard), a synthetic 302 through the refusing opener ending in an error after exactly one hop, the host matcher edge cases, and `get_result` refusing an off-host result without fetching it.

Also added the get_result/ffmpeg tests you asked for, including one against a real ffmpeg (skips when it isn't installed, as on CI).

PR #2221, opened 2026-09-08
Harden deposit bridges: stop unbacked RTC mint (Ergo) + deposit theft (wRTC)

Scottcjn:
## Summary

Defensive audit of the **ERG** and **wRTC** deposit→mint bridges (GPT-6 Astra; **every finding source-verified** before the fix). These endpoints mint internal RTC from external crypto deposits, so a verification gap creates **unbacked RTC** (breaks the fixed-supply premise).

### Fixes
- **D1 — Ergo repeat-mint (CRITICAL).** `ergo_bridge_blueprint.py`: the client `tx_id` was used as **both** the on-chain lookup (`/transactions/{tx_id}`) **and** the `UNIQUE` dedup key, with no validation. An alias (case variant, URL fragment, etc.) resolves to the **same on-chain tx** but a **different** dedup key → the same deposit mints repeatedly. **Fix:** canonicalize + strictly validate `tx_id` to 64-hex lowercase before use, so aliases collapse to one key.
- **D3 — wRTC deposit theft (HIGH).** `wrtc_bridge.py`: `bridge_deposit` credited the authenticated caller for **any** valid `tx_signature`, with no binding to the depositor. Solana signatures are public, so anyone watching the reserve could **front-run and steal** another user's deposit. **Fix:** require the on-chain sender (`info["from_address"]`) to equal the account's linked `sol_address` — mirrors the Ergo bridge (whose comment already claimed "same rule as the Solana / Base bridges", but the check was missing here).
- **D2 — wRTC reserve-ATA spoof (HIGH).** `_is_reserve_ata` fell back to a token account's **current** owner, so an attacker could reassign an account's owner to the reserve **after** a historical transfer into it and claim that transfer. **Fix:** trust only the deterministic derived reserve ATA. (`_check_account_owner` is now unused; left in place.)
- **D4 — wRTC self-transfer credit.** Reject reserve-originating/self transfers so an internal reserve movement can't be claimed as a deposit. **Follow-up (flagged, not in this PR):** credit the reserve's **pre/post net token-balance change** from tx `meta` rather than the instruction's stated amount — that's a Solana-meta rewrite that needs live-tx tests.

### Verification / notes
- Each finding was checked against source; `ast.parse` clean on both files; **CRLF line endings preserved** (diff is 29+/5−, not a reformat).
- Reviewer note: the automated tri-brain reviewer was unavailable this session — these rest on manual source-verification; a second human review before merge is warranted given the money path.
- **Operational:** if the wRTC on-ramp is currently paused/unfunded, live exploitability is limited, but the code defects are real wherever the endpoints are reachable.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01KbyXP4eiiRYEa8GsQtQPhR

### DIFF
--- a/ergo_bridge_blueprint.py
+++ b/ergo_bridge_blueprint.py
@@ -401,6 +401,14 @@ def ergo_deposit():
     if not tx_id:
         return jsonify({"error": "tx_id required"}), 400
 
+    # SECURITY (D1): canonicalize + strictly validate tx_id BEFORE it is used
+    # as both the on-chain lookup and the UNIQUE dedup key. Without this an
+    # alias (case variant, URL fragment, etc.) resolves to the same on-chain
+    # tx but a different dedup key -> the same deposit is minted repeatedly.
+    tx_id = tx_id.lower()
+    if len(tx_id) != 64 or any(ch not in "0123456789abcdef" for ch in tx_id):
+        return jsonify({"error": "tx_id must be 64 hexadecimal characters"}), 400
+
     # Check if already claimed
     existing = db.execute(
         "SELECT id FROM ergo_deposits WHERE tx_id = ?", (tx_id,)
--- a/wrtc_bridge.py
+++ b/wrtc_bridge.py
@@ -235,6 +235,8 @@ def verify_solana_deposit(tx_signature: str):
                 # Verify destination is our reserve wallet's ATA
                 if not _is_reserve_ata(dest_ata):
                     continue
+                if _is_reserve_ata(source_ata):
+                    continue  # D4: ignore reserve-originating/self transfers
 
                 return {
                     "tx_signature": tx_signature,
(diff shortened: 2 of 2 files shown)


Scottcjn:
⚠️ Correction after further checking: the file this PR modifies, `wrtc_bridge.py` (`wrtc_bridge_bp`), is **not registered anywhere** — `bottube_server.py` wires up `wrtc_bridge_blueprint.py` and `base_wrtc_bridge_blueprint.py` instead. So the **wRTC D2/D3/D4 changes here are hardening of dead code**, not a live exploit fix — apologies for the overstated framing in the description.

What still stands: the **Ergo `tx_id` canonicalization (D1)** is on `ergo_bridge_blueprint.py`, which **is** registered/live — that fix is real and worth merging.

Separately, the live wRTC on-ramps are being **disabled entirely** for securities-law (Howey) reasons in #2222. Given that, the reviewer may prefer to **split this PR** to keep only the Ergo D1 fix.

PR #2240, opened 2026-09-11
fix(notifications): validate pagination query parameters strictly

### LINKED ISSUE #2091: Notification APIs silently rewrite invalid pagination

Visio-Vanitas:
## Summary
Replaces silent pagination parameter coercion in `/api/agents/me/notifications` and `/api/notifications` with strict positive integer validation.

## Changes
- Use `_parse_positive_int_query` for `page` and `per_page` query parameters.
- Return HTTP 400 with a descriptive error message on negative, non-integer, or out-of-range values.

Closes #2091

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -11522,8 +11522,12 @@ def web_notification_list():
     if not g.user:
         return jsonify({"error": "Login required", "login_required": True}), 401
 
-    page = max(1, request.args.get("page", 1, type=int))
-    per_page = min(50, max(1, request.args.get("per_page", 20, type=int)))
+    page, error = _parse_positive_int_query("page", 1, max_value=10000)
+    if error:
+        return error
+    per_page, error = _parse_positive_int_query("per_page", 20, max_value=50)
+    if error:
+        return error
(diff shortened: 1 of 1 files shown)


Scottcjn:
## RTC Reward (sent manually)

The automatic merge reward for this PR failed silently: our reward workflow's credentials had gone stale (being fixed now; failures are also now announced on the PR instead of passing silently). We've now paid it by hand: **2 RTC**, pending_id **4664** (24h hold, then confirmed). Sorry for the delay, @Visio-Vanitas, and thank you for the contribution. — Sophia Elya

PR #2243, opened 2026-09-11
Fix /tips/dashboard registration on direct startup

### LINKED ISSUE #2242: [BUG] /tips/dashboard returns 404 although the route exists on main

zaguzovmaksim0-hue:
Fixes #2242.

`/tips/dashboard` was declared after the blocking `if __name__ == "__main__": app.run(...)` block. When BoTTube is launched with the documented `python3 bottube_server.py` path, execution enters `app.run()` before the decorator is reached, so the route is never registered.

This moves the direct-entrypoint block to the real end of the module and adds a focused regression test that pins route-definition order. Import/gunicorn behavior is unchanged.

Validation:
- `python3 -m pytest -q tests/test_tips_dashboard_direct_entrypoint.py` — 1 passed
- `python3 -m py_compile bottube_server.py tests/test_tips_dashboard_direct_entrypoint.py`
- `git diff --check`

AI assistance was used for investigation and implementation; the live public route was reproduced as HTTP 404 three times before the patch.

### DIFF
--- a/bottube_server.py
+++ b/bottube_server.py
@@ -26115,13 +26115,6 @@ def api_embed_discovery():
     return oembed()
 
 
-if __name__ == "__main__":
-    init_db()
-    print(f"[BoTTube] Starting on port 8097 - v{APP_VERSION}")
-    print(f"[BoTTube] DB: {DB_PATH}")
-    print(f"[BoTTube] Videos: {VIDEO_DIR}")
-    app.run(host="0.0.0.0", port=8097, debug=False)
-
 @app.route("/tips/dashboard")
 def tips_dashboard():
     db = get_db()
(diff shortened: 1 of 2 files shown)


MrHyplex9511:
Verified the root cause on `main` by reading the file: `@app.route("/tips/dashboard")` is declared at line 26162, two lines *after* the `if __name__ == "__main__": ... app.run(...)` block at 26155-26160. So `python3 bottube_server.py` enters `app.run()` before the decorator executes and the route is never registered — exactly as reported. The fix direction (moving the entrypoint block below the route) is correct.

Two observations that might strengthen it:

1. The regression test pins *source-text ordering* (`source.index(...)`), which is import-free — sensible, since `bottube_server` constructs the Flask app at module scope and any import-time DB/tracker singletons would make a real import heavy. But a stronger assertion that survives renames/refactors is a runtime one, e.g.:
   ```python
   import bottube_server
   assert any(r.rule == "/tips/dashboard" for r in bottube_server.app.url_map.iter_rules())
   ```
   It catches not just "declared after run" but "route renamed/hand-removed" too. Worth adding alongside the text test if module import stays side-effect-light (the CTR/AB singletons are already lazy per the source comments).

2. This is the second ordering-sensitive failure of this exact shape — the real bug class is "route declared after the entrypoint ends up permanently dead under direct startup". Moving `app.run()` out of the module entirely (a `run.py` entrypoint, or `flask --app bottube_server run`) would make that class impossible rather than fixing one route at a time. Could be worth a follow-up ticket if `python bottube_server.py` is a documented/supported start path beyond this one fix.

PR #2252, opened 2026-09-12
fix(a11y): name the iframe the embed generators hand out

### LINKED ISSUE #2251: a11y: watch-page embed generator outputs iframe without accessible title
## Summary

The watch-page **Embed** generator produces iframe markup without a `title` attribute / accessible name. A user who copies the generated snippet into a site gets an unnamed iframe for screen-reader navigation.

Live reproduction: https://bottube.ai/watch/GAdSoP9G9Q2  
Reviewed source: `Scottcjn/bottube@00973f5b3d2098ad42404afb0cce3945d1eead85`.

## Steps to reproduce

1. Open any current BoTTube watch page, e.g. https://bottube.ai/watch/GAdSoP9G9Q2.
2. Open the Share panel and reveal the Embed panel.
3. Inspect the generated text in the **Embed code** field.

Current generated output has this shape:

```html
<iframe src="https://bottube.ai/embed/VIDEO_ID" width="..." height="..." frameborder="0" allowfullscreen></iframe>
```

There is no `title` attribute.

The live page was fetched again on 2026-09-12 and its `updateEmbedCode()` implementation still generates that exact iframe form. The same code is present on current main:

https://github.com/Scottcjn/bottube/blob/00973f5b3d2098ad42404afb0cce3945d1eead85/bottube_templates/watch.html#L3690-L3696

## Expected

BoTTube's copyable embed snippet should include a stable accessible name, for example:

```html
<iframe ... title="BoTTube video" ...></iframe>
```

or a safely escaped title derived from the video title.

## Actual

The generator omits `title`, so copied BoTTube embeds are unnamed if the embedding site does not manually repair the snippet.

## Accessibility impact

Screen-reader users navigating frames need a concise accessible name to understand the frame's purpose without entering it. Because BoTTube itself generates and offers this markup for copy/paste, the inaccessible output propagates to sites using the official embed snippet.

Relevant WCAG considerations: **4.1.2 Name, Role, Value** and **1.3.1 Info and Relationships**.

## Duplicate check

I checked the current BoTTube issue list and #1618 bounty history. Existing iframe-name reports cover `/community` and the `/badges` verification preview (#2238). I found no report for the distinct watch-page embed-code generator.

## Suggested fix

Add a `title` attribute in `updateEmbedCode()` (prefer a static safe fallback if using the video title would require additional escaping), and add a small template/JS regression test that asserts generated iframe markup contains an accessible name.

Environment: public production page + current source review; no authentication or write request needed. AI assistance disclosed.

L4XB:
## What

The watch page's **Embed code** field generated an iframe with no accessible name, and that snippet is what a creator copies onto their own site, so the unnamed frame travels with it. A screen reader announces frames by name; an unnamed one gives the reader nothing to decide on before entering it.

Fixes #2251.

## How

`updateEmbedCode()` now names the frame with the video's title, escaped for an HTML attribute, and falls back to a plain `BoTTube video` when there is no title:

```
<iframe src="https://bottube.ai/embed/GAdSoP9G9Q2" title="BoTTube video: How agents learn" width="853" height="480" frameborder="0" allowfullscreen></iframe>
```

The title rather than a fixed string, because a page embedding several BoTTube videos would otherwise present a row of frames that all announce identically, which is exactly the case where a name stops helping.

Two escaping layers, because a title is free text:

- it reaches the script through Jinja's `tojson`, so a quote, a backslash or a `</script>` in it cannot break the page;
- `escapeEmbedAttribute()` then keeps it inside the attribute, so `He said "hi" & <b>bold</b>` becomes `&quot;hi&quot; &amp; &lt;b&gt;` in the snippet rather than closing the attribute.

**The embed guide had the same gap** and gets a fixed `title="BoTTube video"`. That block is rendered as HTML entities and decoded again by `copyCode`, so a name taken from the page would have to survive two decodes to stay escaped, and a reader of that page is swapping the id and the title for their own anyway. Noted in a comment there.

The `/badges` verification preview and the `/embed-guide` live preview are already named, so with these two the copyable surfaces are covered.

## Evidence

`tests/js/embed_snippet_title.test.cjs` runs the real generator out of the template through `vm`, the same pattern as `embed_preview_selection.test.cjs`, with `tests/test_embed_snippet_title_runtime.py` as its pytest wrapper:

| input | asserted |
|---|---|
| `How agents learn` | exact snippet, name included |
| `He said "hi" & <b>bold</b>` | escaped in the attribute, and `<b>` never reaches the snippet as markup |
| `'single' quotes` | `&#39;` in the attribute |
| `""`, `"   "`, `undefined`, `null` | still named `BoTTube video` |
| size `100%x480` | responsive inline style preserved, name before it |

`tests/test_watch_template_accessibility.py` gains a template-level check so the `tojson` hand-off and the escape function cannot quietly disappear.

```
pytest tests/test_embed_snippet_title_runtime.py tests/test_watch_template_accessibility.py tests/test_embed_preview_selection_runtime.py
# 4 passed
```

Negative controls: dropping the `title` from the snippet fails the exact-shape cases, and dropping `escapeEmbedAttribute` fails with the title closing its own attribute:

```
'<iframe src="…" title="BoTTube video: He said "hi" & <b>bold</b>" width="853" …>'
```

I also checked Jinja's `tojson` against the inputs that matter, so the hand-off is not an assumption: `He said "hi"` → `"He said \"hi\""`, `</script><script>` → `"</script><script>"`, a missing title → `""`.

### DIFF
--- a/bottube_templates/embed_guide.html
+++ b/bottube_templates/embed_guide.html
@@ -359,7 +359,13 @@ function setVideo(videoId, btn) {
     iframe.title = "BoTTube video preview: " + btn.textContent.trim();
     document.getElementById("embed-demo").style.display = "block";
 
-    var codeText = '&lt;iframe src="https://bottube.ai/embed/' + videoId + '" width="560" height="315" frameborder="0" allowfullscreen&gt;&lt;/iframe&gt;';
+    // The snippet is what a reader copies onto their own site, so its iframe
+    // carries an accessible name (WCAG 4.1.2) exactly as the watch page's
+    // generator does. A fixed name rather than the picked video's: this block
+    // is rendered as HTML entities and decoded again on copy, so a name taken
+    // from the page would have to survive two decodes to stay escaped, and
+    // the reader is expected to swap VIDEO_ID and the title for their own.
+    var codeText = '&lt;iframe src="https://bottube.ai/embed/' + videoId + '" title="BoTTube video" width="560" height="315" frameborder="0" allowfullscreen&gt;&lt;/iframe&gt;';
     document.getElementById("embed-code-text").innerHTML = codeText;
     document.getElementById("embed-code-output").style.display = "block";
 }
--- a/bottube_templates/watch.html
+++ b/bottube_templates/watch.html
@@ -2935,6 +2935,7 @@
 <script>
 var myVote = {{ user_vote|default(0) }};
 var videoId = "{{ video.video_id }}";
+var videoTitle = {{ (video.title or "") | tojson }};
 var prefix = "{{ P }}";
 var uploaderAgent = "{{ video.agent_name }}";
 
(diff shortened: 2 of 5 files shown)


Scottcjn:
## RTC Reward (sent manually)

The automatic merge reward for this PR failed silently: our reward workflow's credentials had gone stale (being fixed now; failures are also now announced on the PR instead of passing silently). We've now paid it by hand: **3 RTC**, pending_id **4666** (24h hold, then confirmed). Sorry for the delay, @L4XB, and thank you for the contribution. — Sophia Elya

PR #2257, opened 2026-09-14
Bounty #16472: Bound deployment-drift job timeouts (complete CI hang fix)

MrHyplex9511:
Fixes #16472: CI hangs up to the 15-minute/360-minute workflow timeout.

Root cause (original burnout): the 2026-08-06 CI hang was a GitHub Actions infrastructure outage (action-download pulls got a 503 Service Unavailable and hosted-runner acquisition never returned a runner) - no contributor step ever executed, so the job sat at its runner-acquisition cap (15m01s). That class of failure was fixed in PR #1742 (job/step timeout-minutes, pip --timeout 30, concurrency cancellation, and the CI outcome report job that distinguishes cancelled-by-timeout from real failure).

Remaining gap fixed here: the deployment-drift job (added 2026-09-07, after #1742) is the only job in ci.yml with no timeout at all - job-level or step-level. It inherits Actions' default 360-minute limit, and its pip install + deployment_drift.py steps hit the network exactly like the 2026-08-06 outage did. This commit:

- Adds timeout-minutes: 10 to the deployment-drift job (same budget as lint/test/security).
- Adds per-step timeouts: checkout 3, setup-python 5, install 8, sentinel-test 5, drift-check 5.
- Adds pip --timeout 30 to its install step so pip retries are bounded, not hung.
- Wires deployment-drift into the existing CI outcome report job (needs + result var) so a timeout on this job is now reported as "cancelled at timeout limit" instead of an opaque red X - covering the "+5 RTC if the fix distinguishes timeout from failure" clause in the bounty.

No behavior changes to the checks themselves; the drift test and drift-check still run with the same commands.


Scottcjn:
`needs.deployment_drift.result` won't resolve; the job id is `deployment-drift`, so that expression is empty and the report step never sees this job. Use `needs['deployment-drift'].result`.

Scottcjn (approved):
Thanks. This gives the deployment-drift job the same timeout budget as the other jobs, and wires it into the report, so a network stall fails fast instead of running to the default. It was the first of three PRs for this residual, so it's the one we're taking. CI passes after updating onto main.

PR #2262, opened 2026-09-16
fix(rtc-services): reject non-object JSON bodies instead of returning 500

pengshenghai:
`/api/rtc/pay`, `/api/rtc/redeem` and `/api/rtc/use` parse the request body with

```python
data = request.get_json(force=True, silent=True) or {}
```

`or {}` only covers a *falsy* parse result. A JSON array, string, number or boolean is
truthy, so it survives the `or`, and the next `data.get(...)` raises `AttributeError` →
an unhandled **HTTP 500**.

`/api/rtc/redeem` and `/api/rtc/use` evaluate the body **before** any authentication
check, so that 500 was reachable with no credentials at all. `/api/rtc/pay` is protected
by its auth check and returned 401.

## Reproduction (before this change)

| request body | `/api/rtc/pay` | `/api/rtc/redeem` | `/api/rtc/use` |
|---|---|---|---|
| `{"service_token": "x"}` (control) | 400 | 404 | 404 |
| `[1, 2, 3]` | 500 | **500 (no auth)** | **500 (no auth)** |
| `"just-a-string"` | 500 | **500 (no auth)** | **500 (no auth)** |
| `42` | 500 | **500 (no auth)** | **500 (no auth)** |

```
AttributeError: 'list' object has no attribute 'get'
  rtc_services.py:236, in rtc_pay
    service_key = data.get("service_key", "")
```

## Change

Adds a small `_json_object_body()` helper to `rtc_services.py` and uses it at the three
call sites. It follows the same contract the rest of the server already uses — the
message and status match `tests/test_crosspost_json_validation.py`, e.g.:

```python
if not isinstance(data, dict):
    return None, (jsonify({"error": "JSON body must be an object"}), 400)
```

A missing or empty body still degrades to `{}` as before, so no existing caller changes
behaviour.

## After

| request body | status |
|---|---|
| `[1, 2, 3]` / `"str"` / `42` / `3.5` / `true` | **400** `JSON body must be an object` |
| well-formed object | unchanged |

## Tests

`tests/test_rtc_services_json_body_shape.py` — 31 cases, all passing:

* every endpoint × every truthy non-object body → 400 with the expected message
* `/api/rtc/redeem` and `/api/rtc/use` also asserted **without** credentials
* a well-formed object body still reaches the existing business logic (no regression)
* a missing body still behaves as before

```
$ python -m pytest tests/test_rtc_services_json_body_shape.py -q
31 passed
```

(The run also reports teardown `PermissionError: [WinError 32]` on Windows because the
temp SQLite file is still open when pytest removes the temp dir. That is pre-existing and
environmental — `tests/test_crosspost_json_validation.py` shows the same on the same
machine — and is not produced by this change.)

## Scope note

This PR fixes the three routes in `rtc_services.py`, including the two that were
reachable unauthenticated. The same `get_json(...) or {}` shape also appears in
`wrtc_bridge.py` (`/api/bridge/deposit`, `/api/bridge/withdraw`), `bottube_x402.py`
and a few `bottube_server.py` admin routes; those all sit behind an auth check first, so
they are lower impact. Happy to follow up on those in a separate PR if that is useful —
I did not want to widen this one.


pengshenghai:
Gentle review ping — 4 days open, no comments yet.

Scope is `rtc_services.py` only, so this does **not** overlap with the merged #2269 (which covered `wrtc_bridge.py`, `scraper_detective.py` and `bottube_x402.py`).

The two worth looking at first are `/api/rtc/redeem` and `/api/rtc/use` — they parse the body **before** any auth check, so the unhandled 500 was reachable with no credentials at all. `/api/rtc/pay` sits behind its auth check and returned 401.

- **2 files:** `rtc_services.py` (+25/−3) + `tests/test_rtc_services_json_body_shape.py` (31 cases, all passing)
- **Fix:** routes the three handlers through a local `_json_object_body()` helper using the same message/status contract as `tests/test_crosspost_json_validation.py`
- **Origin issue:** #16976

Happy to adjust anything on review — including splitting or reshaping if you'd prefer it folded into the same style as the merged change.

Issue #2139, opened 2026-09-02
Concurrent view requests can bypass the cooldown and double-count


nwinkelman2:
## Summary

Both view-recording paths implement the 30-minute `(video_id, IP)` cooldown as a separate recent-row `SELECT` followed by `INSERT` and counter/reward work:

- `GET|POST /api/videos/<video_id>/view`
- `GET /watch/<video_id>`

Concurrent requests can both observe no recent row before either commits, then both insert a view and increment the video's stored count. The API path can also evaluate view rewards and milestones twice for what should be one deduplicated event.

## Reproduction

1. Use a video/IP pair with no view inside the cooldown.
2. Send two view requests concurrently.
3. Let both requests execute the recent-view query before either insert commits.
4. Both record a row and increment the counter because the schema has no uniqueness boundary for a rolling time window.

## Expected behavior

The cooldown decision and view event/counter mutation are atomic per database. Exactly one concurrent request creates the event and runs new-view effects; every other request observes the committed event as a deduplicated view. The stored counter matches the number of admitted events.

## Proposed fix

Serialize the recent-view decision, event insert, and counter increment with a short SQLite write transaction shared by both routes. Keep API reward/milestone work in the winning transaction, and add a multi-connection regression proving one admitted event under concurrent load.

This is an ordinary analytics/reward data-integrity defect under the ongoing RustChain bounty pool; no production data was modified.

rafaio1:
PR #2140 by @nwinkelman2 already fixes this exact concurrent view cooldown bypass with serialized SQLite transactions. My PRs #2173-2176 address related GPU marketplace concurrency issues (#2163, #2165, #2167, #2169) which are separate bugs in the same codebase.

@Scottcjn — requesting review on:
- #2140 (view cooldown fix for this issue)
- #2173-2176 (GPU job claim/settlement atomicity fixes)

All have passing CI checks and are mergeable. Ready for bounty payout upon merge.

PR #2267, opened 2026-09-17
fix(a11y): add accessible group names and aria-pressed states to analytics period selectors (#2263)

### LINKED ISSUE #2263: [Bug] Analytics period selectors expose selected range only visually
## What happened?

The **View Trends** and **Engagement Metrics** sections on `/analytics` each expose three mutually exclusive period buttons: `7 Days`, `30 Days`, and `90 Days`. The current period is represented only by the CSS class `active`.

None of the six buttons exposes `aria-pressed`, `aria-selected`, or another programmatic selected-state attribute. Neither three-button set has an accessible group name. When another period is activated, the script removes and adds only the `active` class:

```js
this.parentElement.querySelectorAll('button').forEach(b => b.classList.remove('active'));
this.classList.add('active');
```

Consequently, a screen-reader user can activate a period and receive the corresponding analytics data but cannot determine which period is selected. The same state is visually apparent from the active styling.

This affects WCAG 2.2 SC 1.3.1 (Info and Relationships) and 4.1.2 (Name, Role, Value).

## Expected behavior

Each period selector should expose its selected state programmatically. The two button sets should also have distinct accessible group names.

A suitable implementation would retain the native buttons, wrap each set in a named `role="group"`, initialize exactly one button with `aria-pressed="true"`, and update every button's `aria-pressed` value from the same branch that updates the `active` class.

## Steps to reproduce

1. Open `https://bottube.ai/analytics` with a screen reader or inspect the accessibility tree.
2. Locate either the **View Trends** or **Engagement Metrics** period selector.
3. Inspect the `7 Days`, `30 Days`, and `90 Days` buttons.
4. Activate a different period.
5. Observe that the visual `active` class changes, but no programmatic selected state is exposed before or after activation.

## Related page

`https://bottube.ai/analytics`

## Source verification

Reverified against current `main` at commit `6af6b63f7a5a87353a30cd4552dc5c0e183a96af` in `bottube_templates/analytics.html`. The two selector groups are around lines 254–257 and 269–272, and the selection handler is around lines 512–526.

The separate **Top Videos** metric controls already implement a named group and `aria-pressed`; this report concerns only the two analytics period selectors.

I searched the current BoTTube issues and the RustChain bounty #1618 discussion for `period-selector`, the 7/30/90-day selector, and analytics-related `aria-pressed` reports. I found no existing report for this defect. BoTTube #2041 concerned the separate Top Videos ranking controls, and #2102 concerned stale asynchronous responses.

AI assistance disclosure: the live behavior, public source, duplicate history, and report text were audited with AI assistance.

Cid-oe:
## Summary
Fixes https://github.com/Scottcjn/bottube/issues/2263

### Problem
The View Trends and Engagement Metrics sections on `/analytics` each expose three period buttons (`7 Days`, `30 Days`, `90 Days`). Previously, the active state was represented solely via CSS class `active`, with no `aria-pressed`, `role="group"`, or accessible group labels. Consequently, assistive tech/screen reader users could not programmatically determine which period was active.

### Solution
1. Added `role="group"` and distinct accessible labels (`aria-label="View trends time period"` and `aria-label="Engagement metrics time period"`) to the `.period-selector` containers.
2. Initialized buttons with `type="button"` and `aria-pressed="true"` for the default active button (`7 Days`) and `aria-pressed="false"` for others.
3. Updated the click event listener in `bottube_templates/analytics.html` to toggle both the `active` class and `aria-pressed` attribute atomically across peer buttons.

### Verification
- Verified HTML template syntax.
- Confirmed pattern matches existing accessible controls (e.g. Top Videos metric tabs).


Scottcjn (approved):
Approved. The period buttons get the same `aria-pressed` pattern as the metric tabs from #2139, and it's template/JS only. Thanks @Cid-oe. This is the right size of PR.

Scottcjn:
## RTC Reward (sent manually)

The automatic merge reward for this PR failed silently: our reward workflow's credentials had gone stale (being fixed now; failures are also now announced on the PR instead of passing silently). We've now paid it by hand: **2 RTC**, pending_id **4669** (24h hold, then confirmed). Sorry for the delay, @Cid-oe, and thank you for the contribution. — Sophia Elya

PR #2280, opened 2026-09-19
fix(websocket): authenticate connection, prevent payload identity spoofing, and authorize mod actions (#2274)

### LINKED ISSUE #2274: websocket_server.py has no authentication: must not be wired in until it does
Found while reviewing #2271. **Not a live issue**: `init_socketio()` is never called anywhere on `main`, so this Socket.IO layer isn't served today. It's filed so nobody wires it in as it stands.

### Problems
- **No identity at all.** `chat_message` takes `user_id` and `username` from the client payload, so anyone can post as anyone.
- **`mod_action` has no authorization.** Any connected client can `ban` (permanent if `duration` is omitted), `timeout`, or set `slow_mode` for any user on any video. `banned_by` is the client-supplied `mod_name`.
- **Bans can be bypassed.** They're keyed on the client-asserted `user_id`, so a banned user just sends a different one.

### Before this can be enabled
1. Authenticate the socket on connect, via the existing session cookie or an `X-API-Key`-equivalent handshake. Never take identity from the event payload.
2. Derive `user_id`/`username` server-side from that identity.
3. `mod_action`: require the caller to own the video's channel, or be an admin. Record the real actor in `banned_by`.
4. Regression tests for each of the above.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

dwaddle:
### Summary
Resolves #2274 by implementing authentication and authorization controls for `websocket_server.py`:

1. **Connection & Event Authentication**:
   - Extracts user/agent identity from Flask session (`user_id`), `X-API-Key` / `Authorization: Bearer <key>` headers, or payload handshake tokens.
   - Rejects unauthenticated requests with `emit("error", {"message": "Authentication required"})`.

2. **Payload Identity Spoofing Protection**:
   - Rejects events where payload `user_id` or `username` conflicts with the server-derived identity.
   - Sets `user_id` and `username` exclusively from server-side authenticated context.

3. **Channel Ownership & Moderator Authorization**:
   - `on_mod_action` (`ban`, `timeout`, `slow_mode`) requires caller to be a platform moderator/admin or own the channel (`videos.agent_id` / `videos.uploader`).
   - Records real authenticated actor username in `chat_bans.banned_by`.

### Verification
- **Whitespace Audit**: `git diff --check` passed with 0 errors.
- **Automated Tests**: Ran `pytest tests/test_websocket_server_validation.py tests/test_chat_video_existence_validation.py` (24/24 passing).
- **Regression Tests**: Added `test_unauthenticated_websocket_event_rejected`, `test_payload_identity_spoofing_rejected`, and `test_mod_action_rejected_for_non_mod_non_owner`.


Scottcjn (changes requested):
Thanks, the structure is right: identity from the server, spoof checks, and a gated mod_action. The problem is that BoTTube's login only sets `session["user_id"] = agents.id` (an integer, bottube_server.py around line 5224). It never sets username, is_mod or is_admin, so the tests fake session state production never has. Concretely:

1. Resolve the session user_id to its agents row. Use agents.id as the user id and agent_name or display_name as the username, for both the cookie and API-key paths, so one person has one identity. Right now the same agent is "42" by cookie and "alice" by key, so a ban on one misses the other.
2. Do the owner check as videos.agent_id == agents.id.
3. Take mod/admin from an agents column, not the session. If there isn't one, keep it owner-only and say so.
4. Reject banned agents on the API-key path, the way require_api_key does.
5. Rewrite the owner and mod tests to log in the way the real login does, and add a test that cookie and API key give the same user id.

dwaddle:
Updated PR #2280 per review feedback:

1. **Unified Identity Resolution (`agents.id`)**: Both cookie session (`session["user_id"] = agents.id`) and API-key paths now resolve to the matching `agents` row, returning `user_id = str(agents.id)` and `username = display_name or agent_name`. Cookie and API-key now share the identical user ID (`'42'`).
2. **Channel Owner Authorization**: Updated `_is_mod_or_channel_owner()` to perform owner check strictly via `videos.agent_id == agents.id` (`str(videos.agent_id) == auth_user["user_id"]`).
3. **Agent Column Moderation**: `is_mod` is extracted directly from the `agents` table column (`is_mod` / `is_admin`). If missing from schema, moderation remains owner-only.
4. **Banned Agent Rejection**: API-key path now checks `is_banned` and rejects banned agents with 401/403 authentication failure matching `require_api_key`.
5. **Unit Tests**: Added tests `test_cookie_and_api_key_give_same_user_id` and `test_banned_agent_rejected_on_api_key_path`, and updated owner/mod tests to use integer `session["user_id"] = agents.id` matching real login. All 18 tests passing (0.87s).

PR #2281, opened 2026-09-19
test(feed): add regression test for _fetch_videos outside request context

dwaddle:
## Summary
Adds a regression unit test in `tests/test_feed_blueprint.py` (`test_fetch_videos_outside_request_context_uses_api_base`) locking in the existing behavior of `_fetch_videos()` when executing outside a Flask request context.

The test verifies that when no Flask request context is active (`has_request_context() == False`), `_fetch_videos()` correctly falls back to resolving `BOTTUBE_API_BASE` without raising a `RuntimeError`.

## Verification
- `git diff --check` passes with 0 whitespace/formatting errors.
- `pytest tests/test_feed_blueprint.py` passes 100% (37/37 passed in 0.59s).


Scottcjn (changes requested):
Thanks @dwaddle. The test is good and passes (37/37), but the PR body doesn't match the diff.

The body says it "guarded request host access in `_fetch_videos()` using `has_request_context()`" — **that guard is already on main** (`feed_blueprint.py:238`, merged 2026-09-13, before this PR). This PR changes **tests only**.

Two small things and this can merge:
1. Rewrite the description to say what it is: a regression test locking in the already-merged `has_request_context()` guard. Please drop the claim of adding the guard.
2. Remove the 2 extra blank lines before `test_feed_routes_escape_url_attributes_and_cdata`.

Also, don't close #1983 on this PR unless the original guard commit is credited instead.

To be clear about why this matters: a description claiming a source fix that isn't in the diff reads like an overclaim, and we check. A regression test for existing behaviour is genuinely welcome — just describe it as one. — Sophia Elya

dwaddle:
Thanks @Scottcjn for the feedback!

I have updated the PR as requested:
1. Updated the PR title and description to accurately reflect that this PR adds a regression test locking in the existing `has_request_context()` behavior outside a request context, without claiming source code changes or closing #1983.
2. Removed the extra blank lines in `tests/test_feed_blueprint.py` before `test_feed_routes_escape_url_attributes_and_cdata`.

